PR #222, opened 2018-10-05
Maintenance auto-pts zephyr week 40

kolodgrz:
- According to updated GATT db also values of auto-pts pre-stored attribute values has to be changed
- Generic attribute handle method is implemented to simplify maintenance


rmstoi:
Thanks!

PR #231, opened 2018-12-10
zephyr: gatt: Update gatt attribute database

kolodgrz:
Patch update GATT test cases attributes for verification with current
Zephyr OS predefined attributes.


rmstoi:
Thanks!

PR #242, opened 2019-03-06
Mtn y19 w10

kolodgrz:
Fixed:
- Parsing non printable ASCII characters in logger (causing exceptions and blocks autoptsserver)
- Revert set PIXITS for test case. Next test case shouldn't rely on previous test case setup
- Re-set recovery items if already exist
- Remove unused PIXITS

Upgrade:
- Intentionaly set all PIXITS for profiles to avoid using non-controlable setup
- Upgrade Zephyr OS test plan to TCRL 2018-2


autoptsclient-zephyr.py
```diff
@@ -160,6 +160,13 @@ def main():
     stack_inst.synch_init(callback_thread.set_pending_response,
                           callback_thread.clear_pending_responses)
 
+    # Setup project PIXITS
+    autoprojects.gap.set_pixits(ptses[0])
+    autoprojects.gatt.set_pixits(ptses[0])
+    autoprojects.sm.set_pixits(ptses[0])
+    autoprojects.l2cap.set_pixits(ptses[0])
+    autoprojects.mesh.set_pixits(ptses)
```
MariuszSkamra: causes an error while running with one PTS dongle:

Traceback (most recent call last):
  File "./autoptsclient-zephyr.py", line 205, in <module>
    main()
  File "./autoptsclient-zephyr.py", line 168, in main
    autoprojects.mesh.set_pixits(ptses)
  File "/home/mskamra/projects/auto-pts/ptsprojects/zephyr/mesh.py", line 57, in set_pixits
    pts2 = ptses[1]
IndexError: list index out of range

kolodgrz: Yeah, I've forget about it since I'm always using two dongles. Good catch :)

PR #246, opened 2019-03-19
Partial update to PTS 7.4.1

MariuszSkamra:
This updates the autopts to PTS 7.4.1. The work is still not complete, and there will be more related PR's soon.


tedd-an:
The result looks much better with this PR.
Tested with 7.3.0 though.

PR #292, opened 2019-03-28
Fix GATT/SR/GAR/BV-06-C test case

MariuszSkamra:


tedd-an:
@MariuszSkamra 

I tried this PR and confirmed that ```GATT/SR/GAR/BV-06-C``` test pass.
But ```GAP/SEC/AUT/BV-14-C``` is ```INCONC``` and here is the log...
[GAP_GAP_SEC_AUT_BV-14-C.log](https://github.com/intel/auto-pts/files/3019323/GAP_GAP_SEC_AUT_BV-14-C.log)

MariuszSkamra:
This test case will not pass. I only changed the IO capabilities so that authenticated pairing would be performed. See the PTS issue I reported https://github.com/intel/auto-pts/issues/256

Issue #298, opened 2019-04-05
bot: Add result text file


tedd-an:
Need to add result text file for easy parsing later with other tools like TestRail

Issue #300, opened 2019-04-05
bot: Use git-describe instead of commit


tedd-an:
There was a request from Zephyr team to include git-describe for source info instead of raw commit id.

Issue #299, opened 2019-04-05
bot: Attach the result files to email


tedd-an:
To share the test result easily, it is a good idea to attach the result.xlsx to email.

PR #301, opened 2019-04-05
bot: small changes for email and result

tedd-an:
Small changes to add result.txt and attach result files to email.

Fixes: #298, #299, #300

### DIFF
--- a/bot/zephyr.py
+++ b/bot/zephyr.py
@@ -424,8 +424,11 @@ def main(cfg):
         reg_html = bot.common.regressions2html(_regressions)
 
         # Zephyr commit id link in HTML format
-        zephyr_hash_html = bot.common.url2html(zephyr_hash_url(zephyr_hash),
-                                               zephyr_hash)
+
+        # Commit id may have "-dirty" if the source is dirty.
+        commit_id = zephyr_hash["commit"].split('-')[0]
+        zephyr_hash_html = \
+            bot.common.url2html(zephyr_hash_url(commit_id), zephyr_hash["desc"])
 
         # Log in Google drive in HTML format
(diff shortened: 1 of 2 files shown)


bot/config.py.sample
```diff
@@ -104,4 +104,14 @@ z['scheduler'] = {
     'friday': '20:00',
 }
 
+# ****************************************************************************
+# Post run configuration
+#
+# To do any post run tasks. It runs 'script' and followed by 'cleanup'
+# ****************************************************************************
+z['postrun'] = {
+    'script': '/path/to/script',
```
MariuszSkamra: I see you are running 'script' right after 'cleanup' so I guess you could simply provide a list of paths to scripts and just run all of those in for loop.
z['postrun'] = ['/path/to/script', '/path/to/cleanup/script']

This way you will not duplicate the code [1]

bot/zephyr.py
```diff
@@ -443,6 +444,29 @@ def main(cfg):
...
+        if 'script' in postrun:
+            script = postrun['script']
+            if os.path.isfile(script):
+                script_proc = subprocess.Popen(shlex.split(script),
+                                               shell=True)
+                if script_proc.wait():
+                    print("Failed to run post script")
+
+        # Run post run script
+        if 'cleanup' in postrun:
```
MariuszSkamra: [1] here

tedd-an: I was going to use this routine to upload the test result to Zephyr TestRail site and reset the VM. I need to more think about it...   :)

bot/config.py.sample
tedd-an: Good idea.

PR #302, opened 2019-04-10
bot/common: Improve updating sources

michal-narajowski:
Previously you could only specify a remote from which
you could update the sources and you could only use 'master'
branch. Moreover 'reset --hard' had been used to do an update
so there was a possibility of losing some changes.

This patch introduces an improvement to the mechanism of
updating source code repositories. There is a new section
in the bot configuration that specifies what repositories are to
be updated. This functionality supports multiple repositories.
You can specify a path to the repository, absolute
or relative, a remote and a branch. You can also specify
if you want to stash non-committed changes. If the value
is False and there are changes then the repository will not be updated.

### DIFF
--- a/bot/common.py
+++ b/bot/common.py
@@ -346,19 +346,54 @@ def archive_recursive(dir_path):
     return zip_file_path
 
 
-def update_sources(repo, remote):
+def update_sources(repo, remote, branch, stash_changes=False):
     """GIT Update sources
-    :param repo: project git repository path
-    :param remote: remote name
-    :return:
+    :param repo: git repository path
+    :param remote: git repository remote name
+    :param branch: git repository branch name
+    :param stash_changes: stash non-committed changes
+    :return: Commit SHA at HEAD
     """
+    print('Updating ' + repo)
+
     repo = git.Repo(repo)
+    dirty = repo.is_dirty(untracked_files=True)
+
+    if dirty and (not stash_changes):
+        print('Repo is dirty. Not updating')
+        return repo.git.show('-s', '--format=%H') + '-dirty'
+
+    if dirty and stash_changes:
+        print('Repo is dirty. Stashing changes')
+        repo.git.stash('--include-untracked')
+
     repo.git.fetch(remote)
-    repo.git.checkout('master')
-    repo.git.reset('--hard', '{}/master'.format(remote))
-    repo.git.clean('-fd')
+    repo.git.checkout('{}/{}'.format(remote, branch))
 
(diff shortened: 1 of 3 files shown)


kolodgrz:
> This PR proposes a change to the update_sources common method.
> 
> Some reasons behind it:
> 
> 1. Currently, you can only specify remote, not the branch that you want to check out. So basically you are restricted to 'master' branch.
I'll rather prefer to extend config with branch parameter (e.g. "branch_name") and change current naming of "git_branch" to "remote_name".
> 2. If a repo is dirty, changes will be purged without warning.
Primal idea for bot was to do periodical "runs" with HEAD on masters latest commit.

To extend also this additional config may be required e.g. "allow_dirty_branch". Which will ignore updating sources.

> 
> With this change, the bot does not rely on config to know which branch and remote to use. The bot uses the current branch and it's tracking branch to update the sources.
> 
> We could also add a return value to stop bot when the repo is dirty.
Having many configs will resolve any problem. No problem with developer branch on upstream, no problem with master weekly etc.
> 
> Comments are welcome.

michal-narajowski:
@kolodgrz I proposed this approach because Mynewt requires multiple repositories (at least two: core and nimble) to work. It would be more complicated to do this with a config.

kolodgrz:
> @kolodgrz I proposed this approach because Mynewt requires multiple repositories (at least two: core and nimble) to work. It would be more complicated to do this with a config.

That shouldn't be complicated also. Maybe it'll be worth to create array of git setups in config:
"'git_setups': (('path_to_repo1/git', 'origin', 'HEAD'),
                      ('path_to_repo2/git', 'test', None))

and in script do update source in loop based on git_setups len.

MariuszSkamra:
@kolodgrz I have a feeling it would be even cleaner to have separate BotProject key "git" configuration with it's key-value pairs.
z['git'] = {
'repo1' : { 'path' : 'path_to_repo1/git', 'remote' : 'origin', 'branch' : 'master' },
...
}

kolodgrz:
> @kolodgrz I have a feeling it would be even cleaner to have separate BotProject key "git" configuration with it's key-value pairs.
> z['git'] = {
> 'repo1' : { 'path' : 'path_to_repo1/git', 'remote' : 'origin', 'branch' : 'master' },
> ...
> }

Agree.

michal-narajowski:
@kolodgrz @MariuszSkamra Updated code and description.

michal-narajowski:
@tedd-an It is updated.

PR #303, opened 2019-04-12
ptscontrol: Do not send response parameter on implicit send callback

michal-narajowski:
These parameters are only useful on server side. This fixes using latest
auto-pts with PTS 7.3.0. Autopts server cannot marshall response which
is of IntPtr type.

Relevant autoptsserver.log fragment:

```
BEGIN OnImplicitSend:
project_name: GAP <type 'str'>
wid: 47 <type 'UInt16'>
test_case_name: GAP/BROB/BCST/BV-01-C <type 'str'>
description: Please enter broadcasting mode using specified TSPX_advertising_data value and start advertising. <type 'str'>
style: MMI_Style_Ok_Cancel2 0x11141
response:  <System.IntPtr object at 0x000000000000009B [7318720]> <type 'IntPtr'> 155
response_size: 2048 <type 'UInt32'>
response_is_present:  0 <type 'StrongBox[int]'>
Calling callback.on_implicit_send
Caught exception
cannot marshal <type 'IntPtr'> objects
PTS_LOGTYPE_IMPLICIT_SEND PTS ControlServer  Server: SUCCESS but response is not present

### DIFF
--- a/autoptsclient_common.py
+++ b/autoptsclient_common.py
@@ -151,7 +151,7 @@ class ClientCallback(PTSCallback):
             sys.exit("Exception in Log")
 
     def on_implicit_send(self, project_name, wid, test_case_name, description,
-                         style, response, response_size, response_is_present):
+                         style):
         """Implements:
 
         interface IPTSImplicitSendCallbackEx : IUnknown {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by rmstoi** [comment]
> This fixes using latest auto-pts with PTS 7.3.0.

What issue does this fix? Please provide a reference or describe.

**Comment by michal-narajowski** [comment]
@rmstoi Updated the description.

File: autoptsclient_common.py [hunk 1]
```diff
@@ -160,10 +160,7 @@ def on_implicit_send(self, project_name, wid, test_case_name, description,
                     [in] unsigned short wID,
                     [in] LPWSTR pszTestCase,
                     [in] LPWSTR pszDescription,
-                    [in] unsigned long style,
```
rmstoi: This is copy/paste output from oleview.exe. Makes sense to keep it as it was.

MariuszSkamra:
LGTM

michal-narajowski:
@rmstoi Updated

PR #308, opened 2019-04-16
bot: Ignore the untracked files

tedd-an:
Turning on the untracked_files option prevent the bot from reusing the
same repo again because the bot generates the .conf file depends on the
iut_config and it doesn't delete it after the execution.


tedd-an:
@michal-narajowski FYI...

PR #309, opened 2019-04-17
btp: Add some missing commands and mechanisms

michal-narajowski:
This PR adds:
- Discover all primary services command
- Parsing GATT Find Included Services response
- AdType define for URI type


pybtp/btp.py
```diff
@@ -1100,7 +1100,25 @@ def gatts_start_server():
...
+
+    db_attr_off, db_attr_cnt = struct.unpack_from(fmt,
+                                                  tuple_data[0])
+
+    logging.debug("%s %d %d", gatts_start_server.__name__,
+                  db_attr_off, db_attr_cnt)
+
+    stack = get_stack()
+    stack.gap.db_attr_off = db_attr_off
+    stack.gap.db_attr_cnt = db_attr_cnt
```
kolodgrz: If I remember correctly this approach (attr offset, attr cnt) is deprecated. Now server "should" get whole GATT database which contains handles.

michal-narajowski:
@MariuszSkamra @kolodgrz Thank you, I removed the GATT Server Start changes.

pybtp/btp.py
```diff
@@ -1344,16 +1384,66 @@ def gattc_find_included(bd_addr_type, bd_addr, start_hdl, stop_hdl):
...
+def gattc_find_included(bd_addr_type, bd_addr, start_hdl=None, end_hdl=None):
+    logging.debug("%s %r %r %r %r", gattc_find_included.__name__,
+                  bd_addr_type, bd_addr, start_hdl, end_hdl)
+    gap_wait_for_connection()
+
+    if start_hdl and end_hdl:
+        _gattc_find_included_req(bd_addr_type, bd_addr, start_hdl, end_hdl)
+        return
+
+    gattc_disc_prim_svcs(bd_addr_type, bd_addr)
```
tedd-an: If the start_hdl and end_hdl are NULL, I think it should return as "invalid parameter", instead of running a "discovery primary services". I couldn't find any info about it in the SPEC.
Please provide a reference or details.

MariuszSkamra:
@michal-narajowski ping. Are you going to update this PR?

PR #310, opened 2019-04-17
bot: Add support postrun option

tedd-an:
This patch adds a support for postrun option which executes the scripts
after running the tests.

### DIFF
--- a/bot/config.py.sample
+++ b/bot/config.py.sample
@@ -115,4 +115,12 @@ z['scheduler'] = {
     'friday': '20:00',
 }
 
+# ****************************************************************************
+# Post run configuration
+#
+# To do any post run tasks.
+# ****************************************************************************
+z['postrun'] = ['/path/to/script params1 params2',
+		'/path/to/script2 params1 params2 param3']
+
 BotProjects.append(zephyr_nrf52)
--- a/bot/common.py
+++ b/bot/common.py
@@ -23,6 +23,8 @@ import sqlite3 as sql
 import git
 import shutil
 import zipfile
+import shlex
+import subprocess
 
 import smtplib
 from email.MIMEMultipart import MIMEMultipart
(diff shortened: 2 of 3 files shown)


bot/config.py.sample
```diff
@@ -115,4 +115,12 @@ z['scheduler'] = {
     'friday': '20:00',
 }
 
+# ****************************************************************************
+# Post run configuration
+#
+# To do any post run tasks.
+# ****************************************************************************
+z['postrun'] = ['/path/to/script params1 params2',
+		'/path/to/script2 params1 params2 param3']
```
MariuszSkamra: align

kolodgrz:
@tedd-an do you have any specific scenario with usage of this postrun ?
I'm wondering if it'll be sufficient to introduce importable postrun iut specific module ("*.py" as is done for config.py). You may develope for all (public pushed to autopts) or have your own (private on disk and your personal use) Do we need another separate process ? I see you start process and next script command is just waiting for its end.

tedd-an:
> @tedd-an do you have any specific scenario with usage of this postrun ?
> I'm wondering if it'll be sufficient to introduce importable postrun iut specific module ("*.py" as is done for config.py). You may develope for all (public pushed to autopts) or have your own (private on disk and your personal use) Do we need another separate process ? I see you start process and next script command is just waiting for its end.

tedd-an:
@kolodgrz The main use case for postrun was to upload the test result to TestRail (https://zephyrproject.testrail.io/index.php?/projects/overview/6) which uses the script from here(https://github.com/zephyrproject-rtos/qm/blob/master/scripts/test_reports/report.py).
Could you be more specific about
>I'm wondering if it'll be sufficient to introduce importable postrun iut specific module ("*.py" as is done for config.py)
It can be imported as a module but don't want to make it a public code since not everyone can access the site and it is sort of internal things.

tedd-an:
Discard the PR at this ponit.

PR #313, opened 2019-04-19
Some fixes to gatt_server_fetch_db

michal-narajowski:
Apart from these fixes, I think this method should be moved to a common file.

### DIFF
--- a/ptsprojects/zephyr/gatt_wid.py
+++ b/ptsprojects/zephyr/gatt_wid.py
@@ -82,7 +82,10 @@ def gatt_server_fetch_db():
             hdr_len = struct.calcsize(hdr)
             uuid_len = val_len - hdr_len
             incl_svc_hdl, end_grp_hdl, uuid = struct.unpack(hdr + "%ds" % uuid_len, val)
-            uuid = btp.btp2uuid(uuid_len, uuid)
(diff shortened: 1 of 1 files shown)


ptsprojects/zephyr/gatt_wid.py
```diff
@@ -82,7 +82,10 @@ def gatt_server_fetch_db():
             hdr_len = struct.calcsize(hdr)
             uuid_len = val_len - hdr_len
             incl_svc_hdl, end_grp_hdl, uuid = struct.unpack(hdr + "%ds" % uuid_len, val)
-            uuid = btp.btp2uuid(uuid_len, uuid)
+            if uuid_len > 0:
```
MariuszSkamra: good catch!

ptsprojects/zephyr/gatt_wid.py
```diff
@@ -41,11 +41,16 @@ def gatt_wid_hdl(wid, description, test_case_name):
 
 def gatt_server_fetch_db():
     db = GattDB()
+    char_val_set = set()
```
MariuszSkamra: In fact it was intended. I don't like this patch, because GattDB will lack Characteristic Values. Please add new class representation for those attributes

PR #316, opened 2019-05-16
Add coloring to test result logs

szymon-czapracki:
Make results of test cases stand out a bit more
with font color matching the given result.

### DIFF
--- a/autoptsclient_common.py
+++ b/autoptsclient_common.py
@@ -32,6 +32,7 @@ from collections import OrderedDict
 import xml.etree.ElementTree as ET
 import time
 import datetime
+from termcolor import colored
 
 from ptsprojects.testcase import get_max_test_case_desc
 from ptsprojects.testcase import PTSCallback
@@ -503,6 +504,17 @@ def get_test_case_description(cache, test_case_name):
             return test_case.find('description').text
 
 
+def get_result_color(status):
+    if status == "PASS":
+        return "green"
+    elif status == "FAIL":
+       return "red"
+    elif status == "INCONC":
+        return "yellow"
+    else:
+        return "white"
+
+
 def run_test_case_wrapper(func):
     def wrapper(*args):
         test_case = args[1]
@@ -549,10 +561,16 @@ def run_test_case_wrapper(func):
         end_time = str(round(datetime.timedelta(
             seconds=end_time).total_seconds(), 3))
 
-        print("{}".format(test_case.status).ljust(16) +
-              end_time.rjust(len(end_time)) +
-              retries_msg.rjust(len("#{}".format(retries_max)) + margin) +
-              regression_msg.rjust(len("REGRESSION") + margin))
+        result = ("{}".format(test_case.status).ljust(16) +
+                end_time.rjust(len(end_time)) +
+                retries_msg.rjust(len("#{}".format(retries_max)) + margin) +
+                regression_msg.rjust(len("REGRESSION") + margin))
+
+        if sys.stdout.isatty():
+            output_color = get_result_color(test_case.status)
+            print(colored((result), output_color))
+        else:
+            print(result)
 
     return wrapper
 
--- a/README.md
+++ b/README.md
@@ -33,6 +33,12 @@ Over 460 test cases have been automated for Zephyr OS which reduced testing time
 
 **Bluetooth Test Protocol (BTP)**: Used to communicate with the IUT. See `doc/btp_spec.txt`
 
+# Linux Prerequisites
+
+Install required modules with:
+
+`python2 -m pip install --user -r autoptsclient_requirements.txt`
+
 # Windows Prerequisites
 
 To be able to run PTS in automation mode, there should be no PTS instances running in the GUI mode. Hence, before running these scripts close the PTS GUI.
--- a/autoptsclient_requirements.txt
+++ b/autoptsclient_requirements.txt
@@ -1,0 +1,1 @@
+termcolor


autoptsclient_common.py
```diff
@@ -32,6 +32,7 @@
 import xml.etree.ElementTree as ET
 import time
 import datetime
+from termcolor import colored
```
rmstoi: Note that using termcolor is not necessary, [btpclient.py](https://github.com/intel/auto-pts/blob/master/tools/btpclient.py#L44) is using SGR ANSI escape codes to set colors. Nevertheless, I do understand that termcolor makes things easier.

autoptsclient_common.py
```diff
@@ -548,11 +549,26 @@ def wrapper(*args):
...
-        print("{}".format(test_case.status).ljust(16) +
-              end_time.rjust(len(end_time)) +
-              retries_msg.rjust(len("#{}".format(retries_max)) + margin) +
-              regression_msg.rjust(len("REGRESSION") + margin))
+        if test_case.status == "PASS":
+            print(colored(("{}".format(test_case.status).ljust(16) +
+                end_time.rjust(len(end_time)) +
+                retries_msg.rjust(len("#{}".format(retries_max)) + margin) +
+                regression_msg.rjust(len("REGRESSION") + margin)), "green"))
+        elif test_case.status == "FAIL":
```
rmstoi: Colored text should not be printed if stdout is redirected into a file, so that file won't contain ANSI escape codes. Viewing text with escape codes in a text editor shows:

^[[32mauto-pts is awesome!^[[0m

Solution could be not to use colors if stdout is not a TTY:

```
txt = 'auto-pts is awesome!'
if sys.stdout.isatty():
    txt = colored(txt, 'green')
print txt
```

autoptsclient_common.py
szymon-czapracki: I've tried using escape codes, at the beginning that seemed like the best way, but it turned out that using them messes up justifying text output in the terminal so I decided to use termcolor.

autoptsclient_common.py
```diff
@@ -548,11 +549,26 @@ def wrapper(*args):
 
         end_time = str(round(datetime.timedelta(
             seconds=end_time).total_seconds(), 3))
-
-        print("{}".format(test_case.status).ljust(16) +
-              end_time.rjust(len(end_time)) +
-              retries_msg.rjust(len("#{}".format(retries_max)) + margin) +
-              regression_msg.rjust(len("REGRESSION") + margin))
+        if test_case.status == "PASS":
```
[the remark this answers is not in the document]
szymon-czapracki: Done

autoptsclient_common.py
szymon-czapracki: Done

autoptsclient_common.py
```diff
@@ -549,10 +561,17 @@ def wrapper(*args):
         end_time = str(round(datetime.timedelta(
             seconds=end_time).total_seconds(), 3))
 
-        print("{}".format(test_case.status).ljust(16) +
-              end_time.rjust(len(end_time)) +
-              retries_msg.rjust(len("#{}".format(retries_max)) + margin) +
-              regression_msg.rjust(len("REGRESSION") + margin))
+        if sys.stdout.isatty():
+            output_color = get_the_color(test_case.status)
+            print(colored(("{}".format(test_case.status).ljust(16) +
```
MariuszSkamra: remove duplicates

szymon-czapracki: Done

autoptsclient_common.py
```diff
@@ -503,6 +504,17 @@ def get_test_case_description(cache, test_case_name):
             return test_case.find('description').text
 
 
+def get_the_color(status):
```
rmstoi: get_the_color is not easy to understand. I would suggest to change functions name to:

```
get_result_color
get_term_color
get_text_color
```

Or something similar, without 'the' but actually specifying what the color is used for.

Other than that PR looks great

szymon-czapracki: Done

MariuszSkamra:
@tedd-an Are you OK with this PR?

tedd-an:
LGTM.

autoptsclient_common.py
```diff
@@ -549,10 +561,16 @@ def wrapper(*args):
...
-              end_time.rjust(len(end_time)) +
-              retries_msg.rjust(len("#{}".format(retries_max)) + margin) +
-              regression_msg.rjust(len("REGRESSION") + margin))
+        result = ("{}".format(test_case.status).ljust(16) +
+                end_time.rjust(len(end_time)) +
+                retries_msg.rjust(len("#{}".format(retries_max)) + margin) +
+                regression_msg.rjust(len("REGRESSION") + margin))
+
+        if sys.stdout.isatty():
+            output_color = get_the_color(test_case.status)
```
tedd-an: I actually ran the PR and found the issue: it should be get_result_color().

Please update and submit again.

autoptsclient_common.py
```diff
@@ -503,6 +504,17 @@ def get_test_case_description(cache, test_case_name):
             return test_case.find('description').text
 
 
+def get_result_color(status):
```
tedd-an: Add more coloring for different status like PTS TIMEOUT, BTP TIMEOUT and BTP ERROR?

autoptsclient_common.py
szymon-czapracki: Done, sorry for a silly mistake.

autoptsclient_common.py
szymon-czapracki: I think that might be too much, with coloring for PASS, FAIL, INCONC and no coloring for other statuses it's clearer, otherwise, it looks too rainbowish.

PR #319, opened 2019-06-05
Fix test case related to SAR timeout

MariuszSkamra:
Relates to: https://github.com/zephyrproject-rtos/zephyr/pull/16601


MariuszSkamra:
@michal-narajowski FYI

PR #323, opened 2019-07-01
[RFC] BTP API changes

michal-narajowski:
These are the proposed changes and additions to the BTP specification. Some of these changes can be useful for autopts testing and others for general Bluetooth test automation

### DIFF
--- a/doc/btp_spec.txt
+++ b/doc/btp_spec.txt
@@ -296,6 +296,7 @@ Commands and responses:
 					Adv_Data (0-255 octets)
 					Scan_Rsp (0-255 octets)
 					Duration (4 octets)
+					Use own ID address (1 octet)
 		Return Parameters:	Current_Settings (4 Octets)
 
 		This command is used to start advertising.
@@ -328,6 +329,7 @@ Commands and responses:
 					2 = Use limited discovery procedure
 					3 = Use active scan type
 					4 = Use observation procedure
+					5 = Use own ID address
 
 		This command is used to start discovery.
 
@@ -355,6 +357,7 @@ Commands and responses:
 		Controller Index:	<controller id>
 		Command parameters:	Address_Type (1 octet)
 					Address (6 octets)
+					Use own ID address (1 octet)
 		Return Parameters:	<none>
 
 		Valid Address_Type parameter values:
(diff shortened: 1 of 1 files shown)


doc/btp_spec.txt
```diff
@@ -562,6 +562,24 @@ Events:
...
+	Opcode 0x89 - Security Level event
+
+		Controller Index:	<controller id>
+		Event parameters:	Address_Type (1 octet)
+					Address (6 octets)
+					Security Level (1 octet)
+
+		Possible values for the Security Level parameter map directly
+		to those defined in Core Specification in LE security mode 1:
+					1 = No security (No authentication and no encryption)
```
MariuszSkamra: I would not stick into Core Specification in this case. Receiving this event with "No security" level looks odd. Meaning this should be sent after each GAP Connected event. It would be better to send this event when the level changes only IMO.

doc/btp_spec.txt
```diff
@@ -550,6 +550,18 @@ Events:
...
+	Opcode 0x88 - Connection Parameters event
+
+		Controller Index:	<controller id>
+		Event parameters:	Address_Type (1 octet)
+					Address (6 octets)
+					Connection Interval (2 octets)
+					Connection Latency (2 octets)
+					Supervision Timeout (2 octets)
+
+		This event can be sent when a connection is established or
```
MariuszSkamra: The same situation. I would send this event on parameters change only. Sending another event after GAP Connected event is a bit odd. If you really need to know connection parameters of newly established connection, just extend the GAP Connected event with those.

doc/btp_spec.txt
```diff
@@ -295,10 +295,14 @@ Commands and responses:
 					Scan_Rsp_len (1 octet)
 					Adv_Data (0-255 octets)
 					Scan_Rsp (0-255 octets)
+					Duration (4 octets)
```
MariuszSkamra: Is this related to limited advertising? If so it would be good to refactor Opcode 0x08 - Set Discoverable command/response parameters in this PR as well:
"0x00 = Off
0x01 = General Discoverable
0x02 = Limited Discoverable"

so that :
"0x00 = Off
0x01 = On"

michal-narajowski:
I've added one more command: Connection Parameter Update.

PR #324, opened 2019-07-01
doc/btp_spec: Refactor GATT Server database initialization

MariuszSkamra:
The aim of this patch is to refactor the way the GATT Server
is initialized so that the API can be easily used by different
host implementations.
Instead of dynamically initialization, GATT attributes are
static defined in IUT tester application. It is up to
implementation to compose it's attribute database to allow
to test all of the features it supports.


MariuszSkamra:
@michal-narajowski FYI

PR #325, opened 2019-07-02
workspace: Add workspace for 7.4.1 based on BT spec 5.1

tedd-an:
This patch adds new workspace for 7.4.1 based on BT spec 5.1.


tedd-an:
Updated to the latest selections based on the discussion on #zephyr-sof@slack.

PR #332, opened 2019-08-01
Add missing GATT test cases

tedd-an:
This PR adds missing GATT test cases after updating the ICS to 1.14.


ptsprojects/zephyr/gatt_wid.py
```diff
@@ -140,26 +149,7 @@ def hdl_wid_17(desc):
...
-    for service in pts_services:
-        if service in iut_services:
-            iut_services.remove(service)
-            logging.debug("Service %s found", service)
-            continue
-        else:
-            logging.error("Service %s not found", service)
-            return False
-    return True
+    return btp.verify_desc_values(pts_services)
```
MariuszSkamra: I guess it could break the verification for server side test cases that use this WID.
GATT/GATT/SR/GPA/BV-01-C,
GATT/GATT/SR/GAD/BV-01-C

[1] I suggest you to add

def hdl_wid_17(desc):
    return btp.verify_description(desc)

to gattc_wid.py

ptsprojects/zephyr/gatt.py
```diff
@@ -549,6 +549,9 @@ def test_cases_client(pts):
         ZTestCase("GATT", "GATT/CL/GAC/BV-01-C",
                   pre_conditions,
                   generic_wid_hdl=gatt_wid_hdl),
+        ZTestCase("GATT", "GATT/CL/GAD/BV-01-C",
+                  pre_conditions,
+                  generic_wid_hdl=gatt_wid_hdl),
```
MariuszSkamra: use gattc_wid_hdl if you decide to [1]

pybtp/btp.py
```diff
@@ -282,6 +284,47 @@ def verify_description(description):
     return True
 
 
+def verify_desc_values(desc_values):
```
MariuszSkamra: btp.verify_description(desc) should be sufficient

pybtp/btp.py
```diff
@@ -2031,6 +2091,40 @@ def gattc_disc_prim_uuid_find_attrs_rsp(exp_svcs, store_attrs=False):
...
+    svcs_tuple = gatt_dec_disc_rsp(tuple_data[0], "service")
+    logging.debug("%s %r", gattc_disc_all_prim_rsp.__name__, svcs_tuple)
+
+    if store_rsp:
+        global VERIFY_VALUES
+
+        VERIFY_VALUES = []
+
+        for svc in svcs_tuple:
+            start_handle = "%04X" % (svc[0],)
```
MariuszSkamra: I've noticed PTS does not verify start, end handles so this is probably not needed

ptsprojects/zephyr/gatt_wid.py
```diff
@@ -1001,6 +1001,48 @@ def hdl_wid_107(desc):
...
+    if not uuid:
+        logging.debug("parsing error")
+        return False
+
+    btp.gattc_read_uuid(btp.pts_addr_type_get(None), btp.pts_addr_get(None),
+                        "0001", "FFFF", uuid)
+
+    try:
+        btp.gattc_read_uuid_rsp(False, True)
+    except socket.timeout:
```
MariuszSkamra: this looks like a bug on socket timeout you should probably return False

ptsprojects/zephyr/gatt_wid.py
```diff
@@ -1001,6 +1001,48 @@ def hdl_wid_107(desc):
...
```
MariuszSkamra: ditto

ptsprojects/zephyr/gap.py
```diff
@@ -229,6 +229,12 @@ def test_cases(pts):
         ZTestCase("GAP", "GAP/BROB/OBSV/BV-02-C",
                   cmds=pre_conditions,
                   generic_wid_hdl=gap_wid_hdl),
+        ZTestCase("GAP", "GAP/BROB/OBSV/BV-03-C",
```
MariuszSkamra: Those test cases are available on PTS workspace while not in TestPlan. so I would omit those. Do you see those in your test plan?

tedd-an: It is enabled in my TestPlan due to the supported spec is 5.1.
If we choose to use 5.0, these will be disabled.

PR #346, opened 2019-11-05
btp: fix advertisement/scan response construction

mringwal:
btp.gap_adv_ind_on constructs advertisement data as a list of { ad_type, ad_len, ad_data} chunks: https://github.com/intel/auto-pts/blob/master/pybtp/btp.py#L527. 

However, it should be { ad_len, ad_type, ad_data } according to Core V5.1, [Vol 3] Part C, Section 1.) 

I didn't see any comments in the btp_spec.txt which would indicate a different format for the BTP message.

### DIFF
--- a/pybtp/btp.py
+++ b/pybtp/btp.py
@@ -524,14 +524,14 @@ def gap_adv_ind_on(ad={}, sd={}):
 
     for ad_type, ad_data in ad.iteritems():
         data = binascii.unhexlify(bytearray(ad_data))
+        ad_ba.extend(chr(len(data)+1))
         ad_ba.extend(chr(ad_type))
-        ad_ba.extend(chr(len(data)))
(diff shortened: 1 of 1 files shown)


mringwal:
I just found the relevant code in the bluez btp:
https://kernel.googlesource.com/pub/scm/bluetooth/bluez/+/5.50/tools/btpclient.c#1062

which expects ad_type, ad_len, ad_data (which would be the better order).

So, this PR kind of morphes into an issue: either fix the order in btp.py + bluez + other btps, or, update the documentation. 

In going with "least number of surprises", I'd vote for updating btp.py and the BTP clients...

mringwal:
Anyway, to minimize change, I've created another pull request to update the documentation/btp-sped: https://github.com/intel/auto-pts/pull/347

PR #348, opened 2019-11-25
btp: use own_addr_type instead of use own address

mringwal:
Specifying own addr type allows to configure IUT for GAP/BROB/BCST/BV-04

### DIFF
--- a/doc/btp_spec.txt
+++ b/doc/btp_spec.txt
@@ -373,12 +373,13 @@ Commands and responses:
 		Controller Index:	<controller id>
 		Command parameters:	Address_Type (1 octet)
 					Address (6 octets)
-					Use own ID address (1 octet)
+					Own_Addr_Type (1 octet)
 		Return Parameters:	<none>
 
-		Valid Address_Type parameter values:
-					0x00 = Public
-					0x01 = Random
+		Valid Own_Addr_Type parameter values:
+					0x00 = Identity Address
+					0x01 = Resolvable Private Address
+					0x02 = Non-resolvabe Private Address
 
 		This command is used to create a Link Layer connection with
 		remote device.
--- a/doc/btp_spec.txt
+++ b/doc/btp_spec.txt
@@ -305,12 +305,17 @@ Commands and responses:
 					Adv_Data (0-255 octets)
 					Scan_Rsp (0-255 octets)
 					Duration (4 octets)
-					Use own ID address (1 octet)
+					Own_Addr_Type (1 octet)
 		Return Parameters:	Current_Settings (4 Octets)
+
+		Valid Own_Addr_Type parameter values:
+					0x00 = Identity Address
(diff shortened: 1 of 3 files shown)


pybtp/btp.py
```diff
@@ -507,7 +507,7 @@ def gap_wait_for_disconnection(timeout=30):
     stack.gap.wait_for_disconnection(timeout)
 
 
-def gap_adv_ind_on(ad={}, sd={}):
+def gap_adv_ind_on(ad={}, sd={}, duration=4294967295, own_addr_type=OwnAddrType.le_identity_address):
```
michal-narajowski: I would create a constant value for max duration and use it here.

mringwal: Agree. Where to add it? (I'm still new to the codebase). defs.py seems like the correct place for this, but it is autogenerated from zepyhr sources, types.py has mostly enum style items.

michal-narajowski: I think types.py is fine.

mringwal: Ok. I've added a AdDuration class with a forever constant.

pybtp/types.py
```diff
@@ -86,6 +86,8 @@ class AdFlags:
     sim_le_br_edr_contr = 0x08
     sim_le_br_edr_host = 0x10
 
+class AdDuration:
+    forever = 4294967295
```
michal-narajowski: Just a nitpick: you can use hexadecimal values in Python, so I think it would be more readable if the value was 0xFFFFFFFF

mringwal:
Good idea, thanks.  I'm mainly using C all day...

doc/btp_spec.txt
```diff
@@ -365,9 +365,10 @@ Commands and responses:
 					Use own ID address (1 octet)
```
michal-narajowski: You should update the parameter here

doc/btp_spec.txt
```diff
@@ -365,9 +365,10 @@ Commands and responses:
...
 		Return Parameters:	<none>
 
-		Valid Address_Type parameter values:
-					0x00 = Public
-					0x01 = Random
+		Valid Own_Addr_Type parameter values:
```
michal-narajowski: Address_Type is a different parameter, so just add a note about Own_Addr_Type below this.

mringwal: Thanks, I've missed that line.

mringwal:
I'm back. Could this be merged onto master? Is there anything missing? Thanks!

mringwal:
Hi again. I'll continue with the integration of our stack. Could this change be merged into master please? Anything missing? Thanks!

michal-narajowski (approved):
LGTM

michal-narajowski:
Hi @mringwal, thanks for updating the PR!

mringwal:
Thanks for review & merge.

PR #349, opened 2019-12-04
btp: add log message command

mringwal:
This command allows to insert log messages into whatever logging mechanism the BTP Client uses.

I've seen intermittent failures, e.g. an advertisement not received by PTS while the autoptsclient log shows sending the start adv command. While I have full log of the IUT, it contains the complete trace and it would be hard to find the corresponding section. With the BTP Log Message Command, the ptsproject can send the current test name and allow at least to find the revelant part of the log. I can imagine there might be other uses cases as well, that's why it's a general 'log message' command and not a 'set test name' command.


doc/btp_spec.txt
```diff
@@ -130,6 +130,14 @@ Commands and responses:
 
 		In case of an error, the error response will be returned.
 
+	Opcode 0x05 - Log message at IUT
+
+		Controller Index:	<non-controller>
+		Command parameters:	log message (variable)
```
MariuszSkamra: I would suggest to add "log_message_len" parameter, just like we have for other variable length parameters. We could think about log levels as well

mringwal: Sure. I've added a Log_Message_Len as in other commands.

For the use cases described above, I don't see a need for log levels. It could still be added after the log message if ever needed.

PR #350, opened 2019-12-06
Some common updates, fixes and update of Mynewt project to PTS 7.5.1

michal-narajowski:


michal-narajowski:
@MariuszSkamra Thanks, I added missing definitions to btp_spec.txt.

MariuszSkamra:
@tedd-an Do you have a time to run the tests with this PR?

michal-narajowski:
Most of these changes are Mynewt specific. There are only a few small changes that touch common code so the chance that they break something is small. Still it would be good if you could test it.

MariuszSkamra:
@michal-narajowski I think there is no point to hold this PR from being merged. Even if @tedd-an reports it broke something, it can be fixed in following PR

PR #357, opened 2020-02-21
zephyr/iutctl: Flush serial before board reset

MariuszSkamra:
This is needed to clean up any data left before board reset.
Any data left may affect the BTP commands flow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,9 @@ Over 460 test cases have been automated for Zephyr OS and Mynewt OS which reduce
 
 # Linux Prerequisites
 
-Install required modules with:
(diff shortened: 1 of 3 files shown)


joerchan:
pyserial is a new dependency.
This should be added to the requirements.txt

joerchan:
Tested and it works. Thank you.

joerchan:
Complete list of missing dependencies (that I ran into)
from apt: python-setuptools and socat,
from pip: wheel and pyserial.

MariuszSkamra:
@joerchan Dependencies updated. Thanks!

PR #365, opened 2020-02-25
More minor fixes for Zephyr

michal-narajowski:


bot/config.py.zephyr.sample
```diff
@@ -84,7 +84,7 @@ z['gdrive'] = {
 # ****************************************************************************
 
 z['iut_config'] = {
-    "default.conf": {},  # Default config file name
+    "prj.conf": {},  # Default config file name
```
carlescufi: This could be a bit of a problem, because the Zephyr build system picks up automatically `boards/<board>.conf`, and we rely on that now. So if you are overloading this, make sure it doesn't affect the system

michal-narajowski: `default.conf` was used before and it was renamed to prj.conf by `6de5c4b7b30c00e3b0e6c62a15de4548dc093596`. If this is a problem then we should change that in the near future, but it works for now and it fixes a bug where `default.conf` is missing.

PR #386, opened 2020-05-28
Mynewt: WID 81 correction

KKopyscinski:
Test was failing for odd values of val_mtp, because multiplied string was of length 1, and should be even. Switched to value '11' of length 2.


michal-narajowski:
Which test?

KKopyscinski:
> Which test?

GATT/CL/GAW/BI-34-C

michal-narajowski:
This is mynewt-specific so add a prefix to the PR title "mynewt:".

PR #387, opened 2020-06-01
btp: changed VERIFY_VALUES saving for gattc_read_uuid_rsp

KKopyscinski:
In wid 50 data was required to mach against PTS querry. Previously, that data was countinous string wich caused failed checks. Now, this data is split into characters (modified gattc_read_uuid_rsp).


michal-narajowski:
This PR is not mynewt specific. Also this change does not touch WID 50 so please fix the description so that it is more clear. Please specify which test it fixes.

PR #397, opened 2020-07-07
Some common and Zephyr specific updates

michal-narajowski:
Features improved sync mechanism that allows us to pass tests that require two PTS instances.


michal-narajowski:
FYI @trond-snekvik

PR #399, opened 2020-08-28
btp: fix core_log_message

mringwal:
Existing code causes a run time error with Python 2.7, where the result from struct.pack(..) is a string, which does not have an extend method. The new code does not work with Python 3.x. Any suggestion for a clean Python 2+3 compatible implementation?

### DIFF
--- a/pybtp/btp.py
+++ b/pybtp/btp.py
@@ -545,9 +545,7 @@ def core_unreg_svc_rsp_succ():
 def core_log_message(message):
     logging.debug("%s", core_log_message.__name__)
 
-    message_data = bytearray(message)
-    data = struct.pack('H', len(message_data))
-    data.extend(message_data)
+    data = bytearray( struct.pack('H', len(message)) + message)
 
     iutctl = get_iut()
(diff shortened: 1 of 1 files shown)


MariuszSkamra:
@KKopyscinski could you please look at this one?
We are about to move to Python 3, dropping the Python 2 support. There's patch https://github.com/intel/auto-pts/pull/373 that does the thing. We are close to merge it I guess.

KKopyscinski:
Because Python 3 changed a lot about handling bytes/bytearrays, and converting them to strings I think there is neither easy not clean way to build an implementation in Python 3 that is backward compatible.

mringwal:
I'm all for Python 3 :) This PR is more about running the current version on Master (with Python 2) until the move to Python 3 is complete.

KKopyscinski:
@mringwal Since Python 3 port was already merged I think this can be closed now - all byte/bytearray/string conversions seem to work fine

mringwal:
Agree. (we still need to update and switch to Python 3, but that's a different topic). Thanks!

PR #402, opened 2020-09-21
mesh.py: update_pixit_param for 2nd pts

KKopyscinski:
Second pts instance hasn't got updated pixit param.


MariuszSkamra (approved):
good catch!

PR #405, opened 2020-09-25
common: Use rtt2pty to automatically collect logs from device

michal-narajowski:
Collects logs from device for each test case automatically and puts
it in separate folder for easy access.


MariuszSkamra (changes requested):
RTT is something that can be not supported by board, thus maybe you should consider adding a flag that indicates RTT board support

michal-narajowski:
@MariuszSkamra Thanks, I fixed that by adding a command line argument.

MariuszSkamra (approved):
LGTM

PR #420, opened 2021-01-11
Mynewt: various mesh fixes

KKopyscinski:
Mynewt mesh sync caused some tests to fail, these commits patch these issues


wid/mesh.py
```diff
@@ -458,14 +461,19 @@ def hdl_wid_37(desc):
...
-    if not stack.mesh.last_seen_prov_link_state.data:
-        return False
+    flag = Event()
+    flag.set()
+
+    t = Timer(5, timeout_cb, [flag])
+    t.start()
+    while flag.is_set():
+        stack = get_stack()
+        state, bearer = stack.mesh.last_seen_prov_link_state.data
```
michal-narajowski, line 472: Move this logic to stack.py. It is used there in a couple of cases.

michal-narajowski (changes requested):

michal-narajowski (approved):

PR #421, opened 2021-01-12
wid: Minor updates and fixes

michal-narajowski:
Fix a couple of issues omitted when migrating to Python 3 and add a new WID implementation.


MariuszSkamra (approved):
LGTM

PR #423, opened 2021-01-12
btp: Add support for Friend/LPN events

michal-narajowski:


michal-narajowski:
@trond-snekvik FYI

doc/btp_spec.txt
```diff
@@ -1852,3 +1852,54 @@ Events:
 
 		This event indicates that segmented message incomplete timer
 		expired.
+
+	Opcode 0x89 - Friendship Established Event
```
MariuszSkamra, line 1856: Would it be possible to have common Established and Terminated events for Friend and LPN with one additional Role parameter?

michal-narajowski: It would be possible but I'm not sure if this is the way to go. How would we pass additional parameters that are specific to one role? And since this PR already has an LPN-specific event then I think it is cleaner to have separate events per role. I'm open to discussion :)

michal-narajowski:
@jhedberg  FYI

michal-narajowski:
Depends on https://github.com/intel/auto-pts/pull/423

doc/btp_spec.txt
MariuszSkamra: I agree, let's keep those as separate commands as the arguments differ

MariuszSkamra (approved):
LGTM

PR #424, opened 2021-01-13
ptscontrol: Update get_bluetooth_address API to PTS 8.0.0

michal-narajowski:
PTS since version 8.0.0 requires the automation framework to connect to
a dongle by calling a new API. This patch introduces this change to
auto-pts framework.

### DIFF
--- a/ptscontrol.py
+++ b/ptscontrol.py
@@ -748,8 +748,35 @@ class PyPTS:
 
     def get_bluetooth_address(self):
         """Returns PTS bluetooth address string"""
+        try:
+            address = self._pts.GetPTSBluetoothAddress()
+        except Exception as e:
+            address = ""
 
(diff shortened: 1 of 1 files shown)


ptscontrol.py
```diff
@@ -748,8 +748,35 @@ def save_test_history_log(self, save):
...
+            address = self._pts.GetPTSBluetoothAddress()
+        except Exception as e:
+            address = ""
 
-        return self._pts.GetPTSBluetoothAddress()
+        if address == "":
+            self.connect_to_dongle()
+            try:
+                address = self._pts.GetPTSBluetoothAddress()
+            except Exception as e:
```
MariuszSkamra, line 760: Looks like this will silently handled, maybe we should raise this exception instead if we failed to obtain the address?

MariuszSkamra (approved):
LGTM

Issue #52, opened 2016-03-03
Pending PTS issues


MariuszSkamra:
- GATT
- [ ] [ TC_GAN_SR_BV_01_C](https://www.bluetooth.org/pts/issues/view_issue.cfm?id=14182)
- [ ] [TC_GPA_SR_BV_11_C](https://www.bluetooth.org/pts/issues/view_issue.cfm?id=14437)
- [ ] [TC_GAR_SR_BV_08_C](https://www.bluetooth.org/pts/issues/view_issue.cfm?id=14416)
- [x] [TC_GAW_CL_BI_07_C
      TC_GAW_CL_BI_08_C
      TC_GAW_CL_BI_14_C
      TC_GAW_CL_BI_15_C
      TC_GAW_CL_BI_25_C
      TC_GAW_CL_BI_26_C](https://www.bluetooth.org/pts/issues/view_issue.cfm?id=14328)
- [ ] [TC_GAD_SR_BV_03_C](https://www.bluetooth.org/pts/issues/view_issue.cfm?id=14418)
- [ ] [TC_GAR_SR_BI_27_C
  TC_GAR_SR_BI_31_C
  TC_GAR_SR_BV_07_C
  TC_GAR_SR_BV_08_C
  TC_GAW_SR_BI_05_C
  TC_GAW_SR_BI_06_C
  TC_GAW_SR_BI_22_C
  TC_GAW_SR_BI_27_C
  TC_GAW_SR_BI_29_C
  TC_GAW_SR_BI_30_C
  TC_GAW_SR_BI_32_C
  TC_GAW_SR_BI_33_C
  TC_GAW_SR_BI_35_C
  TC_GAW_SR_BV_01_C
  TC_GAW_SR_BV_05_C
  TC_GAW_SR_BV_06_C
  TC_GAW_SR_BV_07_C
  TC_GAW_SR_BV_09_C
  TC_GAW_SR_BV_10_C](https://www.bluetooth.org/pts/issues/view_issue.cfm?id=14428)
- GAP
- [ ] [TC_BOND_BON_BV_04_C](https://www.bluetooth.org/pts/issues/view_issue.cfm?id=14569)
- [ ] [TC_BOND_BON_BV_02_C](https://www.bluetooth.org/pts/issues/view_issue.cfm?id=14449)
- [ ] [TC_BOND_BON_BV_01_C](https://www.bluetooth.org/pts/issues/view_issue.cfm?id=14444)
[...]

PR #426, opened 2021-01-20
wid: Fix GAP WID #52 that IUT is in Undirected Connectable Mode

ryanjh:
It is used by GAP/DISC/GENM/BV-04-C.

### DIFF
--- a/ptsprojects/zephyr/gap_wid.py
+++ b/ptsprojects/zephyr/gap_wid.py
@@ -216,7 +216,14 @@ def hdl_wid_51(desc):
 
 
 def hdl_wid_52(desc):
-    hdl_wid_51(desc)
+    stack = get_stack()
+
+    btp.gap_adv_off()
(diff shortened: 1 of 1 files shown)


ptsprojects/zephyr/gap_wid.py
```diff
@@ -216,7 +216,14 @@ def hdl_wid_51(desc):
 
 
 def hdl_wid_52(desc):
-    hdl_wid_51(desc)
+    stack = get_stack()
+
+    btp.gap_adv_off()
```
michal-narajowski, line 221: Is it necessary to disable advertising here?

ryanjh: yes, similar to hdl_wid_90. The new configuration is taken by restarting advertisement.

michal-narajowski (approved):
LGTM

Thank you for your contribution @ryanjh!

PR #428, opened 2021-01-29
Zephyr: multi device support for Nordic devices

AyturkDuzen:


ptsprojects/zephyr/iutctl.py
```diff
@@ -312,7 +314,7 @@ def _get_reset_cmd_nrf52(self):
         Dependency: nRF5x command line tools
 
         """
-        return 'nrfjprog -f nrf52 -r'
+        return 'nrfjprog -f nrf52 -s {} -r'.format(BotProjects[0]['auto_pts']['board_id'])
```
michal-narajowski: I don't like this change, `BotProjects[0]` can be something else than Zephyr's config. I would prefer to add another parameter to ZephyrCtl.

bot/zephyr.py
```diff
@@ -111,7 +112,7 @@ def build_and_flash(zephyr_wd, board, board_id, conf_file=None):
     cmd_flash = ['west', 'flash', '--snr', board_id]
     check_call(cmd_flash, env=env, cwd=tester_dir)
 
-    return get_tty_path("J-Link")
+    return get_tty_path(board, board_id)
```
michal-narajowski, line 115: You could just use Serial number as a string here instead of "J-Link" and it should find the correct device.

README.md
```diff
@@ -130,6 +130,7 @@ This may contain few sections:
     - `project_path` - path to project source directory
     - `workspace` - PTS workspace path to be used
     - `board` - IUT used. Currently nrf52 is supported only
+    - `board_id` - Segger ID of the IUT used.
```
michal-narajowski, line 133: Is this the serial number of the device?

bot/zephyr.py
```diff
@@ -87,26 +87,29 @@ def source_zephyr_env(zephyr_wd):
...
-    cmd = ['west',  'build', '-p', 'auto', '-b', board]
+    cmd_build = ['west',  'build', '-p', 'auto', '-b', board]
     if conf_file:
-        cmd.extend(('--', '-DCONF_FILE={}'.format(conf_file)))
+        cmd_build.extend(('--', '-DCONF_FILE={}'.format(conf_file)))
+    check_call(cmd_build, env=env, cwd=tester_dir)
 
-    check_call(cmd, env=env, cwd=tester_dir)
-    check_call(['west', 'flash'], env=env, cwd=tester_dir)
+    cmd_flash = ['west', 'flash', '--snr', board_id]
```
michal-narajowski, line 112: What if the `board_id` is empty?

michal-narajowski (changes requested):
I like the idea but we need to remember to not break anything and I think some of these changes can be simplified if you just use the device serial number as `board_id`.

ptsprojects/zephyr/iutctl.py
AyturkDuzen: agree, good point.

bot/zephyr.py
AyturkDuzen: I think board_id is already passing the serial number parameter here

README.md
AyturkDuzen: Yes I will clarify the naming. "Serial number of the IUT used" more proper?

bot/zephyr.py
AyturkDuzen: before changes (when it was only west flash) and no device attached. It should return the same error.

This case when there is more than one device connected, it should find the correct one to flash.

Do you want to add control if board_id is empty?

bot/zephyr.py
michal-narajowski: What I mean here is: if board_id is set then pass it to `get_tty_path()` instead of `"J-Link"`. If not, then keep old behavior.

bot/zephyr.py
michal-narajowski: Yes, I want to keep old behavior if `board_id` is empty.

AyturkDuzen: updated the PR also covering the empty cases

AyturkDuzen:
We are dividing the PR to several PRs since it has different upgrades with different titles, so closing this.

PR #430, opened 2021-02-26
wid: Fix L2CAP WID_43 and WID_38 for PTS 8.0.0

ryanjh:
They are used by L2CAP/COS/CFC/BV-01-C and L2CAP/COS/CFC/BV-02-C.


michal-narajowski (approved):
LGTM

PR #433, opened 2021-03-05
mynewt/sm: change testcase names to PTS 8 versions

KKopyscinski:
PTS 8 changed testcase names to use inclusive language


michal-narajowski (approved):
LGTM

PR #439, opened 2021-03-19
Mynewt: fix GAP wid 157

KKopyscinski:
Wid 157 asks to check if advertising report and scan response were received by IUT. Previously,
we checked if addr type and addr fit PTS address. PTS 8.0.1 uses random address, so these values
will not be valid. Instead, we shall parse scan report and response data and check if it was
received from any of discovered devices.


KKopyscinski:
This allows GAP/BROB/OBSV/BV-05-C to pass

michal-narajowski (approved):
LGTM

PR #440, opened 2021-03-23
PTS 8.0.1 Mynewt fixes

KKopyscinski:
Tests that cannot be fixed for now are:
- GAP/SEC/SEM/BV-29-C - PTS errata CASE0070978 - first alternative of test ending is not working
- L2CAP/ECFC/BV-24-C - TSE 15582 - Test demands reconfiguration of MPS, which has static size in Mynewt (as it's not required to be dynamic as per Core spec)

### DIFF
--- a/ptsprojects/mynewt/gap_wid.py
+++ b/ptsprojects/mynewt/gap_wid.py
@@ -57,6 +57,15 @@ def hdl_wid_5(desc):
     return True
 
 
+def hdl_wid_9(desc):
+    stack = get_stack()
+
+    btp.gap_set_conn()
+
+    btp.gap_adv_ind_on(ad=stack.gap.ad, sd=stack.gap.sd)
+    return True
+
+
 def hdl_wid_10(desc):
     btp.gap_stop_discov()
     return btp.check_discov_results(discovered=True)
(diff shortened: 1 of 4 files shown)


KKopyscinski:
@michal-narajowski Please review these changes

bot/config.py.mynewt.sample
```diff
@@ -102,6 +102,15 @@ m['iut_config'] = {
         ],
     },
 
+      "mtu_400.conf": {
+        "overlay": {
+            'BTTESTER_L2CAP_COC_MTU': '400',
```
michal-narajowski, line 107: Can we change MTU for all test cases or is this one special?

michal-narajowski (approved):
LGTM

bot/config.py.mynewt.sample
KKopyscinski: I investigated this test a little more and there's an issue of how the WID describes what should be done. This test checks if credits are increasing when IUT sends SDUs of size equal to peer (PTS) MPS. The wording of WID is: "Upper Tester command IUT to send at least 4 frames of LE data packets to the PTS." It caused confusion, same as in PTS CASE0070320, that l2cap_send_data should be called with data of size 4*peer_mps, but in fact, we just need to send SDU of size peer_mps. Then, PTS will ask for sending more such SDUs, and will ask it 3 more times.

What's more, PTS doesn't follow MPS value from IXIT in this test. In conclusion, MTU might stay as it is, I will change WID 57 and test will pass

michal-narajowski (approved):

PR #441, opened 2021-03-24
btp: fix for passkeys with leading zeros

KKopyscinski:
Sometimes passkey begins with one or more zeros. During unpacking,
this information is lost, as passkey in bytes packet is converted to
int. Now, passkey is converted to string and lost zeros are added.
This prevents returning invalid passkey to PTS with len <6.

### DIFF
--- a/pybtp/btp.py
+++ b/pybtp/btp.py
@@ -3351,6 +3351,10 @@ def gap_passkey_disp_ev_(gap, data, data_len):
     addr_type, addr, passkey = struct.unpack(fmt, data)
     addr = binascii.hexlify(addr[::-1])
 
+    # unpacking passkey to int loses leading 0s,
+    # let's add them back if lost
+    passkey = str(passkey).zfill(6)
+
     logging.debug("passkey = %r", passkey)
 
     gap.passkey.data = passkey


pybtp/btp.py
```diff
@@ -3353,6 +3353,13 @@ def gap_passkey_disp_ev_(gap, data, data_len):
 
     logging.debug("passkey = %r", passkey)
 
+    passkey = str(passkey)
+    # unpacking passkey to int loses leading 0s,
+    # let's add them back if lost
+    pk_len = len(passkey)
```
michal-narajowski, line 3359: https://docs.python.org/2/library/stdtypes.html#str.zfill

KKopyscinski: now it uses `zfill()`

pybtp/btp.py
```diff
@@ -3353,6 +3353,10 @@ def gap_passkey_disp_ev_(gap, data, data_len):
 
     logging.debug("passkey = %r", passkey)
 
+    # unpacking passkey to int loses leading 0s,
+    # let's add them back if lost
+    passkey = str(passkey).zfill(6)
```
michal-narajowski, line 3358: Let's move debug log after this operation as the result will be actually used later.

KKopyscinski: I agree, done

michal-narajowski (approved):
LGTM

michal-narajowski (approved):

PR #448, opened 2021-05-06
mynewt: adjust timings in MESH/NODE/CFG/HBP/BV-05-C

KKopyscinski:


ptsprojects/mynewt/mesh.py
```diff
@@ -42,8 +42,8 @@
 from time import sleep
 
 
-device_uuid = hexlify(uuid4().bytes)
-device_uuid2 = hexlify(uuid4().bytes)
+device_uuid = "00000000000000000000000000000000"
```
michal-narajowski, line 45: Why is this change here?

michal-narajowski (changes requested):

ptsprojects/mynewt/mesh.py
KKopyscinski: Is there a reason that for Mynewt these are random?

michal-narajowski (approved):

PR #449, opened 2021-05-10
Zephyr: L2CAP/LE/CFC/BV fixes

KKopyscinski:
This commit fixes and/or adds tests expecting Insufficient
: Authentication/Authrization/Key Size errors on L2CAP connect request

### DIFF
--- a/ptsprojects/zephyr/l2cap.py
+++ b/ptsprojects/zephyr/l2cap.py
@@ -37,6 +37,7 @@ from wid import l2cap_wid_hdl
 le_psm = 128
 psm_unsupported = 241
 psm_authentication_required = 242
+psm_authorization_required = 243
 psm_encryption_key_size_required = 244
 le_initial_mtu = 120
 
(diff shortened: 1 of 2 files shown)


ptsprojects/zephyr/l2cap.py
```diff
@@ -146,6 +147,40 @@ def test_cases(pts):
                            TestFunc(stack.l2cap_init, psm_authentication_required, le_initial_mtu),
                            TestFunc(btp.l2cap_le_listen, psm_authentication_required)]
 
+    pre_conditions_keysize = [TestFunc(btp.core_reg_svc_gap),
```
michal-narajowski, line 150: Why not use a common pre_conditions list and combine it with some specific list for a set of test cases?

`pre_conditions_key_size = pre_conditions + [... something specific to a set of test cases ...]`

michal-narajowski (approved):
LGTM with a minor question

ptsprojects/zephyr/l2cap.py
KKopyscinski: Ok, because pre_conditions_auth existed before, I added additional commit that simplifies this

michal-narajowski (approved):

PR #453, opened 2021-05-26
var_store_get_wrong_passkey: convert pk to int before comparison

KKopyscinski:
gap_passkey_disp_ev_ saves passkey as string so we need to convert it
to int before comparing to int.


KKopyscinski:
This affects SM/CEN/PKE/BI-02-C

PTAL @michal-narajowski

michal-narajowski (approved):

PR #454, opened 2021-05-26
Mynewt: update for PTS v8.0.3

KKopyscinski:
Updated ICS with supported and mandatory erratum, added new MESH test, which is passing


KKopyscinski:
@michal-narajowski PTAL

michal-narajowski (approved):

PR #455, opened 2021-05-26
common: Add Mesh Node Added event

michal-narajowski:


MariuszSkamra (changes requested):
please update BTP specification document

MariuszSkamra (approved):
looks good

PR #464, opened 2021-06-09
[Do not merge] Mynewt: add missing recovery arguments

KKopyscinski:
Without these tests won't run and client script will end up with error.

### DIFF
--- a/autoptsclient-mynewt.py
+++ b/autoptsclient-mynewt.py
@@ -69,6 +69,18 @@ def parse_args():
                                  "Requires rtt2pty tool and rtt support on IUT.",
                             action='store_true', default=False)
 
+    arg_parser.add_argument("--recovery", action='store_true', default=False,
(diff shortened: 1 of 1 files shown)


KKopyscinski:
@michal-narajowski PTAL
@mkasenberg Is this all that's required for recovery support?

KKopyscinski:
As discussed offline, this can exist here as workaround for now, but in near future `autoptsclient-{os}.py` files will be joined together into one client, which will solve the issue in proper way

mringwal:
> As discussed offline, this can exist here as workaround for now, but in near future `autoptsclient-{os}.py` files will be joined together into one client, which will solve the issue in proper way

I've also run into this and added the new arguments to our fork. Unifying the autptsclient implementations would be great.

michal-narajowski:
Resolved by https://github.com/intel/auto-pts/pull/474

PR #483, opened 2021-07-01
common: Do not run recovery steps when INCONC or FAIL

mkasenberg:
as these are valid PTS results, not caused by problems with connection
to server, like XML-RPC ERROR.
(--recovery option feature)


MariuszSkamra (approved):
good catch! seems logical ^^

michal-narajowski (approved):
LGTM

PR #489, opened 2021-07-05
Zephyr: fix GATT/SR/GAW/BV-{12, 13}-C

KKopyscinski:


KKopyscinski:
GATT/SR/GAW/BV-14-C requires access to EATT channels with EATT enabled to disconnect one of them, so it can't be easily fixed

michal-narajowski (approved):

PR #490, opened 2021-07-05
bot: autoptsserver: Add uploading BPV logs to GDrive

mkasenberg:
When the gdrive option is enabled on bot, after tests, bot client
will send to autopts server the signal to upload collected Bluetooth
Protocol Viewer logs to the specified folder on the GDrive.

Add default client and server ports at parsing args.

### DIFF
--- a/autoptsserver.py
+++ b/autoptsserver.py
@@ -30,8 +30,8 @@
 
 import argparse
 import copy
-import datetime
 import os
+import shutil
 import subprocess
 import threading
 import time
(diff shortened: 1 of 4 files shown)


michal-narajowski (approved):

michal-narajowski:
It works fine, although I'm not a big fan of adding features to the server (I would like to keep it as lean as possible). But if you think pushing these logs to the client before the client can upload them to GDrive is impractical then I think we can accept that and keep this behavior instead.

mkasenberg:
A client running on Windows could send these logs by itself, but for client on Linux I couldn't think of anything other than ssh which requires more configuration. So it is probably the easiest solution, not the best, I agree.

mkasenberg:
Update: I used xmlrpc to send the logs from auto-pts server to client, so finally no additional setup is needed on server side.

michal-narajowski (approved):

PR #495, opened 2021-07-06
Generate unique names for IUT

KKopyscinski:
When 2 or more IUTs are under test at the same time and they share the
name, PTS might choose wrong board and connect to it. Now, last 3 parts
of PTS dongle address are appended to the end of IUTs' name to
distinguish them from each other.

This affects only tests that rely on IXIT
"TSPX_iut_device_name_in_adv_packet_for_random_address", as only then
this name is actually used. This means that default device name stays
the same as "Tester" for GATT (uses static address), L2CAP and MESH
test suites.

### DIFF
--- a/ptsprojects/mynewt/mesh.py
+++ b/ptsprojects/mynewt/mesh.py
@@ -40,6 +40,7 @@ from uuid import uuid4
 from binascii import hexlify
 import random
 from time import sleep
+from autoptsclient_common import get_unique_name
 
 
 device_uuid = hexlify(uuid4().bytes)
@@ -56,6 +57,9 @@ def set_pixits(ptses):
     ptses -- list of PyPTS instances"""
 
     pts = ptses[0]
+
+    global iut_device_name
+    iut_device_name = get_unique_name(pts)
 
     pts.set_pixit("MESH", "TSPX_bd_addr_iut", "DEADBEEFDEAD")
     pts.set_pixit("MESH", "TSPX_bd_addr_additional_whitelist", "")
@@ -166,7 +170,7 @@ def test_cases(ptses):
     rand_in_actions = random.choice(in_actions) if in_size else 0
     crpl_size = 10  # Maximum capacity of the replay protection list
 
-    stack.gap_init()
+    stack.gap_init(iut_device_name)
     stack.mesh_init(device_uuid, oob, out_size, rand_out_actions, in_size,
                     rand_in_actions, crpl_size)
 
--- a/ptsprojects/mynewt/sm.py
+++ b/ptsprojects/mynewt/sm.py
@@ -30,6 +30,7 @@ except ImportError:  # running this module as script
 
 from pybtp import btp
 from pybtp.types import Addr, IOCap
+from autoptsclient_common import get_unique_name
 from ptsprojects.stack import get_stack
 from .sm_wid import sm_wid_hdl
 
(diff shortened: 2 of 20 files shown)


sjanc (commented):
Why not make name always unique?

KKopyscinski:
@sjanc Now names are always unique :) I left `TSPX_iut_device_name_in_adv_packet_for_random_address` unfilled, where it was as such before, since in these cases this IXIT is unused (for example static address is used). Previous solution would work either way, since PTS would search only for unique names, ignoring plain `Tester` ones. But this change allowed me to fix overwriting unique name with default one (calls of `btp.gap_init()` without arguments)

ptsprojects/mynewt/mesh.py
```diff
@@ -40,6 +40,10 @@
 from binascii import hexlify
 import random
 from time import sleep
+from autoptsclient_common import get_unique_name
+
+
+iut_device_name = 'Tester'.encode()
```
michal-narajowski, line 46: I think this should be removed in favor of new get_unique_name(), no?

ptsprojects/mynewt/sm.py
```diff
@@ -62,15 +63,21 @@ def test_cases(pts):
 
     pts_bd_addr = pts.q_bd_addr
 
+    iut_device_name = get_unique_name(pts)
+
     stack = get_stack()
 
     stack.gap_init()
```
michal-narajowski, line 70: You are using iut_device_name in other *profile.py* files but not in here?

ptsprojects/zephyr/dis.py
```diff
@@ -75,6 +85,8 @@ def set_pixits(pts):
 
     pts -- Instance of PyPTS"""
 
+    global iut_device_name
```
michal-narajowski, line 88: Why here and not above?

ptsprojects/zephyr/gap.py
```diff
@@ -109,6 +111,9 @@ def set_pixits(pts):
 
     pts -- Instance of PyPTS"""
 
+    global iut_device_name
```
michal-narajowski, line 114: Ditto.

ptsprojects/zephyr/gatt.py
```diff
@@ -120,6 +124,9 @@ def set_pixits(ptses):
 
     pts = ptses[0]
 
+    global iut_device_name
```
michal-narajowski, line 127: Ditto

ptsprojects/zephyr/l2cap.py
```diff
@@ -52,6 +55,9 @@ def set_pixits(pts):
 
     pts -- Instance of PyPTS"""
 
+    global iut_device_name
```
michal-narajowski, line 58: Ditto.

michal-narajowski (changes requested):
Looks good, just a couple of nit-picks.

ptsprojects/mynewt/mesh.py
KKopyscinski: Yes, these indeed can be remove. I also found a bug in gap.py with calculating name length: it was in decimal format, not hex. Then, when PTS was checking packet payload it was discarding it, since 10 is not 0a. I included this in this commit, since it only occurs with longer names, when ad_str_name_short_len>=10. I also removed unused code that relied on iut_device_name being outside `set_pixits()`

ptsprojects/mynewt/sm.py
michal-narajowski: Any response here?

ptsprojects/zephyr/dis.py
michal-narajowski: Any updates?

michal-narajowski:
Needs an update @KKopyscinski

KKopyscinski:
@michal-narajowski Rebased onto master

KKopyscinski:
@michal-narajowski Deleted unused code breaking GAP in Mynewt, everything works now after rebase

michal-narajowski (approved):

PR #497, opened 2021-07-07
Fix waiting for socat process to stop

sjanc:
When stopping client should wait for socat process to terminate.
Otherwise we may end up with connection error on next start if
process scheduling is unlucky (seems to happen more often on Windows).
This also allows to get rid of sleep(2) while closing BTP socket saving
4 seconds of every test run.


sjanc:
@mkasenberg  FYI

michal-narajowski (approved):
LGTM

mkasenberg (approved):

PR #515, opened 2021-07-15
Zephyr: fix GATT/SR/GAS/BV-01-C

KKopyscinski:
We need to add two different services, so randomize UUID to avoid
duplication

### DIFF
--- a/ptsprojects/zephyr/gatt.py
+++ b/ptsprojects/zephyr/gatt.py
@@ -510,16 +510,13 @@ def test_cases_server(ptses):
         # TODO rewrite GATT/SR/GAS/BV-01-C
         ZTestCase("GATT", "GATT/SR/GAS/BV-01-C",
                   cmds=pre_conditions +
-                  [TestFunc(btp.gap_set_io_cap, IOCap.display_only),
-
-                   # Service Changed is triggered for bonded devices only
-                   TestFunc(btp.gap_wait_for_connection,
-                            post_wid=1, skip_call=(2,)),
-                   TestFunc(btp.gap_pair, post_wid=1, skip_call=(2,)),
-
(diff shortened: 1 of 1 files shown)


KKopyscinski:
This test also required patched `.ets` file from CASE0071492

sjanc:
hmm could we make it predictable?  eg toggle registration of two known UUIDs?

KKopyscinski:
Well, maybe add flag marking if wid 96 was run before? And register different UUID second time

KKopyscinski:
The problem is that we do not register it in wid 96, we do it after it returns, that's why I made it this way. So we are limited to what we can do using `TestFunc()`

KKopyscinski:
OK, we just need to register the same TestFunc twice, with different UUIDs, no need for randomization :)

sjanc (approved):

ptsprojects/zephyr/gatt.py
```diff
@@ -35,6 +35,7 @@
 from ptsprojects.stack import get_stack
 from ptsprojects.zephyr.gatt_wid import gatt_wid_hdl, gatt_wid_hdl_no_write_rsp_check
 from ptsprojects.zephyr.gattc_wid import gattc_wid_hdl
+from random import randint
```
sjanc, line 38: so this is no longer needed, tight?

sjanc (changes requested):

michal-narajowski (changes requested):
Also needs a rebase and conflicts resolved.

KKopyscinski:
@michal-narajowski Fixed and rebased

sjanc (approved):

michal-narajowski (approved):

PR #517, opened 2021-07-16
common: Unify the pixits setting

mkasenberg:
and add re-setting them up after recovery.

### DIFF
--- a/autoptsclient-zephyr.py
+++ b/autoptsclient-zephyr.py
@@ -56,15 +56,6 @@ class ZephyrClient(autoptsclient.Client):
     def init_iutctl(self, args):
         autoprojects.iutctl.init(args)
 
-    def setup_test_cases(self, ptses):
-        self.test_cases = autoprojects.gap.test_cases(ptses[0])
-        self.test_cases += autoprojects.dis.test_cases(ptses)
-        self.test_cases += autoprojects.gatt.test_cases(ptses)
-        self.test_cases += autoprojects.sm.test_cases(ptses[0])
-        self.test_cases += autoprojects.l2cap.test_cases(ptses[0])
-        self.test_cases += autoprojects.mesh.test_cases(ptses)
-        self.test_cases += autoprojects.mmdl.test_cases(ptses)
-
     def cleanup(self):
         autoprojects.iutctl.cleanup()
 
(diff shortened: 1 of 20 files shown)


autoptsclient_common.py
```diff
@@ -1294,6 +1298,37 @@ def run_recovery(args, ptses):
     except:
         traceback.print_exc()
 
+    try:
```
michal-narajowski, line 1301: Do we need this try/except here? If there are issues with a particular IXIT then we should handle it individually and let other IXIT values be set. Ideally, we would not cause exceptions in PTS, but just return an error from autopts server.

autoptsclient-zephyr.py
```diff
@@ -78,7 +69,7 @@ def cleanup(self):
 def main():
     """Main."""
 
-    client = ZephyrClient(get_iut, "zephyr_")
+    client = ZephyrClient(get_iut, "zephyr")
```
michal-narajowski, line 72: ZephyrClient constructor should pass "zephyr" parameter. Same with Mynewt and Bluez.

autoptsclient-zephyr.py
```diff
@@ -53,15 +53,6 @@ def check_args(self, args):
...
-        autoprojects.gap.set_pixits(ptses[0])
-        autoprojects.dis.set_pixits(ptses[0])
-        autoprojects.sm.set_pixits(ptses[0])
-        autoprojects.l2cap.set_pixits(ptses[0])
-        autoprojects.gatt.set_pixits(ptses)
-        autoprojects.mesh.set_pixits(ptses)
-        autoprojects.mmdl.set_pixits(ptses)
-
     def setup_test_cases(self, ptses):
         self.test_cases = autoprojects.gap.test_cases(ptses[0])
```
michal-narajowski, line 57: You could always pass all PTS instances to `profile.test_cases()`
Also, I think ideally you wouldn't need this setup_test_cases in here but in a common client.

michal-narajowski: But maybe that could be done in a next PR.

michal-narajowski: Also see if we can remove `kernel_image` argument.

michal-narajowski (changes requested):

autoptsclient_common.py
mkasenberg: No, we don't actually need it :) After reinit_pts above is called, setup_project_pixits(ptses) should work fine.

autoptsclient-zephyr.py
mkasenberg: Sure, I'll see what we can do :+1:

michal-narajowski (approved):
LGTM

PR #524, opened 2021-07-22
zephyr: mynewt: Change default test case instance for l2cap

mkasenberg:
Previous default instance covered only L2CAP/ECFC tests. It was
chosen, because they occurred most often. We can add second
default instance for the rest of tests.


KKopyscinski:
LGTM

michal-narajowski (approved):
LGTM

PR #526, opened 2021-07-22
zephyr: Make kernel_image arg optional in HW mode

mkasenberg:
as in this mode it is not needed.

### DIFF
--- a/autoptsclient-zephyr.py
+++ b/autoptsclient-zephyr.py
@@ -33,11 +33,6 @@ class ZephyrClient(autoptsclient.Client):
         arg_parser = autoptsclient.CliParser("PTS automation client",
                                              autoprojects.iutctl.Board.names)
 
-        # IUT specific arguments below
-        arg_parser.add_argument("kernel_image",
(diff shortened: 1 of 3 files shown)


mkasenberg:
Do we support arduino_101 and c1000 boards? It's looks like those functions _get_reset_cmd_arduino_101 and _get_reset_cmd_c1000 from ptsprojects/zephyr/iutctl.py use kernel_image, but only to obtain some path in zephyr repo. I would directly pass zephyr repo path or remove them.
`openocd_cfg = os.path.join(
            os.path.split(self.kernel_image)[0],
            "../../../../../../boards/x86/arduino_101/support/openocd.cfg")`

mkasenberg:
It seems that support for arduino_101 and c1000 was removed in Zephyr 2.0.0 https://docs.zephyrproject.org/latest/releases/release-notes-2.0.html#boards-soc-support

michal-narajowski (approved):
LGTM

sjanc (approved):

PR #529, opened 2021-07-27
bot: Add errata.yaml

mkasenberg:
If test case awaits PTS errata, its name can be listed together with case
number in errata.yaml file, so that an appropriate note can be added to
the reports.


sjanc (approved):
initially I thought we could have separate errata file per IUT, but for now lets keep all in one place (should be easier to maintain)

michal-narajowski (approved):

PR #531, opened 2021-08-02
bot: Fix NoneType error caused by empty errata.yaml

mkasenberg:


MariuszSkamra (approved):

michal-narajowski (approved):
We should also handle exceptions when the file does not exist.

PR #538, opened 2021-08-06
Fix qemu or native posix l2cap conn params update

LingaoM:
Test case L2CAP/LE/CPU/BV-01-C require IUT send L2CAP Connection Parameter
Update Request, Although we use `CONFIG_BT_CTLR_CONN_PARAM_REQ` to turn off
the controller function, but it is not feasible for native posix or qemu.

Signed-off-by: Lingao Meng <menglingao@xiaomi.com>


sjanc:
I think this is now properly handled by PTS (ie lower tester properly disables support for LL conn update and thus IUT can use L2CAP based update without abusing specification requirements)

Feel free to reopen if you stil see this problem.

PR #541, opened 2021-08-10
Zephyr test fixes

KKopyscinski:


ptsprojects/zephyr/gatt.py
```diff
@@ -64,6 +64,29 @@ def __get_attr_u16_hdl_uc_str(offset):
     return '{0:04x}'.format(iut_attr_db_off + offset, 'X')
 
 
+def __add_unique_char_to_service(i=[0]):
```
michal-narajowski, line 67: That's a nice hack but in general, we want to move away from global variables, so let's use a Stack object here to store the service UUID. This way it will be more readable and testable.

pybtp/btp/btp.py
```diff
@@ -71,6 +71,26 @@
 }
 
 
+def clear_verify_values():
```
michal-narajowski, line 74: Since you are changing it, let's use the Stack object to hold these values.

michal-narajowski (changes requested):
Some suggestions on how to solve the unique service problem.

KKopyscinski:
Rebased onto master

pybtp/btp/gatt.py
```diff
@@ -1369,18 +1363,18 @@ def gattc_disc_prim_uuid_rsp(store_rsp=False):
...
+            add_to_verify_values(end_handle)
 
             # avoid repeated service uuid, it should be verified only once, for
             # example:
             # gattc_disc_prim_uuid_rsp ((1, 3, ('\xc9N',)),
             # (48, 50, ('\xc9N',)), (64, 66, ('\xc9N',)),
             # (80, 82, ('\xc9N',)), (144, 150, ('\xc9N',)))
-            if uuid not in VERIFY_VALUES:
-                VERIFY_VALUES.append(uuid)
+            if uuid not in get_verify_values:
```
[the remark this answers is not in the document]
KKopyscinski: done

ptsprojects/zephyr/gatt.py
```diff
@@ -64,6 +64,29 @@ def __get_attr_u16_hdl_uc_str(offset):
...
+    # this is for internal use only. Mutable function argument allows
+    # to store it's value between funcion calls. This is similar to
+    # static variables in C
+    # Get value between 0x3000 and 0x9999 to avoid any 16 bit UUIDs
+    # from BT SIG Assigned Number
+    svc = 12288 + i[0]
```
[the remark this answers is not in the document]
KKopyscinski: switched to hex

ptsprojects/zephyr/gatt.py
KKopyscinski: Switched to `stack.gatt` object

michal-narajowski (approved):

PR #542, opened 2021-08-16
 autoptsclient autoptsserver: Fix start/end run behavior

mkasenberg:
Fixes ctrl+c behaviour under Windows by setting global RUN_END flag, only for simple client for now.

Fixes #513 issue with restarting PTS instances. Thread lock prevents restart of PTSes at the same time, so they won't steal each others dongles.

Fixes logging of multiple autoptsserver threads. Logs for second instance was never created. Now they share one log file and logs are distinguished by thread name column.

PR #539, opened 2021-08-06
bot: common: Unify bot clients

mkasenberg:
Bot client classes extends simple client class, so console input args now enabled for bot and they can override config.py settings.

PR #543, opened 2021-08-17
common: Fix shutting down after Ctrl+C interruption

mkasenberg:
Needed after removing os._exit() at end of run, so console won't hang because of unclosed server thread. The issue occurred mainly under Windows. Shutdown takes now around 10-15 seconds.


mkasenberg:
Fixed for simple client in #542. Solution for bot will be easier to handle in #539.

PR #547, opened 2021-08-18
common: Fix scheduler action state set

agata-ponitka:


michal-narajowski (approved):
Let's merge after https://github.com/nrfconnect/sdk-nrf/pull/5329 gets merged.

PR #551, opened 2021-08-31
tools: Add ICS html-to-rst file parser

mkasenberg:
The tool parses html file, saved form page https://launchstudio.bluetooth.com/ICS/<some-id>, that contains ICS tables, into rst file.

### DIFF
--- a/tools/ics_rst_from_html.py
+++ b/tools/ics_rst_from_html.py
@@ -1,0 +1,155 @@
+#
+# auto-pts - The Bluetooth PTS Automation Framework
+#
+# Copyright (c) 2021, Codecoup.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms and conditions of the GNU General Public License,
+# version 2, as published by the Free Software Foundation.
+#
+# This program is distributed in the hope it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+
+"""ICS Launch Studio html to rst file parser
+
+Log into your bluetooth.com account. Go to
+https://launchstudio.bluetooth.com/MyProjects/DraftProjects/
+and select your project. Go to ICS Selection tab and select profile, e.g. GAP.
+Save the page of the profile as html file.
+
+Run this script as follows:
+
+cd auto-pts
+python3 ics_rst_from_html.py GAP path/to/Launch-Studio-ICS-Selection.html
+
+In directory of run will be created gap-pics.rst file.
+"""
+
+import sys
+import os
+import re
+import pandas as pd
+
+
+def make_table(table, profile):
+    grid = [['Parameter Name', 'Selected', 'Description']]
+
+    for i in range(len(table.values)):
+        new_row = []
+        parameter_name = 'TSPC_{}_{}'.format(profile, table.Item[i].replace('/', '_'))
+        selected = 'True' if table.Support[i] else 'False'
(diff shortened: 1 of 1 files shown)


sjanc (approved):

tools/ics_rst_from_html.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def table_div(max_cols, header_flag=1):
+    out = ""
+    if header_flag == 1:
+        style = "="
+    else:
+        style = "-"
+
+    for max_col in max_cols:
```
michal-narajowski, line 70:
```suggestion
    for col_width in max_width_per_column:
```

tools/ics_rst_from_html.py
```diff
@@ -0,0 +1,155 @@
...
+    out = out.strip() + "\n"
+    return out
+
+
+def normalize_row(row, max_cols):
+    r = ""
+    for i, max_col in enumerate(max_cols):
+        try:
+            r += row[i] + (max_col - len(row[i]) + 1) * " "
+        except Exception as e:
```
michal-narajowski, line 82: Do we need this here? If there is a possibility of an exception we should implement this in such a way as to not trigger it. And definitely not handle generic Exception, at least a specific exception that is triggered.

tools/ics_rst_from_html.py
```diff
@@ -0,0 +1,155 @@
...
+    return r.strip() + "\n"
+
+
+if __name__ == '__main__':
+
+    if len(sys.argv) < 3:
+        sys.exit('Please pass profile name and html file path as arguments, e.g.:\n'
+                 '$ python3 %s GAP path/to/Launch-Studio-ICS-Selection.html\n'
+                 'html file - save https://launchstudio.bluetooth.com/ICS/<some id> '
+                 'page as html file' % sys.argv[0])
```
michal-narajowski, line 94: Use new formatting syntax.

tools/ics_rst_from_html.py
```diff
@@ -0,0 +1,155 @@
...
+
+    table_titles = re.findall(r'(?=<h4>\s*Table\s*.+:\s*).+?(?=</h4>)', content)
+    for i in range(len(table_titles)):
+        table_titles[i] = table_titles[i].split(r':')[1].strip()
+
+    with open(temp_path, 'w') as f:
+        f.write(content)
+
+    # Returns list of all tables on page
+    tables = pd.read_html(temp_path)
```
michal-narajowski, line 128: `pd.read_html()` also takes raw HTML string so no need to create temp file.

tools/ics_rst_from_html.py
```diff
@@ -0,0 +1,155 @@
...
+TCRL version: %s
+
+M - mandatory
+
+O - optional
+
+
+''' % (profile_name.lower(), head, len(head) * '*', tcrl_ver)
+
+    for tab, title in zip(tables, table_titles):
```
michal-narajowski, line 147:
```suggestion
    for table, table_title in zip(tables, table_titles):
```

tools/ics_rst_from_html.py
```diff
@@ -0,0 +1,155 @@
...
+In directory of run will be created gap-pics.rst file.
+"""
+
+import sys
+import os
+import re
+import pandas as pd
+
+
+def make_table(_tab, _profile):
```
michal-narajowski, line 37:
```suggestion
def make_table(table, profile):
```

michal-narajowski (changes requested):
Nice work on the script! I have some comments but otherwise looks good.

tools/ics_rst_from_html.py
mkasenberg: Sorry, it's just debugging leftovers.

michal-narajowski (approved):

PR #554, opened 2021-09-08
Mynewt: fix GATT/SR/GAW/BV-01-C

KKopyscinski:
Before verifying data written by PTS we should handle
gatts_attr_value_changed_ev - calling gatts_get_attr_val first will
lead to receiving value change event when we want response to command -
op code error will occur. Receiving value change event also means
that there's actual data to be checked.

### DIFF
--- a/ptsprojects/mynewt/gatt_wid.py
+++ b/ptsprojects/mynewt/gatt_wid.py
@@ -162,7 +162,9 @@ def hdl_wid_75(desc):
     handle = int(MMI.args[0], 16)
     value = MMI.args[1]
 
-    time.sleep(5)
+    # write was unsuccessful
+    if not btp.gatts_verify_write_success(desc):
+        return False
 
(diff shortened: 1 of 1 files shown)


ptsprojects/mynewt/gatt_wid.py
```diff
@@ -162,16 +162,19 @@ def hdl_wid_75(desc):
...
-    _, _, val = attr
+        _, _, val = attr
 
-    parsed_val = hexlify(val).decode('utf-8').upper()
+        parsed_val = hexlify(val).decode('utf-8').upper()
 
-    return bool(value == parsed_val)
+        return bool(value == parsed_val)
+    # write was unsuccessful
+    else:
```
michal-narajowski, line 176: Exit early

michal-narajowski (changes requested):
Small nitpick but otherwise looks good.

michal-narajowski (approved):

PR #558, opened 2021-09-09
L2CAP tests fixes

sjanc:


wid/gap.py
```diff
@@ -383,7 +383,10 @@ def hdl_wid_76(desc):
 
 
 def hdl_wid_77(desc):
-    btp.gap_disconn()
+    try:
+        btp.gap_disconn()
+    except types.BTPError:
+        pass
```
michal-narajowski, line 389: Add a log here: 
```suggestion
        logging.debug("Ignoring expected error on disconnect'")
```

michal-narajowski (changes requested):
Small nitpick

michal-narajowski (approved):

PR #566, opened 2021-09-13
Add support for L2CAP credits command

sjanc:
This allows to instruct IUT to return credits on specified channel.


michal-narajowski:
Related: https://github.com/zephyrproject-rtos/zephyr/pull/38488

sjanc:
38488 is now merged

PR #569, opened 2021-09-13
mynewt/sm: fixed test cases' conditions

jrotkiewicz:
Fixed wrong conditions according to TS


jrotkiewicz:
@michal-narajowski

michal-narajowski (approved):

PR #578, opened 2021-09-15
Zephyr: Add missing provisioning testes for node

agata-ponitka:


pybtp/btp/mesh.py
```diff
@@ -271,6 +271,15 @@ def mesh_config_prov():
     data = bytearray(struct.pack("<16s16sBHBHB", uuid, static_auth, output_size,
                                  output_actions, input_size, input_actions, auth_method))
 
+    pub_key = stack.mesh.pub_key_get()
+    priv_key = stack.mesh.priv_key_get()
+
+    if pub_key:
+        data.extend(struct.pack("<64s", binascii.unhexlify(pub_key)))
```
michal-narajowski, line 278: What if pub_key is missing? You added a parameter set_keys that contains 64 bytes of pubkey and 32 bytes of priv key. If pub_key is missing then you will only send 32 bytes and IUT will try to read 64+32 which will lead to undefined behavior.

ptsprojects/zephyr/mesh.py
```diff
@@ -253,6 +254,16 @@ def test_cases(ptses):
             "MESH", "TSPX_device_public_key", pub_key)),
         TestFunc(lambda: stack.mesh.pub_key_set(pub_key)),
     ]
+
+    pre_conditions_pub_ptiv_key = pre_conditions + [
```
michal-narajowski, line 258:
```suggestion
    pre_conditions_pub_priv_key = pre_conditions + [
```

michal-narajowski (changes requested):
Some minor comments.

michal-narajowski (approved):

PR #592, opened 2021-09-23
Mynewt: update workspace to PTS 8.1.1

KKopyscinski:
Updated workspace using old ICS file - as there's no new TCRL.


michal-narajowski (changes requested):
Might be better to do it in launch studio and check for consistency because Nimble supports GATT Client/Server over LE and EATT.

KKopyscinski:
As discussed offline, EATT is not yet supported

michal-narajowski (approved):

PR #594, opened 2021-09-29
tools: Add workspace_vs_test_plan.py tool

mkasenberg:


tools/testplan_vs_workspace.py
```diff
@@ -0,0 +1,131 @@
...
+        sys.exit('{} is not a file or workspace (*.pqw6)!'.format(workspace_path))
+
+    if not os.path.isfile(testplan_path) or not testplan_path.endswith('.xlsx'):
+        sys.exit('{} is not a file or test plan (*.xlsx)!'.format(testplan_path))
+
+    print('Starting PTS needed to workspace parsing ...')
+    pts = PyPTSControl()
+    pts.open_workspace(workspace_path)
+
+    if pr_name is not None:
```
michal-narajowski, line 84: There is no need to set pixits so this whole block of code can be removed as well as "project" argument

tools/testplan_vs_workspace.py
```diff
@@ -0,0 +1,131 @@
...
+Usage:
+$ python3 testplan_vs_workspace.py path/to/workspace path/to/test_plan.xlsx <project_name>
+"""
+import importlib
+import sys
+import os
+import pandas as pd
+import win32com
+import wmi
+
```
michal-narajowski, line 27: Add this to be able to run from project folder from command line.
```suggestion
sys.path.insert(
    0, os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```

michal-narajowski (changes requested):
Looks great! Some comments:
- `autoptsclient_requirements.txt` is missing `openpyxl`
- LT2 test cases should be excluded from the report as they are not visible in the Test Plan
- "project name" argument can be removed

michal-narajowski (approved):

PR #609, opened 2021-10-18
btp: Add support for options in L2CAP connect

sjanc:
PTS and TS are required inconsistent behaviour in terms of how IUT
should return credits. Some tests require return on explicit UT request
and some require that IUT returns credits autonomously.

With this patch we have more flexibility on how IUT is handling
L2CAP channels.


sjanc:
https://github.com/zephyrproject-rtos/zephyr/pull/39500   related zephyr change

sjanc:
@KKopyscinski FYI

PR #613, opened 2021-10-18
btp: Fix GAP Pairing Failed event handling

sjanc:
gap_wait_for_pairing_fail() was always returning true.


michal-narajowski:
Solved by https://github.com/intel/auto-pts/pull/614

PR #615, opened 2021-10-19
Mynewt: init mesh in new way

KKopyscinski:
cb19758701bcbf44dc93441e0e34b7c15e7793f8 introduced new Mesh
initialization, apply it to Mynewt to avoid "To many arguments"
error.


KKopyscinski:
@michal-narajowski I think that's all that's needed, because rest of changes takes place in common wid module
FYI @mkasenberg

PR #619, opened 2021-10-20
tools requirements

sjanc:


sjanc:
@mkasenberg   FYI

not sure if more should be added though (only this was missing on my box)

sjanc:
those are executable after all, aren't they?

not sure about create-workspace but others are suppose to be run from shell by user

PR #620, opened 2021-10-21
GATT service rework

sjanc:
@KKopyscinski FYI

This is a bit of an RFC :-)


doc/btp_spec.txt
```diff
@@ -3724,3 +3725,649 @@ Commands and responses:
...
+					Status (1 octet)
+
+	Opcode 0x8f - Configure Indications Response
+
+		Controller Index:	<controller id>
+		Event parameters:	Address_Type (1 octet)
+					Address (6 octets)
+					Status (1 octet)
+
+	Opcode 0x90 - Notification/Indication Received
```
sjanc, line 4269: support for multiple handle values notification - either new even or extend this with new type (and possible explicit count)

MariuszSkamra (approved):
LGTM

doc/btp_spec.txt
```diff
@@ -3724,3 +3725,649 @@ Commands and responses:
...
+		If Status is non-zero Data_Length shall be set to 0.
+
+	Opcode 0x88 - Read Using Characteristic UUID Response
+
+		Controller Index:	<controller id>
+		Event parameters:	Address_Type (1 octet)
+					Address (6 octets)
+					Status (1 octet)
+					Data_Length (2 octets)
+					Data (variable)
```
sjanc, line 4204: Data should be tuples[] (handle, data)

doc/btp_spec.txt
```diff
@@ -3760,3 +3772,718 @@ Commands and responses:
...
+	Opcode 0x8a - Read Multiple Characteristic Values Response
+
+		Controller Index:	<controller id>
+		Event parameters:	Address_Type (1 octet)
+					Address (6 octets)
+					Status (1 octet)
+					Data_Length (2 octets)
+					Data (variable)
+
+		If Status is non-zero Data_Length shall be set to 0.
```
KKopyscinski, line 4300: As we are sending multiple responses of this kind, I suggest that we define that last one has `Status=0` and `Data_Length=0`

doc/btp_spec.txt
```diff
@@ -3760,3 +3772,718 @@ Commands and responses:
...
+
+		Valid Type parameter values:
+					0x01 = Notification
+					0x02 = Indication
+					0x03 = Multiple Notification
+
+		This event indicates that IUT has received notification
+		or indication.
+
+	Opcode 0x91 - Multiple Handle Notification Received
```
KKopyscinski, line 4355: This might be not needed, because we don't have means to easily determine that notification comes from `ATT_MULTIPLE_HANDLE_VALUE_NTF` - IUTs (Both Zephyr and Mynewt) just handle these notifications one by one. What we can expect is multiple `0x90 `events

doc/btp_spec.txt
```diff
@@ -3760,3 +3772,718 @@ Commands and responses:
...
+		If Status is non-zero Data_Length shall be set to 0.
+
+	Opcode 0x88 - Read Using Characteristic UUID Response
+
+		Controller Index:	<controller id>
+		Event parameters:	Address_Type (1 octet)
+					Address (6 octets)
+					Status (1 octet)
+					Data_Length (2 octets)
+					Value_Len (1 octet)
```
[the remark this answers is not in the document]
sjanc: re 1) ATT_READ_BY_TYPE_RSP has common length for all handle-data pairs therefore all must have same length
re 2) I'd expect error response in case it cannot fit into BTP

doc/btp_spec.txt
sjanc: why multiple responses?  it should be single response where data is set of two (or more) values

doc/btp_spec.txt
sjanc: it is IUT implementation detail but I guess we can add it later on if there is explicit need for this

doc/btp_spec.txt
```diff
@@ -3761,3 +3773,700 @@ Commands and responses:
...
+					Data (variable)
+
+		Valid Type parameter values:
+					0x01 = Notification
+					0x02 = Indication
+
+		This event indicates that IUT has received notification
+		or indication.
+
+	Opcode 0x92 - Read Multiple Variable Characteristic Values Response
```
mkasenberg, line 4355: Opcode 0x91

hermabe:
I see no way to handle cases where PTS specifies which bearer to send a request on. For example, in "Multiple Client Read Requests on Unsynchronized Database with Robust Caching" (GATT/CL/GPM/BV-{07-12}-C), where some requests shall be sent on ATT, and some requests shall be sent on EATT.

hermabe:
> I'd rather not extend every GATT command with 'bearer' field, but if needed we can add new BTP command that will allow to select preferred bearer. Would that work for you?

Yes, that should work. The tester application could store the preferred bearer and then use that for the next request.

sjanc:
> > I'd rather not extend every GATT command with 'bearer' field, but if needed we can add new BTP command that will allow to select preferred bearer. Would that work for you?
> 
> Yes, that should work. The tester application could store the preferred bearer and then use that for the next request.

Yeap, this is that I had in mind

michal-narajowski (approved):

PR #634, opened 2021-11-03
wid: Add support for GAP WID 403

sjanc:
This was affecting GAP/CONN/DCEP/BV-05-C and GAP/CONN/DCEP/BV-06-C.

### DIFF
--- a/wid/gap.py
+++ b/wid/gap.py
@@ -954,6 +954,18 @@ def hdl_wid_402(desc):
     return True
 
 
+def hdl_wid_403(desc):
+    #Please perform the Directed Connection Establishment Procedure using RPA
+    #then resolve the PTS address and connect with PTS.
+    stack = get_stack()
+
+    if not stack.gap.iut_has_privacy():
+        return False
+
+    btp.gap_conn()
+    return True
+
+
 def hdl_wid_1002(desc):
     stack = get_stack()
     return stack.gap.get_passkey()


michal-narajowski (approved):

wid/gap.py
```diff
@@ -954,6 +954,18 @@ def hdl_wid_402(desc):
     return True
 
 
+def hdl_wid_402(desc):
```
michal-narajowski, line 957: Did you mean 403?

michal-narajowski (changes requested):

wid/gap.py
```diff
@@ -954,6 +954,18 @@ def hdl_wid_402(desc):
...
+    #Please perform the Directed Connection Establishment Procedure using RPA
+    #then resolve the PTS address and connect with PTS.
+    stack = get_stack()
+
+    if not stack.gap.iut_has_privacy():
+        return False
+
+    btp.gap_conn()
```
michal-narajowski, line 965: I think this does not test DCEP. Under the hood, this sends a Gap Connect command to IUT with PTS address as an argument.

wid/gap.py
michal-narajowski: Might make sense to call 402 if the expected behavior is the same.

sjanc: yes, it was suppose to be 403 :)

sjanc: one is general, and one is directed,  I'd prefer to keep those separate even if currently implementations match

michal-narajowski (approved):

PR #636, opened 2021-11-04
Support for directed advertising using peer RPA

sjanc:


michal-narajowski (approved):
It is ok as a workaround although I wish this wasn't necessary.

PR #639, opened 2021-11-08
iutctl: clear buffer after dialog board reset

jrotkiewicz:
Dialog board sends 0 after reset. Added fix to discard this zero.
Divided PR: https://github.com/intel/auto-pts/pull/585


jrotkiewicz:
@michal-narajowski

PR #647, opened 2021-11-22
Nimble host fixes

KKopyscinski:
This PR contain various fixes for Nimble host for PTS 8.1.2

### DIFF
--- a/ptsprojects/mynewt/gap.py
+++ b/ptsprojects/mynewt/gap.py
@@ -258,6 +258,16 @@ def test_cases(ptses):
                   cmds=pre_conditions +
                   [TestFunc(btp.gap_set_io_cap, IOCap.display_only)],
                   generic_wid_hdl=gap_wid_hdl),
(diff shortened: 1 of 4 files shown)


bot/iut_config/mynewt.py
```diff
@@ -22,6 +22,12 @@
         ],
     },
 
+    "public_address.conf": {
+        "overlay": {' BLE_PUBLIC_DEV_ADDR': '(uint8_t[6]){0xff, 0xff, 0xff, 0xff, 0xff, 0xff}'},
```
sjanc, line 26: Why do we need to use public address for those tests in first place?

KKopyscinski: ok, I deleted this commit - this passes with CASE0072061

sjanc (approved):
some feel a bit hacking (eg sleep:) but I guess we need to live with that until PTS starts to behave better regarding MMIs...

michal-narajowski (approved):

PR #656, opened 2021-12-03
zephyr: Fix GAP/SEC/AUT/BV-19-C test

sjanc:
This is temporary workaround to make test pass until CASE0072154
is resolved. Since requesting service for triggering security is
optional we can just enabled security without it.


michal-narajowski (approved):

KKopyscinski:
I tested this solution on Mynewt and it wasn't working - however, I submitted a patch that works for both https://github.com/intel/auto-pts/pull/657 I think that it's intended way to skip optional part

PR #659, opened 2021-12-08
Zephyr: adjust IO Capabilities

KKopyscinski:


KKopyscinski:
this solves https://github.com/intel/auto-pts/issues/649

wid/sm.py
```diff
@@ -100,6 +100,21 @@ def hdl_wid_141(desc):
...
 
+def hdl_wid_142(desc):
+    """
+    Please confirm the following number matches IUT: [passkey]
+    """
+    pattern = '[\d]{6}'
+    passkey = re.search(pattern, desc)[0]
+    stack = get_stack()
+
+    btp.gap_passkey_confirm_req_ev()
```
sjanc, line 111: what if IUT sends this before MMI is handled?

KKopyscinski:
@sjanc I submitted a patch to tester to set DisplayYesNo IO Cap https://github.com/zephyrproject-rtos/zephyr/pull/41038

wid/sm.py
```diff
@@ -101,6 +119,26 @@ def hdl_wid_141(params: WIDParams):
...
+    bd_addr_type = btp.pts_addr_type_get()
+
+    if stack.gap.get_passkey() is None:
+        return False
+
+    btp.gap_passkey_confirm_rsp(bd_addr, bd_addr_type, passkey)
+
+    if stack.gap.passkey.data == passkey:
+        return True
+    return False
```
michal-narajowski, line 139:
```suggestion
    return stack.gap.passkey.data == passkey
```

michal-narajowski (changes requested):

pybtp/btp/gap.py
```diff
@@ -594,46 +632,39 @@ def gap_passkey_entry_rsp(bd_addr, bd_addr_type, passkey):
...
 
+    data_ba.extend(chr(bd_addr_type).encode('utf-8'))
+    data_ba.extend(bd_addr_ba)
 
-def gap_passkey_entry_req_ev(bd_addr=None, bd_addr_type=None):
-    logging.debug("%s %r %r", gap_passkey_entry_req_ev.__name__, bd_addr,
-                  bd_addr_type)
-    iutctl = get_iut()
+    if isinstance(passkey, str):
+        passkey = int(passkey, 32)
```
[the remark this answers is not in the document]
KKopyscinski: As this command has almost the same structure as `gap_passkey_entry_rsp` i think it's just a copy-paste error - but this begs the question why there is 32 as a base :) I think in both places it can be default (without any base, so base = 10) because we are just making sure all passkeys have the same type.

KKopyscinski:
@michal-narajowski Rebased onto master to check conflicts and added changes you suggested

sjanc (approved):

michal-narajowski (approved):

PR #661, opened 2021-12-10
autoptsclient_common: don't restart PTS for not-implemented test cases

mringwal:
I'd like to argue that there's no point in restarting PTS if a test case is not implemented by the local pts project :)
(I did disable a lot of them as PTS was hanging in some of them. Now that I got a YKUSH hub, I can also just re-enable them)


michal-narajowski (approved):
Sounds good to me :)

PR #670, opened 2021-12-16
boards: Add support for nrf53

hermabe:
Previously, only nrf52 was supported.

### DIFF
--- a/ptsprojects/boards/__init__.py
+++ b/ptsprojects/boards/__init__.py
@@ -3,6 +3,7 @@
 #
 # Copyright (c) 2021, Intel Corporation.
 # Copyright (c) 2021, Codecoup.
+# Copyright (c) 2021, Nordic Semiconductor ASA.
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms and conditions of the GNU General Public License,
@@ -22,6 +23,7 @@ import logging
 import pkgutil
 import importlib
 import subprocess
+from collections import defaultdict
 
 # For each new board just create a <new-board-name>.py file that contains reset_cmd() function and list
 # of supported boards.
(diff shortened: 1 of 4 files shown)


ptsprojects/boards/__init__.py
```diff
@@ -118,14 +119,20 @@ def get_device_list():
...
-def get_free_device():
+def get_free_device(board=None):
     """Returns tty path and jlink serial number of a free device."""
     devices = get_device_list()
 
-    for dev in devices.keys():
-        if dev not in devices_in_use:
-            devices_in_use.append(dev)
-            return dev, devices[dev]
+    SNR_PREFIX_FOR_BOARD = {
```
michal-narajowski, line 126: Maybe a defaultdict could be used here so when there is an unknown board then an empty string would be returned?

michal-narajowski (changes requested):
LGTM, a small nitpick.

ptsprojects/boards/__init__.py
hermabe: Wouldn't that mean that an unknown `board` argument would be ignored? I would rather it fail loudly if I pass an unknown board argument.

michal-narajowski: The thing is I don't want to break this for other boards like reel_board. Anyway, why do we need this prefix check?

hermabe: I hadn't considered that `"reel_board"` may also be given as an argument here. Will switch to a `defaultdict(str)`.

The prefix check is needed if one has both 52 and 53 boards connected. Without the prefix check a 52 board may be selected even if `board` is `"nrf53"`.

michal-narajowski (approved):

PR #673, opened 2022-01-13
Include XML files with test execution logs from PTS side in report

KKopyscinski:
Now every test with PASS verdict has appropriate entry in report.xlsx
with matching filename of XML file from PTS. All XML files are also
added to seperate "XMLs" folder uploaded to drive.


KKopyscinski:
PTAL @sjanc @michal-narajowski

PR #676, opened 2022-01-24
zephyr bot: add recover option for boards with app-protect

AyturkDuzen:
* The nRF5340 has a flash read-back protection feature. When flash read-back protection is active, you will need to recover the chip before reflashing. If you are flashing with west, run this command for more details on the related --recover option.
* Adding recovery does not affect existing devices.
https://docs.zephyrproject.org/latest/boards/arm/nrf5340dk_nrf5340/doc/index.html#flashing


AyturkDuzen:
@sjanc can we merge this? it is a small change but will enable newer nrf52840 and nrf5340 devices that are shipped with app protect to run with auto-pts.

michal-narajowski (approved):

PR #685, opened 2022-02-07
common: Fix Client progress indication

michal-narajowski:
The test result was marked as progress even when there was no database
records with previous status for that test case.


michal-narajowski:
FYI @mkasenberg

sjanc (approved):

PR #686, opened 2022-02-07
zephyr: Update more tests to PTS 8.2

michal-narajowski:
More tests don't require two instances anymore.


michal-narajowski:
MESH/SR/PROX/BV-14-C is failing, needs to be investigated.

sjanc (approved):

PR #688, opened 2022-02-09
Zephyr: bot: use -i instead of depreciated --snr

AyturkDuzen:
Currently while flashing, we receive a warning about the below depreciation. This pr uses the newer argument to select snr

-- west flash: using runner nrfjprog
WARNING: runners: Argument --snr is deprecated, use -i/--dev-id instead.
-- runners.nrfjprog: Recovering and erasing all flash memory.


AyturkDuzen:
Can it be merged?

PR #691, opened 2022-02-12
Zephyr: pixit: set correct default values for TSPX_iut_invalid_conn_int*

AyturkDuzen:
setting the new default values from pts software the following IXIT :

TSPX_iut_invalid_connection_interval_min as 0008
TSPX_iut_invalid_connection_interval_max as 00AA


sjanc:
Hi,

Those are already set in preconditions for tests that use those.
I think we should eventually clean up on how PIXITs are set

@michal-narajowski   should we use set_pixit() for static values and update_pixit() for dynamic?  or we want to move away from set_pixit() at all?

AyturkDuzen:
Hi,

I thought in preconditions, they are set dynamically. Regardless in set_pixit, we should not set invalid value initially anyway, rather the default values. I agree with you about how pixit values are set, so it could be nice to clean up one of the places not to have duplicate set.

michal-narajowski:
I think the idea was to set the default values in set_pixit() and then change whatever is needed per test. If the current values are invalid then we need to change them. I agree, we should do a cleanup.

michal-narajowski (approved):

michal-narajowski:
@AyturkDuzen Was this causing any issues in PTS or auto-pts?

PR #700, opened 2022-02-22
wid: Add a new wid to be ready for next pts release

ilhanates:
**Note: This is fixing an issue that will happen on next PTS release, you can wait to merge or just merge as it is independent change with no harm.**

It looks L2CAP/LE/CID/BI-01-C will fail on next pts
release (>8.2.0). The test will be adding new step
as WID=61 to confirm IUT does not send any l2cap data.

PTS software case:
CASE0072383 (L2CAP/LE/CID/BI-01-C): Updated Test procedure
for utilizing K-frame instead of signaling packet.

I came across with the issue while testing the L2CAP.ets patch (shared in [CASE0072529](https://bluetooth.service-now.com/ess/case_detail.do?sysparm_document_key=x_bsig_case_manage_case,c95e41771b6585100999a9bfbd4bcb2e)) with all l2cap tests.

### DIFF
--- a/wid/l2cap.py
+++ b/wid/l2cap.py
@@ -310,6 +310,11 @@ def hdl_wid_60(params: WIDParams):
     return False
 
 
+def hdl_wid_61(_: WIDParams):
+    """description: Please confirm the IUT does not send the L2CAP Data to the Upper Tester."""
+    return True
+
+
 def hdl_wid_100(_: WIDParams):
     l2cap = get_stack().l2cap
     for channel in l2cap.channels:


wid/l2cap.py
```diff
@@ -310,6 +310,11 @@ def hdl_wid_60(params: WIDParams):
     return False
 
 
+def hdl_wid_61(_: WIDParams):
+    """description: Please confirm the IUT does not send the L2CAP Data to the Upper Tester."""
```
sjanc, line 314: hmm shouldn't we actually check this before confirming ?

ilhanates: to me, it looks like , they want to inform IUT side to not send anything. I tested this with new wid, otherwise giving missing wid error

michal-narajowski: @ilhanates We discussed this offline and we can merge as is. We will create an issue to add verification mechanism in this and other mmi handlers.

michal-narajowski (approved):

PR #709, opened 2022-03-10
common: Refactor client parser

mkasenberg:
Moves the parser to separate file. Simplifies adding a new btp client support. Fixes duplicated parsing that was happening in inherited constructors of bot clients.


michal-narajowski (approved):
Looks good, nice work!

PR #719, opened 2022-03-24
Mynewt: Update workspace with PTS 8.2

KKopyscinski:


workspaces/Mynewt Nimble Host/Mynewt Nimble Host.pqw6
```diff
@@ -2610,29 +2640,41 @@
...
           </Row>
+          <Row>
+            <Name>TSPC_SUM_ICS_31_21</Name>
+            <Description>Core Specification Version 5.2. Adopted 31 December 2019 (C.1)</Description>
+            <Value>FALSE</Value>
+            <Mandatory>FALSE</Mandatory>
+          </Row>
+          <Row>
+            <Name>TSPC_SUM_ICS_31_22</Name>
+            <Description>Core Specification Version 5.3. Adopted 06 July 2021 (C.1)</Description>
```
michal-narajowski, line 2674: This probably should be updated through Launch Studio, to verify consistency. We should support the latest spec.

workspaces/Mynewt Nimble Host/Mynewt Nimble Host.pqw6
```diff
@@ -3086,25 +3148,37 @@
...
           </Row>
+          <Row>
+            <Name>TSPC_L2CAP_2_48a</Name>
+            <Description>Enhanced Credit Based Flow Control Mode - BR/EDR (C.26)</Description>
+            <Value>FALSE</Value>
+            <Mandatory>FALSE</Mandatory>
+          </Row>
+          <Row>
+            <Name>TSPC_L2CAP_2_48b</Name>
+            <Description>Enhanced Credit Based Flow Control Mode - LE (C.27)</Description>
```
michal-narajowski, line 3175: Again this should be verified through Launch studio because we do support ECFC if I recall correctly.

michal-narajowski (changes requested):

workspaces/Mynewt Nimble Host/Mynewt Nimble Host.pqw6
KKopyscinski: It was updated via launch studio, but this had to be bumped manually. I also enabled Periodic Adv. BTW, I think this is a bug in PTS that doesn't change this ICS based on ICS file at workspace generation, because if you look here: https://github.com/intel/auto-pts/pull/719/commits/f0852c516de6e2c9f63943f49e698a418b3b0ca4#diff-0d9cc24875a692b2cbb450fa778bdf36f59ed35df9f4c0f22b0c0d171cecfc68L714 you can clearly see that it was changed from 31_21 to 31_22.

KKopyscinski:
> Rebase please.

done

michal-narajowski (approved):

PR #720, opened 2022-03-24
Mynewt PTS 8.2 fixes

KKopyscinski:


errata.yaml
```diff
@@ -2,8 +2,10 @@
 # GAP/CONN/NCON/BV-01-C: CASE00xxxxx
 # GAP/CONN/NCON/BV-02-C: CASE00xxxxx
 
+
 GAP/CONN/CPUP/BV-04-C: CASE0072503
 GAP/CONN/CPUP/BV-05-C: CASE0072503
+GAP/BOND/BON/BV-04-C: CASE Id#13064
```
michal-narajowski, line 8: Use the same format of CASE ID as in other examples for simplicity. If this is a TSE or SE then add that information.

michal-narajowski (changes requested):

errata.yaml
KKopyscinski: ok, I was wondering if I should differentiate this from others, as this is submitted via new support site, but I guess `CASE13064` works fine. Also, we can differentiate by shorter case number

michal-narajowski: In that case I think the previous way was better. Maybe "Request ID #13064" as they are now called requests and there is no mention of a "case".

michal-narajowski (approved):

michal-narajowski:
Please rebase.

michal-narajowski (approved):

PR #739, opened 2022-04-19
Nimble: Fixes in PTS 8.2.2

KKopyscinski:


michal-narajowski (approved):

sjanc (approved):

errata.yaml
```diff
@@ -4,7 +4,7 @@
 
 
 GAP/CONN/CPUP/BV-04-C: CASE0072503
-GAP/CONN/CPUP/BV-05-C: CASE0072503
+GAP/CONN/CPUP/BV-05-C: Request ID #13766
```
hermabe, line 7: `#` should not be used here, it is parsed as a comment. The results file will only print "Errata Request ID" like it did here: https://github.com/zephyrproject-rtos/bluetooth-qualification/blob/05e373491dabb29086725cd59b76a882d1612f35/host/autopts_report/report_iut_zephyr_2022_04_08_20_13_18.txt#L34

PR #741, opened 2022-04-25
zephyr: Make GATT/SR/GAW/BV-13-C pass consistently

sjanc:
PTS seems to randomly choose characteristics for writes in that test
which results in authentication error if unlucky. Use database 6
which has only 1 long writable characteristic for consistent results.


MariuszSkamra (approved):
LGTM

PR #759, opened 2022-06-16
Add support for L2CAP/TIM/BV-03-C

hermabe:
This is an L2CAP test that tests a requirement on ATT, it needs special casing and a new BTP command.

Corresponding Zephyr PR: https://github.com/zephyrproject-rtos/zephyr/pull/46608


michal-narajowski:
It looks good to me. Please add documentaiton in `doc/btp_spec.txt`. 

@KKopyscinski Could you take a look?

sjanc (changes requested):
This could also be handled in l2cap connect implementation on IUT side (when PSM is EATT) but in general I'm also fine with dedicated BTP command for EATT connection.

but if you decide to go with new BTP command than please update btp.txt specification

hermabe:
> shouldn't there be a dedicated "EATT connected" event too

Yes, that would be useful, but hard to implement with the current EATT API in Zephyr. I tried getting a EATT-connected callback added, but it was rejected upstream.

> but if you decide to go with new BTP command than please update btp.txt specification

Will do

sjanc:
Alright,  we can add event later on if needed

KKopyscinski:
Looks good to me, one thing about event, I can see a `bt_eatt_count` function in Zephyr's API: maybe you can lookup this value before connection is requested and calculate difference on `connected_cb`. Then you can send event with field that will tell how many channels were created. I see that Zephyr PR is already merged, but this can be added in the future.

sjanc (approved):

michal-narajowski (approved):

PR #792, opened 2022-08-11
bot: Split errata.yaml into multiple files

mkasenberg:
There is one errata/common.yaml which can be overwritten with errata/<project_name>.yaml.


errata/zephyr.yaml
```diff
@@ -0,0 +1,23 @@
+# To add note to the reports about awaited errata, list them here e.g.:
+# GAP/CONN/NCON/BV-01-C: CASE00xxxxx
+# GAP/CONN/NCON/BV-02-C: CASE00xxxxx
+
+MESH/PVNR/PROV/BV-01-C: https://github.com/zephyrproject-rtos/zephyr/pull/38492
```
michal-narajowski, line 5: You can keep MESH entries in common file.

michal-narajowski (approved):

PR #795, opened 2022-08-19
mmdl: send Get Column with Raw_Value_X = 0

KKopyscinski:
We cannot guarantee that more than 1 column exist, and PTS was failing
to send us full response for other values.


KKopyscinski:
This is affecting MMDL/CL/SNR/BV-12-C

michal-narajowski (approved):

PR #822, opened 2022-10-24
modify the baud rate instead of the fixed 115200

justin-nxp:
modify the baud rate instead of the fixed 115200


autopts/ptsprojects/mynewt/iutctl.py
```diff
@@ -25,12 +25,11 @@
 from autopts.ptsprojects.boards import Board, get_debugger_snr, tty_to_com
 from autopts.pybtp.types import BTPError
 from autopts.pybtp.iutctl_common import BTPWorker, BTP_ADDRESS, RTT, BTMON, BTPSocketSrv
-
+from autopts.ptsprojects.zephyr import SERIAL_BAUDRATE
```
michal-narajowski, line 28: mynewt should not depend on zephyr.

michal-narajowski (changes requested):
Small nitpick, otherwise LGTM.

PR #827, opened 2022-11-14
Add test specific retry count

KKopyscinski:
This commit adds configuration for bot that allows to set alternative to default setting of retries for certain tests. This allows to override default setting, which still will be valid for rest of the tests. This custom value can be smaller than default, when we know test doesn't pass and this helps us save time on retrying it, or larger, when we know test passes but not consistently, and may require more retries.


KKopyscinski:
@sjanc PTAL

KKopyscinski:
@michal-narajowski PTAL

michal-narajowski (approved):

PR #830, opened 2022-11-17
doc: Split btp docs per service

szymon-czapracki:
This commits extracts btp docs
per service for better readability.


MariuszSkamra (commented):
You may disagree, but I'd choose different naming convention for the files.
I would suggest either:
A) Rename `doc` to `btp` and rename the files to `overview.txt`, `core.txt`, etc.
or
B) Apply common `btp_` prefix so that we end up with something like `btp_overview.txt`, `btp_core.txt` etc.

sjanc (approved):

michal-narajowski (approved):

MariuszSkamra (approved):

sjanc (approved):

PR #835, opened 2022-12-06
Revert "zephyr: gatt: add a new database init_server_9"

mkasenberg:
This reverts commit a72a698e7adb43d74774735d41d88c45e9cd2ff2, which caused the bot to hang on the GATT/SR/GAN/BV-02-C test case after BTP ERROR. Even after superguard timeout and performed recovery the next test cases fail too.
[GATT_SR_GAN_BV-02-C_2022_12_06_08_45_16.log](https://github.com/intel/auto-pts/files/10164324/GATT_SR_GAN_BV-02-C_2022_12_06_08_45_16.log)

### DIFF
--- a/autopts/ptsprojects/zephyr/gatt.py
+++ b/autopts/ptsprojects/zephyr/gatt.py
@@ -241,6 +241,10 @@ def test_cases_server(ptses):
                               Prop.read | Prop.write | Prop.notify | Prop.indicate,
                               Perm.read | Perm.write, UUID.VND16_2),
                      TestFunc(btp.gatts_set_val, 0, Value.eight_bytes_1),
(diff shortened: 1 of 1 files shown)


sjanc:
@laxiLang @KKopyscinski     FYI   this is causing some issues in the bot (not sure why though...)

KKopyscinski:
@laxiLang @sjanc @mkasenberg This is not needed, as commit is ok (we need second CCC). There is a bug in Zephyr tester app, solved by https://github.com/zephyrproject-rtos/zephyr/pull/52816. Although I agree that https://github.com/intel/auto-pts/commit/a72a698e7adb43d74774735d41d88c45e9cd2ff2 was not properly tested, and causes test to hang (which breaks a lot of things, since this happens in preconditions).

PR #843, opened 2022-12-18
Add support to select dongle

justin-nxp:
this can add support to selected dongle，Adding the --dongle parameter allows two boards to be tested at the same time, greatly reducing testing time

### DIFF
--- a/autopts/ptscontrol.py
+++ b/autopts/ptscontrol.py
@@ -794,7 +794,12 @@ class PyPTS:
 
     def get_bluetooth_address(self):
         """Returns PTS bluetooth address string"""
-        device = self._pts.GetSelectedDevice()
+        args = parse_args()
+        if args.dongle:
+            device = args.dongle
+            log("slecte dongle :", device)
+        else:
+            device = self._pts.GetSelectedDevice()
         log(f"Remembered device {self._device}, selected device {device}")
         device_to_connect = None
         if self._device is not None and device != self._device:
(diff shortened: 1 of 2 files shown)


sjanc (approved):

autopts/ptscontrol.py
```diff
@@ -830,7 +835,12 @@ def get_bluetooth_address(self):
...
     def connect_to_dongle(self):
-        device_to_connect = None
+        args = parse_args()
+        if args.dongle:
+            device_to_connect = args.dongle
+            log("connect dongle ", args.dongle)
+        else:
+            device_to_connect = None
         devices = self._pts.GetDeviceList()
         for device in devices:
```
michal-narajowski, line 845: I think we will still override device_to_connect in this `for` loop, so your change will have no effect.

autopts/ptscontrol.py
```diff
@@ -794,7 +794,12 @@ def save_test_history_log(self, save):
 
     def get_bluetooth_address(self):
         """Returns PTS bluetooth address string"""
-        device = self._pts.GetSelectedDevice()
+        args = parse_args()
+        if args.dongle:
+            device = args.dongle
+            log("slecte dongle :", device)
```
michal-narajowski, line 800:
```suggestion
            log("selected dongle :", device)
```

michal-narajowski (changes requested):

michal-narajowski:
@justin-nxp Any update on this?

PR #849, opened 2023-01-23
zephyr: Add LE Audio PACS tests

mkasenberg:


mringwal:
Could you add the doc/btp_pacs.txt, too?

How does the IUT know that it should setup/expose the PACS GATT Service?

mkasenberg:
> How does the IUT know that it should setup/expose the PACS GATT Service?

I'm currently doing tests for Zephyr where the PACS server is statically configured. If it had to be configured at runtime, this probably could be handled with Register Service command of BTP Core Service (ID 0) and/or some commands from GATT Server Service (ID 7).

Until now these tests are very simple, maybe when I get to BAP tests, there will be a need to extend BTP PACS commands? Anyway, I'm open to suggestions.

sjanc:
as for PACS (and eventually other services) setup on IUT I'd simply do that in service register, so if PACS service was successfully registered, IUT shall have all required functionality set-up

mringwal:
Makes sense. If it's done with GATT Server Service, the complete service would need to be constructed by the Python profile implementation, so let's see if it's enough to enable PACS (or others) with the service. I 2 Sink and 2 Source ASEs  will be sufficient for BAP.

mringwal:
@mkasenberg it looks like the tests here cover the server side of things. What about client functionality? 

Should there be:
- different services for client & server, e.g. PACS Client and PACS Server
- a single profile service with separate commands for client and server side
- a single profile service with some kind of hack indicating client or server side in each command.

I'm currently working on BAP & ASCS and finally realized that most ASCS actions can be triggered on either side (e.g. Configure Codec) - and PTS is testing these, too. My current preference would be on a) different services for client & server - to keep adding commands easier. I've tried to pimp the ASCS Client commands to also support the server side, but it didn't save much work and leads to ugly code.

mkasenberg:
Oh, I see now. I've planned to do the tests in the order PACS, ASCS, BASS, BAP, so this PR covers just tests from PACS.TS.pdf https://www.bluetooth.org/docman/handlers/downloaddoc.ashx?doc_id=524504

The tests for PACS client were described in BAP part of documentation https://www.bluetooth.com/specifications/specs/basic-audio-profile-1-0-1/, so they are still ahead of me.

doc/btp_pacs.txt
```diff
@@ -0,0 +1,30 @@
...
+		that command is supported. Bit 0 is reserved and shall always
+		be set to 0. If specific bit is not present in response (less
+		than required bytes received) it shall be assumed that command
+		is not supported.
+
+		In case of an error, the error response will be returned.
+
+	Opcode 0x02 - Update characteristic
+		Controller Index:    <controller id>
+		Command parameters:  ID of the characteristic (1 octet)
```
sjanc, line 24: add allowed values for ID

sjanc (commented):
other than minor comment to doc, it looks good to me

doc/btp_pacs.txt
mringwal: For the current  PACS.TS.p0, it looks like any change is sufficient. I'm wondering if there should be specific commands like Update Source PAC, with an explicit PAC Record. Or should we wait until a PTS requires it? The BAP tests don't require that though.

sjanc: we can extend this (or add new dedicated command) if needed
those values here are for BTP,  IUT doesn't have to implement all of them

sjanc (approved):

doc/btp_pacs.txt
mkasenberg: I've extended the command.

autopts/ptsprojects/zephyr/pacs.py
```diff
@@ -0,0 +1,90 @@
...
+
+    iut_device_name = get_unique_name(pts)
+    stack = get_stack()
+
+    pre_conditions = [TestFunc(btp.core_reg_svc_gap),
+                      TestFunc(stack.gap_init, iut_device_name),
+                      TestFunc(btp.gap_read_ctrl_info),
+                      TestFunc(lambda: pts.update_pixit_param(
+                            "PACS", "TSPX_bd_addr_iut",
+                            stack.gap.iut_addr_get_str())),
```
szymon-czapracki, line 71: nitpick: Good to keep the same indentation here and in the few lines below.

michal-narajowski (approved):

PR #860, opened 2023-02-10
doc: add ASCS

mringwal:
This is the draft for BTP ASCS service and I'd like to some feedback on it. (Looks like I should have selected 'create draft pull request' and now cannot mark it as draft. I'll use that the next time). 

Our btp-iut implementation is based on it and it allows to pass the BAP Unicast Client & Server Configuration (SCC) and Streaming (STR) tests.

### DIFF
--- a/doc/btp_ascs.txt
+++ b/doc/btp_ascs.txt
@@ -1,0 +1,145 @@
+Audio Stream Control Service (ID tbd)
+=====================================
+
+Opcodes 0x02 and higher are use in server and client roles and require a Chan_ID.
+In Client role, ASCS Client Connect to peer is used to setup the GATT connection and return the Chan_ID. 
+In Server role, the lower test will query the ASCS Service and register for notifications. This is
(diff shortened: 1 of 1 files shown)


sjanc:
so I'm not sure if we should put such detailed parameters for BTP commands,  are those required by tests?

mringwal:
> so I'm not sure if we should put such detailed parameters for BTP commands, are those required by tests?

Which ones? But yes, I'm afraid the tests require that level of detail. E.g. the BAP Unicast Client Streaming Tests, BAP/UCL/STR/BV-xxx-C, the client needs to configure the server with a particular configuration. 

While I've added everything that looked like it would be needed, there's a chance that there might be a field or another which is not actually needed, but in general, all these are necessary.

MariuszSkamra:
@mringwal FYI, there's PR open https://github.com/auto-pts/auto-pts/pull/866 that adds the support for ASCS and BAP

mringwal:
As #866 has been merged and contains a ASCS service, this one has become obsolete. Closing.

PR #866, opened 2023-03-01
zephyr: Add LE Audio ASCS and BAP tests

mkasenberg:
Support for tests that use ASCS operations Codec Config, Config QOS, Enable, Disable, Update Metadata, Release.
TODO: Start streaming, sending data.


autopts/wid/bap.py
```diff
@@ -0,0 +1,768 @@
...
+
+    return ltv
+
+
+def parse_pac_char_value(data):
+    ltvs = {}
+    number_of_pac_records = data[0]
+    i = 1
+
+    for n in range(0, number_of_pac_records):
```
MariuszSkamra, line 262: Is it worth to check whether the data is not truncated?

doc/btp_ascs.txt
```diff
@@ -0,0 +1,220 @@
...
+
+		Controller Index:	<controller id>
+		Command parameters:	Address_Type (1 octet)
+					Address (6 octets)
+					ASE ID (1 octet)
+					Coding Format (1 octet)
+					Frequency (1 octet)
+					Frame Duration (1 octet)
+					Audio Locations (4 octets)
+					Octets Per Frame (2 octets)
```
mkasenberg, line 33: I was thinking of replacing these parameters with LTV values to make this command more flexible in case a codec with different parameters appears in the future in the PTS.
@sjanc

sjanc: Lets keep it as is now, I'd not expect anything else than LC3 being used in those tests
if needed we can make this "union" based on codec in future or just add new command

sjanc (approved):

sjanc (approved):

doc/btp_ascs.txt
mringwal: In all ASCS commands, the remote side is specified via { Address Type / Address }. I assume even Zephyr has some kind of ASCS Client Connect function that will result in some kind of connection identifier. We have used a Chan_ID in our ASCS BTP Doc draft. https://github.com/auto-pts/auto-pts/pull/860/files

What's your opionen on using { address type / address } vs. a connetion identifier (Chan_ID)?

doc/btp_ascs.txt
```diff
@@ -0,0 +1,220 @@
...
+
+        Valid Coding Format values:
+                    0x06 = LC3
+                    0x0ff = Vendor Specific
+
+        The values of the other parameters as defined in Assigned
+        Numbers (2023), 6.12.5 Codec_Specific_Configuration LTV
+        structures, i.e.:
+
+        Valid Frequency values:
```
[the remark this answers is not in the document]
mkasenberg: Actually, I plan to convert these fields to LTV values to make the command more flexible. I consulted a Zephyrs LE Audio developer and there is a high chance that there will be another codec in the future, with different codec specific configuration.

doc/btp_ascs.txt
mkasenberg: In Zephyr there is a similar connection ID, but it is used only internally, not visible to public API. I have to lookup the connection reference object by address.

mringwal: Unless the internal connection ID is a single byte, I also need to lookup, that's fine.

I was wondering more about the fact, that e.g. a ASCS Codec Configure requires: that an ASCS Client already did its job: find the service and characteristics, subscribe to all etc... without an explicit ASCS Connect command, this needs to happen implicitly. In theory, every ASCS command would first need to trigger an ASCS 'connect' operation, before the actual command can be executed. 

Another detail: while there is only one ASCS Client per remote device, there can e.g. multiple Media Control Services, so we'd need a way to select one of them for sending commands. By using a general "SERVICE connect" scheme, which might contain additional parameters, the actual commands can be implemented in a similar way.

MariuszSkamra: @mringwal Chan_ID would be something not specified by ASCS. However I think that one could consider structure like:
```
Address_Type (1 octet)
Address (6 octets)
ASE ID (1 octet)
```
to be Chan_ID.

mkasenberg: I see I'm a bit behind in testing compared to you. Can you provide an example test case number for which this is needed so I can read the scenario?

mringwal: @mkasenberg I didn't get there as well, but my colleague currently implement client/server for MCS, and you can have one MCS Service per Media Player. A quick glance at the MCSP TS doesn't give me a case with multiple parallel instances.

So, I have no hard facts here, just the idea that an explicit connect would be easier to debug and it's really not hard to imagine test cases where this could be necessary.

mringwal: Thanks for your reply. 

Instead of JLink RTT and btmon, I'm using local HCI log files that combine these, but it's the same here otherwise. As I'm switching between projects/customers often, it takes me significant time to remember how to debug AutoPTS (which is mainly my problem, but anything that simplifies this for me, probably also might helper others/beginners).

I did't have BAP IUT commmands, instead just ASCS Client for configuration and GAP to create CIS.
However, we don't have a BAP Client yet, as we're still trying to figure out how everything works together.

As for additional debug info, I think that it could be a good idea. It would allow to have some higher-level IUT event (message) in the auto-pts log. Do you consider a simple core event, e.g. "0x80 - IUT Message"?

With this changes already merged, I'll try to update my code to use that and compare status.

sjanc: Hi,

I like the idea of adding event in core service with IUT logs, that reminds me what linux is doing with BT logging injected into HCI snoop (and those are quite useful when debugging linux BT stack) .

There are some details to discuss ie if we want BTP command to enable IUT logs or tune levels. But I think those details can be discussed in PR thats add this to BTP

PR #867, opened 2023-03-02
Update mesh pts workspace

m-alperen-sener:
This PR is to update PIXIT values for mesh to be prepared for PTS 8.4. Changes doesn't effect PTS 8.3 which is currently in use.

- Adding wid 234 for resolved test case MESH/NODE/IVU/BI-06-C in PTS 8.4
- PTS 8.4 doesn't allow using default sample public and private keys.


autopts/wid/mesh.py
```diff
@@ -958,6 +958,13 @@ def hdl_wid_255(_: WIDParams):
     """
     return True
 
+def hdl_wid_234(_: WIDParams):
+    """
+    Implements:
+    description: Please confirms that the IUT receives the beacon with iv index
+                 set to 12
+    """
+    return True
```
KKopyscinski, line 967: It would be nice to have some verification here, but in Zephyr there is no callback for receiving beacons, IVI and other things from packet are just pulled and passed into `net_beacon_resolve`, and I don't see there any callback either, everything is done in backend. I guess we need to live with this like that. `bt_mesh_scan_cb` also doesn't give app access to itself, so we cannot see what Mesh scanned (maybe it's good, it'd be a lot of events)

KKopyscinski (approved):

autopts/wid/mesh.py
KKopyscinski: I created an issue to add API for testing this: https://github.com/zephyrproject-rtos/zephyr/issues/55366

m-alperen-sener: I considered the same concerns, however the latter steps in test implicitly verifies this step. But having an method to access that info from stuck better help to make sure intermediate steps.

sjanc:
Hi,

I wonder if we need to update zephyr-master.pqw6...  it may cause issues on workspaces update

m-alperen-sener:
The changes is only to make workspace file in sync with mesh.py in which the PIXIT values are overwritten. Actually I suggest to move new PTS release and TCRL, however, even though mesh stack is OK I am not sure about the rest test domains. As long as the  workspace file is in sync with mesh.py I don't think that it will cause an issue.

sjanc:
Yeap, I've just merged https://github.com/auto-pts/auto-pts/pull/875 which updated workspace to PTS 8.4 and test plan to TCRL 2022-2

sjanc:
and note that workspace is (should be) based on default values and modifications to ixits are done by autopts when tests are executed

m-alperen-sener:
I ll update the PR.

sjanc (approved):

PR #877, opened 2023-03-13
Fix mesh btp mesh_cfg_netkey_get cmd

m-alperen-sener:
added missing net_hey_idx parameter to netkey_get


sjanc (approved):

m-alperen-sener:
My mistake, I updated the doc.

KKopyscinski (approved):
I guess it was working because net_key_idx is usually zero in tests, but it surely has to passed as argument :)

PR #885, opened 2023-03-20
Zephyr WID 173

sjanc:


autopts/ptsprojects/zephyr/gap.py
```diff
@@ -59,6 +59,7 @@ class CHAR:
 iut_svc_data = '1111'
 iut_flags = '11'
 iut_svcs = '1111'
+iut_uri = UriScheme.https + 'github.com/intel/auto-pts'.encode()
```
michal-narajowski, line 62:
```suggestion
iut_uri = UriScheme.https + 'github.com/auto-pts/auto-pts'.encode()
```

michal-narajowski (approved):

autopts/ptsprojects/zephyr/gap.py
sjanc: yeah, I'd have to update ixit with encoded value,  will do that in separate PR

michal-narajowski (approved):

autopts/ptsprojects/zephyr/gap.py
sjanc: I've updated anyway :)

PR #905, opened 2023-04-19
bot: common: Don't run all remaining tests with default config

m-alperen-sener:
In cases default config already has intentionally defined test cases in it we should not run all the remaining test cases with default config. Only if default config has no specific test cases we might run all remaining test cases with it.


autopts/bot/common.py
```diff
@@ -137,10 +137,6 @@ def run_test_cases(self):
```
mkasenberg, line 137: I would add also smth like this:
```suggestion

        # Ensure the default config has the lowest priority
        iut_config_items = list(self.iut_config.items())
        for i, value in enumerate(iut_config_items):
            if value[0] == self.config_default:
                iut_config_items.pop(i)
                iut_config_items.append(value)

```
Otherwise, we'd have to remember that the default configuration in the iut_config dictionary has to be at the very end.

mkasenberg (changes requested):

autopts/bot/common.py
```diff
@@ -137,10 +137,6 @@ def run_test_cases(self):
 
         # Distribute test cases among .conf files
         for config, value in list(self.iut_config.items()):
```
mkasenberg, line 139: and then
```suggestion
        for config, value in iut_config_items:
```

autopts/bot/common.py
m-alperen-sener: Can you elaborate a bit more please why should we have priorities between configs and why the order they run matters. I don't see any point in running tests with order. In my understanding and most of the use case default config is the base test list that should run. I don't see any gain in that.

autopts/bot/common.py
m-alperen-sener: and also that?

autopts/bot/common.py
m-alperen-sener: I see the point, will update the PR. Thanks a lot for the explanation.

autopts/bot/common.py
```diff
@@ -152,21 +157,18 @@ def run_test_cases(self):
             _args[config].excluded = []
             _args[config].test_cases = []
 
-            remaining_test_cases = []
```
mkasenberg, line 155: Please take a note that in python this line: `remaining_test_cases = filtered_test_cases` will assign a reference to a variable, not create a copy of the list. Because of that, `remaining_test_cases.remove(tc)` will result in skipping next test case from the list. Here is an example you can run to see what will happen:
```python
test_cases = [
    'GAP/BROB/BCST/BV-01-C',
    'GAP/BROB/BCST/BV-02-C',
    'GAP/BROB/BCST/BV-03-C',
    'GAP/BROB/BCST/BV-04-C',
    'GAP/BROB/BCST/BV-05-C',
    'GAP/BROB/BCST/BV-06-C',
    'GAP/BROB/BCST/BV-07-C',
]

test_cases_not_deep_copy = test_cases

for tc in test_cases:
    print(tc)
    if tc == 'GAP/BROB/BCST/BV-04-C':
        test_cases.remove(tc)

print('test_cases_not_deep_copy:', test_cases_not_deep_copy)
print('test_cases:', test_cases)
```
In the example the 'GAP/BROB/BCST/BV-05-C' is skipped in `for` enumeration, because of the remove(), and the `test_cases_not_deep_copy` does not have 'GAP/BROB/BCST/BV-04-C' any more too.

autopts/bot/common.py
mkasenberg: Missing use of the `iut_config_items`

mkasenberg (changes requested):

mkasenberg (approved):

PR #906, opened 2023-04-20
iutctl_common.py: Fix encoding error

piotrnarajowski:
Error was occuring on machines with encoding cp1250:
 File "Python\Python310\lib\encodings\cp1250.py", line 19, in encode
     return codecs.charmap_encode(input,self.errors,encoding_table)[0]
     UnicodeEncodeError: 'charmap' codec can't encode character '\xf5'
     in position 16
     1: character maps to <undefined>

### DIFF
--- a/autopts/pybtp/iutctl_common.py
+++ b/autopts/pybtp/iutctl_common.py
@@ -19,6 +19,8 @@ import queue
 import socket
 import sys
 import threading
+import binascii
+
 
 from abc import abstractmethod
 
@@ -87,9 +89,10 @@ class BTPSocket:
             data_memview = data_memview[nbytes:]
             toread_data_len -= nbytes
 
-        tuple_data = bytes(str(dec_data(data)), 'utf-8').decode("unicode_escape").replace("b'", "'")
+        data_string = binascii.hexlify(data).decode('utf-8')
+        data_string = ' '.join(f'{data_string[i:i+2]}' for i in range(0, len(data_string), 2))
+        log(f"Received data: { {data_string} }, {data}")
 
-        log("Received data: %r, %r", tuple_data, data)
         self.conn.settimeout(None)
         return tuple_hdr, dec_data(data)
 


autopts/pybtp/iutctl_common.py
```diff
@@ -87,9 +89,10 @@ def read(self, timeout=20.0):
             data_memview = data_memview[nbytes:]
             toread_data_len -= nbytes
 
-        tuple_data = bytes(str(dec_data(data)), 'utf-8').decode("unicode_escape").replace("b'", "'")
+        data_decode = binascii.hexlify(data)
+        data_string = re.sub(r"(?<=\S)(?=(\S\S)+\b)", " ", data_decode.decode('utf-8'))
```
mkasenberg, line 93: The solution works, but has a very high computational complexity.
Please test out this example:
```python
import re
import time

test_bytes = ['0' for i in range(1000000)]
test_string = ''.join(test_bytes)

start = time.time()
result = ' '.join(f'{test_string[i:i+2]}' for i in range(0, len(test_string), 2))
end = time.time()
# Printed 0.3820197582244873 seconds
print(end - start)

start = time.time()
result = data_string = re.sub(r"(?<=\S)(?=(\S\S)+\b)", " ", test_string)
end = time.time()
# 5 minutes passed and I am still waiting for this print
print(end - start)
```
So my suggestion:
```suggestion
        data_string = binascii.hexlify(data).decode('utf-8')
        data_string = ' '.join(f'{data_string[i:i+2]}' for i in range(0, len(data_string), 2))
```

mkasenberg (changes requested):

autopts/pybtp/iutctl_common.py
```diff
@@ -19,6 +19,8 @@
 import socket
 import sys
 import threading
+import binascii
+import re
```
mkasenberg, line 23: a leftover
```suggestion
```

mkasenberg (approved):

PR #910, opened 2023-04-28
bot: server: Add swap_ets option

mkasenberg:
Experimental feature for swapping original .ets files with those received in PTS errata requests.
Ready for testing, but let's keep it draft for a while.

To use, enable the option in config.py in 
z['auto_pts']  = {
...
    'swap_ets': True,
}

Then put inside the ets/ folder your errata .ets files, i.e.:
ets
├── BAP.ets
├── BAP_UCL_SCC_BV_033_C.ets
└── GAP.ets

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@
 */client_secret.json
 */credentials.json
 autopts/bot/config.py
+*.ets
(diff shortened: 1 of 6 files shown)


mkasenberg:
This should be implemented on PTS side as IXIT.

PR #938, opened 2023-06-07
iutctl: Fix infinite loop when reading frame data from socket

laxiLang:
Sometimes the socket reading task would get stuck in an infinite loop. This issue has been handled for the frame header reading but not for the frame data, which may affect GATT/SR/GAW/BV-14-C on nRF53.


laxiLang:
@sjanc : will you review or assign a reviewer? This repo cannot assign reviewers automatically, so PR will be pending for some days possibly.

sjanc (approved):

PR #940, opened 2023-06-12
Change BTP L2CAP Listen command to not include response parameter

KKopyscinski:
This was workaround for PTS issue, now not needed. `btp.l2cap_listen` functoin left for now as-is to not cause regressions in Zephyr - application needs to adopt this change in BTP spec


sjanc:
So I'm not sure if we should remove the param (mainly due to other IUTs).

If you want to be able to remove this knowledge from specific ptsproject, we can simply add 0xff value which would mean "IUT specific" or "don't care"

PR #942, opened 2023-06-13
bot: mynewt: Add GAP/CONN/ACEP/BV-01-C into privacy overlay

szymon-czapracki:
Adding privacy option with this test enables
GAP/CONN/ACEP/BV-03-C to pass.


sjanc (changes requested):
according to TS that test shall be using public or static address...

PR #968, opened 2023-07-06
mynewt: change val_len check in wid_151

szymon-czapracki:
Add specific wid_151 for MyNewt with different approach for checking length. Changing generic one could break stuff while testing zephyr (thanks @KKopyscinski).


szymon-czapracki:
@sjanc FYI

PR #971, opened 2023-07-10
tests: csis: Added support for testing CSIS.

MSEEHenrik:
- Initialize service
- Set and clear member lock.
- Get member RSI.


doc/btp_csis.txt
```diff
@@ -0,0 +1,59 @@
+CSIS Service (ID 17)
+===================
+
+Commands and responses:
+
+	Opcode 0x00 - Error response
+
+	Opcode 0x01 - Read Supported Commands command/response
+
+		Controller Index:	<controller id>
```
sjanc, line 10: should be  \<non-controller\>

doc/btp_csis.txt
```diff
@@ -0,0 +1,59 @@
...
+		This command is used to set the member Lock.
+
+		In case of an error, the error response will be returned.
+
+	Opcode 0x03: CSIS Get Member RSI
+
+		Controller Index:	<controller id>
+		Command parameters:	Address_Type (1 octet)
+					Address (6 octets)
+		Response parameters:	Status (4 octets)
```
sjanc, line 48: This is a bit unorthodox for BTP since for errors we have dedicated error response, so your return RSI on success and BTP error response on error

MSEEHenrik: No - It will always return a response with Status and RSI.  It will never return a BTP error response.
As stated below - "In case of an error, the Status field in the response will reflect the error."

doc/btp_csis.txt
MSEEHenrik: Done.

doc/btp_csis.txt
sjanc: generic BTP code in zephyr and mynewt will return error eg. if size mismatch.  I'd prefer if we keep BTP commands behavior consistent regarding error handling.

Is this status some mesh specific status (4 bytes)?

MSEEHenrik: No - this is the normal int error code returned by bt_csip_set_member_generate_rsi().

MSEEHenrik: Changed the CSIS Get Member RSI function to comply to the normal error response functionality.

mkasenberg (approved):

PR #973, opened 2023-07-13
BTP core: Add IUT ready event

agata-ponitka:


mkasenberg (approved):
Finally this log `Unhandled event! svc_id 0 op 128` disappeared, nice!

PR #993, opened 2023-08-28
cron: bot: Post a comment under PR in case of bot failure

mkasenberg:
Until now, a message about an error was sent only to the admin e-mail. Let's post a comment under a PR that triggered a failing PR job, indicating that the results will not be delivered due to an error.


sjanc (approved):
I like this idea,  feel free to merge after resolving conflicts

PR #1018, opened 2023-10-05
mesh: Fix Remote Provisioning tests with persistence storage

PavelVPV:
This PR fixes MESH/SR/RPR/BV-02/03/04-C and MESH/SR/RPR/BI-01/02/03-C tests. Mesh initialization is moved to init command instead of service registration command to be able to select composition data that needs to be used by the stack. This is required by RPR tests when new composition data needs to be loaded.

Corresponding PR to zephyr: https://github.com/zephyrproject-rtos/zephyr/pull/63556

PR #1026, opened 2023-10-07
mesh: Fix provisioning context initialization

PavelVPV:
This fixes regression introduced in
https://github.com/auto-pts/auto-pts/pull/1018

In the PR above `bt_mesh_init()` call was moved to `MESH_INIT` command to allow to select alternative composition data when starting the stack. AutoPTS sends `MESH_CONFIG_PROVISIONING` command before `MESH_INIT` to prepare provisioning context. But this configuration is reset to the default configuration in `bt_mesh_init()` which makes PTS tests to fail.

To solve this, this commit calls introduces a new command `MESH_START` (`btp.mesh_start()`), which will replace the original `MESH_INIT` (`btp.mesh_init()`) command. `btp_mesh_init()` should now always be called after `btp.core_reg_svc_mesh()` call to initialize the stack. All `btp.mesh_init()` calls are replaced by `btp.mesh_start()`.

Zephyr PR: https://github.com/zephyrproject-rtos/zephyr/pull/63658


Andrewpini (approved):

sjanc (commented):
looks good (I like it better than delayed init command) but please update docs/btp_mesh too

BTW we may soon run out of opcodes in this service...

PavelVPV:
> BTW we may soon run out of opcodes in this service...

We can register MESH2 service.

sjanc (approved):

PR #1027, opened 2023-10-09
mesh: fix rpr link 13 lt2 usage

alxelax:
MESH/SR/RPR/LNK/BV-13-C was implemented initially wrong in PTS. Test should use both LT and LT2.
PTS patch has been provided and will be available in the next PTS release.


alxelax:
Related PTS ticket: https://support.bluetooth.com/hc/en-us/requests/28486

sjanc:
can it be merged now or should we wait for next PTS release?

alxelax:
It is better to merge now, then test will pass with the patch that can be found on the provided link with current PTS. 
The change will not break Mesh 1.0.1 CI compatibility anycase.
FYI @m-alperen-sener

sjanc:
BTW PTS support request are private ;-)

sjanc (approved):

PR #1032, opened 2023-10-12
Use `realpath` instead of `abspath`

theob-pro:
`realpath` resolve symlink which may end up in an exception on Windows when sharing a filesystem with Linux.

Use `abspath` instead.


mkasenberg (approved):

theob-pro:
Rebased to update commit title.

PR #1047, opened 2023-10-25
ptscontrol: Do not close PTSSender on Cancel response

mkasenberg:
Some WIDs return a Cancel response to select a different test variant. Let's use a different key word, END_TEST_CASE, to block sending next MMIs to the autopts client.


sjanc (approved):
works for me!:)

Issue #1013, opened 2023-10-02
cron: Add function to merge multiple TestCase.db files


mkasenberg:
If a test case exists in both .db files, take a result from database with highest 'count'
![image](https://github.com/auto-pts/auto-pts/assets/37016009/5343e9b8-5d2a-4f76-8843-adf1b70449ab)

PR #1061, opened 2023-11-22
tool: Merge TestCase.db files

agata-ponitka:
Add tool to merge TestCase.db files.
If a test case exists in both files, the one with the highest 'count' is taken into output.db file.


agata-ponitka:
Fix #1013

PR #1067, opened 2023-12-01
zephyr: Add support for CAP client test cases

mkasenberg:
enabled in zephyr-master workspace.

Related PR https://github.com/zephyrproject-rtos/zephyr/pull/66040


autopts/wid/cap.py
```diff
@@ -32,6 +38,96 @@ def cap_wid_hdl(wid, description, test_case_name):
...
+    btp.cap_broadcast_source_start(source_id)
+
+    data = bytearray([j for j in range(0, 41)])
+
+    for i in range(1, 10):
+        try:
+            btp.bap_send(0, data)
+        except BTPError:
+            # Buffer full
+            pass
```
MariuszSkamra, line 126: Ditto. It's unclear why the data has to be sent in this wid context

doc/btp_cap.txt
```diff
@@ -0,0 +1,269 @@
...
+	Opcode 0x03 - Ordered Access
+		Controller Index:	<controller id>
+		Command parameters:	Address_Type (1 octet)
+					Address (6 octets)
+		Response parameters:	<none>
+
+		This command is used to execute the CSIP Ordered Access.
+		In case of an error, the error status response will be returned.
+		In case of a success, the IUT continues processing
+		the command asynchronously.
```
MariuszSkamra, line 46: Hmm, I have mixed feelings about that one, for 2 reasons:
1. It's CSIP procedure, so it should be placed in CSIP
2. The actual procedure is missing. The ordered access performs some procedure on set. It's unclear to me what this BTP command does

doc/btp_cap.txt
```diff
@@ -0,0 +1,269 @@
...
+		Response parameters:	<none>
+
+		This command is used to configure CISes locally before executing
+		the Unicast Audio Start procedure.
+		In case of an error, the error status response will be returned.
+
+	Opcode 0x05 - Unicast Audio Start
+		Controller Index:	<controller id>
+		Command parameters: CIG_ID (1 octet)
+					Set Type (1 octet)
```
MariuszSkamra, line 77: indent

doc/btp_cap.txt
```diff
@@ -0,0 +1,269 @@
...
+
+		This command is used to execute the Unicast Audio Start procedure.
+		In case of an error, the error status response will be returned.
+		In case of a success, the IUT continues processing the command
+		asynchronously.
+
+	Opcode 0x06 - Unicast Audio Update
+		Controller Index:	<controller id>
+		Command parameters: Stream Count (1 octet)
+					Update Data (varies)
```
MariuszSkamra, line 93: ditto

doc/btp_cap.txt
```diff
@@ -0,0 +1,269 @@
...
+					Retransmission Num (1 octet)
+					Max Transport Latency (2 octets)
+					Presentation Delay (3 octets)
+					Codec Config LTVs length (1 octet)
+					Metadata LTVs length (1 octet)
+					LTVs (varies)
+
+		Response parameters:	<none>
+
+		This command is used to configure CISes locally before executing
```
MariuszSkamra, line 70: As this command takes single CIS as parameter, I would say it's `used to configure CIS` and mention that it can be called multiple times per ASE.
This raises a question whether this shouldn't be called `Unicast Setup ASE` as it can be called once per ASE (while twice per CIS, as as the Sink/Source ASE might share single CIS)

doc/btp_cap.txt
```diff
@@ -0,0 +1,269 @@
...
+		In case of an error, the error status response will be returned.
+
+	Opcode 0x05 - Unicast Audio Start
+		Controller Index:	<controller id>
+		Command parameters: CIG_ID (1 octet)
+					Set Type (1 octet)
+
+		Valid Set Type values:
+					0x00 = Ad-Hoc Set
+					0x01 = CSIP Coordinated Set
```
MariuszSkamra, line 81: What's the meaning of this parameter?

doc/btp_cap.txt
```diff
@@ -0,0 +1,269 @@
...
+
+	Opcode 0x06 - Unicast Audio Update
+		Controller Index:	<controller id>
+		Command parameters: Stream Count (1 octet)
+					Update Data (varies)
+
+		where Update Data can contain one or multiple structures defined as:
+					Address_Type (1 octet)
+					Address (6 octets)
+					ASE_ID (1 octet)
```
MariuszSkamra, line 98: Is it allowed? The CAP spec mentions `7.3.1.6.1 Updating Context Type and CCID Value Metadata` only

doc/btp_cap.txt
```diff
@@ -0,0 +1,269 @@
...
+					Retransmission Num (1 octet)
+					Max Transport Latency (2 octets)
+					Presentation Delay (3 octets)
+					Flags (1 octet)
+					Broadcast Code (16 octet)
+		where:
+			- Source ID is a local identifier.
+			- Flags is a bit map of settings:
+				bit 0: Use encryption
+				bit 1: Use the given Broadcast Code
```
MariuszSkamra, line 172: If the encryption is requested, then Broadcast Code has to be given. So is `bit 1` needed?

doc/btp_cap.txt
```diff
@@ -0,0 +1,269 @@
...
+		Controller Index:	<controller id>
+		Command parameters:	Source ID (1 octet)
+					Sdu Interval (3 octets)
+					Framing (1 octet)
+					Max SDU (2 octets)
+					Retransmission Num (1 octet)
+					Max Transport Latency (2 octets)
+					Presentation Delay (3 octets)
+					Flags (1 octet)
+					Broadcast Code (16 octet)
```
MariuszSkamra, line 167: Isn't Broadcase_ID missing somewhere?

doc/overview.txt
```diff
@@ -108,4 +108,8 @@ ID  Name
  21 CAS Service
  22 MCP Service
  23 GMCS Service
+<<<<<<< HEAD
  24 HAP Service
+=======
+ 24 CAP Service
+>>>>>>> dfe6fa28 (zephyr: Add support for CAP client test cases)
```
MariuszSkamra, line 115: Rebase problem

MariuszSkamra (changes requested):

autopts/ptsprojects/zephyr/cap.py
```diff
@@ -136,46 +145,588 @@ def test_cases(ptses):
         TestFunc(stack.aics_init),
         TestFunc(btp.core_reg_svc_vocs),
         TestFunc(stack.vocs_init),
+        TestFunc(btp.core_reg_svc_gmcs),
+        TestFunc(stack.gmcs_init),
```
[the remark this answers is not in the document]
mkasenberg: A leftover

autopts/wid/bap.py
```diff
@@ -312,6 +316,15 @@ def hdl_wid_114(params: WIDParams):
...
 
+    data = bytearray([j for j in range(0, 41)])
+
+    for i in range(1, 100):
+        try:
+            btp.bap_send(0, data)
+        except BTPError:
+            # Buffer full
+            pass
+
```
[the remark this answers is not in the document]
mkasenberg: PTS does not send an explicit message, but for test cases with SINK it looks like it expects to receive any ISO data and will not continue the test case, nor send a next wid.

autopts/wid/bap.py
```diff
@@ -1144,108 +1198,144 @@ def get_audio_locations_from_pac(addr_type, addr, audio_dir):
...
-
     stack = get_stack()
+    cig_id = 0x00
+
+    if params.test_case_name.endswith('LT2'):
+        addr = lt2_addr_get()
+        addr_type = lt2_addr_type_get()
+        cis_id = stack.bap.ase_configs[-1].cis_id
+    else:
+        cis_id = 0x00
```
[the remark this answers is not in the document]
mkasenberg: Plus, there is a bug here, should be `cis_id = stack.bap.ase_configs[-1].cis_id + 1`

doc/btp_cap.txt
mkasenberg: Related to
```
7.3 CAP procedures
(...)A set of Acceptors may be a Coordinated Set identified by the 
procedures in Section 7.4, or it may be an ad-hoc set where the members are determined by the 
implementation.
```
but RFU for now.

doc/btp_cap.txt
mkasenberg: I created the CSIP BTP and renamed the command to Start Ordered Access.

doc/btp_cap.txt
mkasenberg: I think `Update Data` is misleading. The address and ASE_ID is used only for selection of the right instance.

doc/btp_cap.txt
mkasenberg: I will add the field as RFU, because Zephyr stack generates the ID by its own

doc/btp_cap.txt
mkasenberg: A typo, the bit was meant to be used for `BTP_CAP_BROADCAST_SOURCE_SETUP_FLAG_SUBGROUP_CODEC = 0x02`

autopts/ptsprojects/zephyr/cap.py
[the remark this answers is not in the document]
mkasenberg: However, no. I forgot that it is needed to create a second CCID in Zephyr stack. I will add a comment about this.

PR #1083, opened 2023-12-21
zephyr: Allow RTT logs even while debugging

mkasenberg:
The RTT logs have not been enabled if the --gdb option was used.


MariuszSkamra (approved):
LGTM

PR #1107, opened 2024-02-05
zephyr: tester: OTS tests

piotrnarajowski:
Add support for OTS server tests


piotrnarajowski:
@mkasenberg

piotrnarajowski:
@mkasenberg deleted BTP_SERVICE_ID_OTS and everything related

sjanc (approved):
at some point we should have dedicated OTS BTP interface which would allow to test it directly, but for now I guess we can do it that way

PR #1113, opened 2024-02-13
Update pts workspace for mesh

m-alperen-sener:
Mesh is by default support v1.1 currently. In this regard
this commit updates the workspace to reflect supported
features. Also MBT 1.0 and DFU 1.0 models and features
supported by Mesh stack are enabled.

Adding the errata for mesh

MESH

Mesh v1.0.1 is dropped and Mesh v1.1 is now selected.

Added features and models:
- Provisioning algorithm ECDH_P256 + HMAC_SHA256 + CCM_AES128.
- Node Provisioning Protocol Interface.
- Node address and composition refresh procedures.
- Private beacon server and client models.
- Remote provisioning server and client models.
- Composition data page 1 and 2.
- On-demand proxy server and client models.
- Solicitation PDU RPL configuration server model.
- SAR configuration server and client models.
- Opcodes aggregator server and client models.
- Large composition data server and client models.
- Advertising with private network and node identity.

MBT

Mesh binary large object transfer model is added. Both client and
server models are supported.

DFU

Mesh device firmware update models are added.
- Firmware update server with pull/push mode BLOB transfer.
- Firmware distribution server with pull/push mode BLOB
transfer, store OOB firmware procedure and multiple image
support.
- Firmware update client with BLOB transfer.

### DIFF
--- a/autopts/ptsprojects/zephyr/dfu.py
+++ b/autopts/ptsprojects/zephyr/dfu.py
@@ -59,7 +59,7 @@ def set_pixits(ptses):
     pts.set_pixit("DFU", "TSPX_device_uuid", "00000000000000000000000000000000")
     pts.set_pixit("DFU", "TSPX_device_uuid2", "001BDC0810210B0E0A0C000B0E0A0C00")
     pts.set_pixit("DFU", "TSPX_use_pb_gatt_bearer", "FALSE")
-    pts.set_pixit("DFU", "TSPX_iut_comp_data_page", "0")
+    pts.set_pixit("DFU", "TSPX_iut_comp_data_page", "2")
     pts.set_pixit("DFU", "TSPX_oob_state_change", "FALSE")
     pts.set_pixit("DFU", "TSPX_enable_IUT_provisioner", "FALSE")
     pts.set_pixit("DFU", "TSPX_Procedure_Timeout", "60")
--- a/autopts/ptsprojects/zephyr/mbt.py
+++ b/autopts/ptsprojects/zephyr/mbt.py
@@ -59,7 +59,7 @@ def set_pixits(ptses):
     pts.set_pixit("MBT", "TSPX_device_uuid", "00000000000000000000000000000000")
     pts.set_pixit("MBT", "TSPX_device_uuid2", "001BDC0810210B0E0A0C000B0E0A0C00")
     pts.set_pixit("MBT", "TSPX_use_pb_gatt_bearer", "FALSE")
-    pts.set_pixit("MBT", "TSPX_iut_comp_data_page", "0")
+    pts.set_pixit("MBT", "TSPX_iut_comp_data_page", "2")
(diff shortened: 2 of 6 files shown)


sjanc:
Please update launch Studio project first  and than PTS workspace will be updated based on that  (current workspace is, or at least should be, based on Launch Studio project from Zephyr's main branch)

This is to ensure that we have qualifiable (ie no conflicts) workspace as well as help to avoid issues when updating TCRL

see
https://github.com/zephyrproject-rtos/zephyr/tree/main/tests/bluetooth/qualification
https://github.com/zephyrproject-rtos/zephyr/pull/64239

m-alperen-sener:
The ICS and Launch Studio Project file PR: https://github.com/zephyrproject-rtos/zephyr/pull/68975

Did not touch the IXIT values in the workspace file; however, made sure that they are overwritten with correct values from AutoPTS.

m-alperen-sener:
They should be enabled when they are supported by tester app and AutoPTS. Workspace should represent the what is tested and confident  IMO.

PavelVPV:
> They should be enabled when they are supported by tester app and AutoPTS. Workspace should represent the what is tested and confident IMO.

Isn't it supported?
https://github.com/zephyrproject-rtos/zephyr/blob/f9778472105d756fff7d1e5b54353421d356ed43/tests/bluetooth/tester/src/btp_mesh.c#L1848
Also, some tests are passing (I just checked few).

PavelVPV:
> They should be enabled when they are supported by tester app and AutoPTS. Workspace should represent the what is tested and confident IMO.

https://github.com/zephyrproject-rtos/zephyr/pull/58313

m-alperen-sener:
> > Removed features:
> > Proxy client and GATT client.
> > GAP Requirements central and authorization procedure.
> 
> Should we touch this if it was enabled before? We don't qualify these features in Nordic but it is implemented in the upstream, so I'd keep it running even if we don't pay attention to these tests. Maybe someone else would like to qualify the proxy client.

I checked test case map for Proxy Client feature. There are 18 test cases for this feature to claim it is supported and qualifiable:. Only five of them pass:  https://github.com/zephyrproject-rtos/bluetooth-qualification/blob/340435e7373ac322c0136773854e28ec15942e1b/mesh/autopts_report/report.txt#L35-L52

Additionally, there is no integration test scope for this feature. So it seems that we don't know if this feature is up to date with the mesh specs and if it is fully supported.

I suggest enabling this feature (Proxy Client MESH 12/5) with at least tester and AutoPTS support, even though there is no plan for integration test scope.

PavelVPV:
> I suggest enabling this feature (Proxy Client MESH 12/5) with at least tester and AutoPTS support, even though there is no plan for integration test scope.

Sounds good to me.

m-alperen-sener:
update the workspace commit. Not disabling the features that are already enabled. Did the same here: https://github.com/zephyrproject-rtos/zephyr/pull/68975

sjanc (approved):

PR #1115, opened 2024-02-15
tools: add script to sort projects in PTS Workspace by name

mringwal:
We've created some PTS Workspaces manually over time and PTS keeps the order in which Profiles/Protocols are added, which leads to an unsorted list. This script sorts the "projects" in a .pqw6 PTS workspace.


sjanc (approved):
nice!  I was always irritated by this when imported new ICS settings

mkasenberg (approved):

PR #1178, opened 2024-05-24
tools: cron: Fix check_supported_profiles

mkasenberg:


mkasenberg:
Fixed with https://github.com/auto-pts/auto-pts/pull/1188

PR #1183, opened 2024-06-07
Support CAP Capturing and Rendering Control Procedures

mringwal:
Add missing WIDs for CAP/COM/CRC tests

btstack/cap.py is here: https://github.com/bluekitchen/auto-pts/blob/develop-le-audio-broadcast/autopts/ptsprojects/btstack/cap.py

### DIFF
--- a/autopts/wid/cap.py
+++ b/autopts/wid/cap.py
@@ -772,6 +772,14 @@ def hdl_wid_402(_: WIDParams):
     return True
 
 
+def hdl_wid_404(_: WIDParams):
+    """
+        Please wait to establish 2 connections. After that, Click OK to send notification.
+    """
+
+    return True
+
+
 def hdl_wid_405(params: WIDParams):
     """
         Please read lock characteristic.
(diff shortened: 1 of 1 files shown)


autopts/wid/cap.py
```diff
@@ -772,6 +772,14 @@ def hdl_wid_402(_: WIDParams):
     return True
 
 
+def hdl_wid_404(_: WIDParams):
+    """
+        Please wait to establish 2 connections. After that, Click OK to send notification.
```
sjanc, line 777: no need to actually wait for those connections here?

mringwal: Before this wid, all PTS are connected via the serialized auto-pts sync functions.

Even with the sync functions, this means that the dialog might be OK'ed before the second device is connected. 
It didn't seem to make a difference.

If I read the test spec correctly, some of the CAP/COM/CRC tests expects that a notification from one PTS triggers and update on the other PTS via the IUT, but PTS does not test for that. The new wids don't cheat (e.g. implementing this functionality manually in auto-pts), so I think it's good enough.

sjanc (approved):

PR #1187, opened 2024-06-14
bot: config: add repeat_until_fail param

piotrnarajowski:
This commit adds repeat_until_fail parameter that can be used when running autoptsclient_bot. This can be helpful when debugging and collecting logs from tests that randomly fail from time to time.


autopts/client.py
```diff
@@ -1220,6 +1220,7 @@ def run_test_cases(ptses, test_case_instances, args):
     test_cases = args.test_cases
     projects = ptses[0].get_project_list()
     retry_config = getattr(args, 'retry_config', None)
+    repeat = getattr(args, 'repeat_until_fail', False)
```
mkasenberg, line 1223: a little suggestion
```suggestion
    repeat_until_fail = getattr(args, 'repeat_until_fail', False)
```

autopts/client.py
```diff
@@ -1259,6 +1260,9 @@ def run_test_cases(ptses, test_case_instances, args):
             else:
                 retry_limit = args.retry
 
+            if repeat and status == 'PASS':
```
mkasenberg, line 1263:
```suggestion
            if repeat_until_fail and status == 'PASS':
```

mkasenberg (approved):

autopts/client.py
piotrnarajowski: Done

PR #1217, opened 2024-08-07
cron: Restart bot again in case of timeout

mkasenberg:
Rarely, but it happens that restart of the testing setup fails, e.g. because of some TCP/IP socket. Let's try restart the bot setup again if bot terminated unexpectedly.

### DIFF
--- a/tools/cron/common.py
+++ b/tools/cron/common.py
@@ -587,6 +587,7 @@ def _run_test(config):
     backup = config['auto_pts'].get('use_backup', False)
     timeguard = config['cron']['test_run_timeguard']
     results_file_path = config['file_paths']['TC_STATS_JSON_FILE']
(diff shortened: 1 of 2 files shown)


tools/cron/common.py
```diff
@@ -531,6 +531,7 @@ def terminate_processes(config):
 
     if 'active_hub_server_start_cmd' in config['cron']:
         terminate_process(cmdline='active_hub_server.py')
+        sleep(5)
```
KKopyscinski, line 534: maybe in `terminate_process` we should call `process.wait()` after `process.terminate`?

KKopyscinski: and of course we won't need to sleep here, I guess we're sleeping here for terminate to actually be done?

KKopyscinski (approved):

PR #1219, opened 2024-08-08
tools: bisect: Allow to set the path of bisected repo

mkasenberg:


tools/cron/autopts_bisect.py
```diff
@@ -125,7 +125,10 @@ def bisect(cfg, test_case, good_commit, bad_commit=''):
     included = '-c {} '.format(test_case)
 
     cfg_dict, cfg_path = load_cfg(cfg)
```
KKopyscinski, line 127: `cfg_dict, _= load_cfg(cfg)` ?

KKopyscinski (approved):
nitpick, otherwise LGTM

PR #1233, opened 2024-08-23
zephyr: Add support for the nRF5340 Audio devkit

Frodevan:
Adds the nRF5340 audio devkit board, with capabilities to build both application and network cores with configurable separate builds.

### DIFF
--- a/autopts/ptsprojects/boards/__init__.py
+++ b/autopts/ptsprojects/boards/__init__.py
@@ -171,12 +171,22 @@ def get_free_device(board=None):
     """Returns tty path and jlink serial number of a free device."""
     devices = get_device_list()
 
+    ret_snr = None
+    ret_tty = None
+
     for tty, snr in devices.items():
         if tty not in devices_in_use and len(snr) >= 9 and snr.isnumeric():
-            devices_in_use.append(tty)
-            return tty, snr
+            ret_snr = snr
+            ret_tty = tty
+            # Opposite enumeration-order for TTY to coproccessor cores on nRF5340 Audio devkit.
+            if board != 'nrf53_audio':
+                break
 
-    return None, None
+    if ret_tty is not None:
+        devices_in_use.append(ret_tty)
+
+    log("Got free rtt for device {}: {}".format(ret_snr, ret_tty))
+    return ret_tty, ret_snr
 
 
 def get_debugger_snr(tty):
@@ -195,7 +205,7 @@ def get_debugger_snr(tty):
     return jlink
 
 
-def get_tty(debugger_snr):
+def get_tty(debugger_snr, board=None):
     """Return tty or COM of the device with given serial number.
     """
     tty = None
@@ -204,8 +214,11 @@ def get_tty(debugger_snr):
     for dev in devices.keys():
         if devices[dev] == debugger_snr:
             tty = dev
-            break
+            # Opposite enumeration-order for TTY to coproccessor cores on nRF5340 Audio devkit.
+            if board != 'nrf53_audio':
+                break
 
+    log("Got tty for device {}: {}".format(debugger_snr, tty))
     return tty
 
 
--- a/autopts/ptsprojects/boards/nrf53_audio.py
+++ b/autopts/ptsprojects/boards/nrf53_audio.py
@@ -1,0 +1,81 @@
+#
+# auto-pts - The Bluetooth PTS Automation Framework
+#
+# Copyright (c) 2024, Nordic Semiconductor ASA.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms and conditions of the GNU General Public License,
+# version 2, as published by the Free Software Foundation.
+#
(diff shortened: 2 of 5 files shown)


autopts/ptsprojects/boards/nrf53_audio.py
```diff
@@ -0,0 +1,80 @@
...
+                         debugger_snr,
+                         app_core_configs,
+                         True)
+
+    config_dir_net = os.getenv("AUTOPTS_SOURCE_DIR_NET")
+    if config_dir_net is None:
+        net_core_configs = [f'OVERLAY_CONFIG=\'nrf5340_cpunet_iso-bt_ll_sw_split.conf;'
+                            f'../../../tests/bluetooth/tester/nrf5340_hci_ipc_cpunet.conf\'']
+    else:
+        net_core_configs = [f'OVERLAY_CONFIG=\'../../../{config_dir_net}/hci_ipc.conf\'']
```
m-alperen-sener, line 74: Is it possible to use path as absolute rather than relative; or at least basing the path to zephyr_wd if it is possible?

m-alperen-sener (approved):
Seems reasonable; small comments.
Can we document AUTOPTS_SOURCE_DIR_NET and AUTOPTS_SOURCE_DIR_APP;

autopts/ptsprojects/boards/nrf53_audio.py
Frodevan: I mean, AUTOPTS_RESET_CMD is not documented in the board_ext file. I am not sure a good place to put this documentation, since it only applies to one group of boards, or just one board in particular.

However, I can try to have the path absolute. The build here will in any ways always be referenced from the hci_ipc source folder, so /../../../ should always be the zephyr root.

Frodevan: The relative path should now have been changed to absolute

Frodevan:
> Seems reasonable; small comments. Can we document AUTOPTS_SOURCE_DIR_NET and AUTOPTS_SOURCE_DIR_APP;

I mean, AUTOPTS_RESET_CMD is not documented in the board_ext file. I am not sure a good place to put this documentation, since it only applies to one group of boards, or just one board in particular.

theob-pro (approved):
👍

autopts/ptsprojects/boards/__init__.py
```diff
@@ -204,8 +211,8 @@ def get_tty(debugger_snr):
     for dev in devices.keys():
         if devices[dev] == debugger_snr:
             tty = dev
-            break
```
sjanc, line 207: why keep iterating here?

autopts/ptsprojects/zephyr/iutctl.py
```diff
@@ -205,6 +205,7 @@ def rtt_logger_start(self):
 
     def rtt_logger_stop(self):
         if self.rtt_logger:
+            time.sleep(0.1)
```
sjanc, line 208: why is this needed?

sjanc (commented):
I like the changes in general code, and we should also extend nrf5340DK same way,

but code changes are not matching commit message,  since there are more features than just audioDK support,  would be good to maybe split it for commit with proper commit message so that it is easier to figure eg. why changes to get_free_device() or get_tty() are needed

autopts/ptsprojects/zephyr/iutctl.py
Frodevan: In the case where an issue in the host has caused an operation to fail and in turn a test-case to fail, the rtt logger will often not have the time to pick up an error-print from the IUT before being closed. When this happens, the IUT logs will be incomplete, since any hints to the actual reason of the failure will be missing. This delay is to ensure enough time to prevent such a situation.

autopts/ptsprojects/boards/__init__.py
Frodevan: This may not be needed, but at the time I was adding that change I seem to remember that there was problems where it was picking the UART for the wrong core of the nRF5340 chip.

autopts/ptsprojects/zephyr/iutctl.py
sjanc: OK, I guess short comment in code would be good to not forget about this :)

Frodevan: Comment added

autopts/ptsprojects/boards/__init__.py
Frodevan: I added the conditional check, with a comment which should spread some light on why the change is needed to support the nRF5340 Audio devkit.

sjanc (approved):

autopts/ptsprojects/zephyr/iutctl.py
sjanc: btw with assert message it seems that 0.1 is not enough

PR #1236, opened 2024-08-26
zephyr: Fix multiple CAP/INI/UST tests

sjanc:
CAP WIDs are using MICP and VCP commands and thus shall initialize required BTP services.


sjanc:
@mringwal FYI

mkasenberg (approved):

PR #1238, opened 2024-08-29
tests: bas: Added support for testing Battery Service

MicImbriani:
Currently only support Mandatory Battery Level and optional Battery Level Status Characteristic.


doc/btp_bas.txt
```diff
@@ -0,0 +1,132 @@
...
+		is not supported.
+
+		In case of an error, the error response will be returned.
+
+Events:
+	Opcode 0x02 - BAS Set Battery Level
+
+		Controller Index:	<controller id>
+		Event parameters:	Address_Type (1 octet)
+					Address (6 octets)
```
sjanc, line 28: Hmm this is server configuration,   do we need address here? (normally address is needed for BTP only if it is suppose to performed action in specified peer)  

I'd remove address from all those commands

sjanc (changes requested):

autopts/wid/bas.py
```diff
@@ -0,0 +1,105 @@
...
+        Please Change the Battery Level field of Battery Level Status characteristic.
+    """
+    btp.bas_set_battery_level();
+    return True
+
+def hdl_wid_7(_: WIDParams):
+    """
+        Please Change the Power State field of Battery Level Status characteristic.
+    """
+    btp.bas_set_battery_present();
```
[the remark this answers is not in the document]
Thalley: One alternative (I don't really like either), is to not provide the value to the DUT and just make it a `bas_change_battery_present` instead, and let the DUT decide the value.

autopts/wid/bas.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    stack = get_stack()
+    btp.gap_set_conn()
+    btp.gap_adv_ind_on(ad=stack.gap.ad)
+    return True
+
+def hdl_wid_20108(_: WIDParams):
+    """
+        Please send notifications for Characteristic 'Battery Level Status' to the PTS.
+    """
```
Thalley, line 94: Does this come together with the others above? Are we in "risk" of changing the battery present value or level twice?

Thalley (changes requested):

autopts/wid/bas.py
```diff
@@ -0,0 +1,109 @@
...
+    global _step
+    _step = 1 if not _step else _step + 1
+
+    if _step == 1:
+        btp.bas_set_battery_present(battery_present=BAS_BATTERY_PRESENT.BATTERY_PRESENT)
+    elif _step == 2:
+        btp.bas_set_battery_level(battery_level=69)
```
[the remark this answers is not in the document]
MicImbriani: I like your approach but it would not work, because during the execution of test BAS/SR/CN/BV-20-C, the same hdl_wid function 'hdl_wid_20108' is called multiple times with different description, and expectind a different behaviour: sending notification for Characteristic 'Battery Level Status' first, and for 'Battery Level' second. Hence the 2-steps approach, which i would suggest to keep.

I agree that the description of the function currently doesn't match. I will update it to also add the second step's description, and maybe add a brief explanation (along the lines of what I said above).

autopts/wid/bas.py
```diff
@@ -0,0 +1,109 @@
...
+    btp.gap_set_conn()
+    btp.gap_adv_ind_on(ad=stack.gap.ad)
+    return True
+
+def hdl_wid_20108(_: WIDParams):
+    """
+        Please send notifications for Characteristic 'Battery Level Status' to the PTS.
+    """
+    global _step
+    _step = 1 if not _step else _step + 1
```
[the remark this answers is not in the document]
MicImbriani: Good suggestion, but the current implementation does not require a BAS instance to be initialized in the Stack, and doing so only to not have a global variable seems overkill, but I am open to change it if you think it is worth :)

autopts/wid/bas.py
MicImbriani: Do you mean, whether this hdl_wid function is called during the execution of a test that calls one of the previous functions? 
I have not come across this situation with the BAS tests I have worked with. But I suppose there could be this possibility with other tests I have not encountered

autopts/pybtp/btp/bas.py
```diff
@@ -0,0 +1,91 @@
...
+    return tuple_data
+
+def address_to_ba(bd_addr_type=None, bd_addr=None):
+    data = bytearray()
+    
+    bd_addr_ba = addr2btp_ba(pts_addr_get(bd_addr))
+    bd_addr_type_ba = chr(pts_addr_type_get(bd_addr_type)).encode('utf-8')
+    
+    data.extend(bd_addr_type_ba)
+    data.extend(bd_addr_ba)
```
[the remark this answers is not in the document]
MicImbriani: They can be None:

If `bd_addr` is `None `, `pts_addr_get()` will use `PTS_BD_ADDR.addr` as value
Same for `pts_addr_type_get()` which will use `PTS_BD_ADDR.addr_type` as value

autopts/pybtp/btp/bas.py
```diff
@@ -0,0 +1,91 @@
...
+    
+    return data
+
+def bas_set_battery_level(battery_level):
+    logging.debug("%s", bas_set_battery_level.__name__)
```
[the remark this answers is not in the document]
MicImbriani: I agree. 
However all the other modules are implemented similar to this. 
I would suggest keeping this logging as it is, and opening an issue ticket for requesting this change, so that it can be addressed in a separate PR

autopts/wid/bas.py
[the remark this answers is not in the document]
Thalley: Aha. That's terrible design from PTS :D Suggest to create a report requesting that different WID should be used here. 

We can probably grep the description for what needs to be done. Maybe something as simple as `description.contains("Battery Level Status")` or `description.contains("Battery Level")`. That would make it easier to follow what goes on IMO. Our behavior should IMO not depend on some global variable, as that is harder to maintain.

autopts/pybtp/btp/bas.py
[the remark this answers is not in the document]
Thalley: :upside_down_face: 

Suggest to keep as is then. Created https://github.com/auto-pts/auto-pts/issues/1344

autopts/wid/bas.py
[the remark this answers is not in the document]
Thalley: I wouldn't consider anything that removes a global variable as overkill :D

autopts/wid/bas.py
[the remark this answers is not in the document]
MicImbriani: Sounds like a good enough approach. I updated the code now.

autopts/pybtp/btp/bas.py
```diff
@@ -0,0 +1,91 @@
...
+import logging
+import struct
+
+from autopts.pybtp import defs
+from autopts.ptsprojects.stack import get_stack
+from autopts.pybtp.btp.btp import CONTROLLER_INDEX, get_iut_method as get_iut,\
+    btp_hdr_check, pts_addr_get, pts_addr_type_get
+from autopts.pybtp.types import addr2btp_ba
+
+log = logging.debug
```
Thalley, line 25: Unused variable `log`?

autopts/pybtp/btp/bas.py
```diff
@@ -0,0 +1,91 @@
...
+BAS = {
+	'set_battery_level': 	( defs.BTP_SERVICE_ID_BAS,
+                              defs.BAS_SET_BATTERY_LEVEL,
+                              CONTROLLER_INDEX),
+	'set_battery_present': 	( defs.BTP_SERVICE_ID_BAS,
+                              defs.BAS_SET_BATTERY_PRESENT,
+                              CONTROLLER_INDEX),
+	'set_charging_state': 	( defs.BTP_SERVICE_ID_BAS,
+                              defs.BAS_SET_BATTERY_CHARGE_STATE,
+                              CONTROLLER_INDEX),
```
Thalley, line 36: Indentation still seems off. Consider only using spaces for indentation

autopts/wid/bas.py
```diff
@@ -0,0 +1,114 @@
...
+from autopts.pybtp.types import WIDParams, BAS_BATTERY_PRESENT
+from autopts.wid import generic_wid_hdl
+
+log = logging.debug
+
+_step = 0
+
+
+def bas_wid_hdl(wid, description, test_case_name):
+    log(f'{bas_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
```
Thalley, line 30:
```suggestion
_step = 0


def bas_wid_hdl(wid, description, test_case_name):
    logging.debug(f'{bas_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
```

autopts/wid/bas.py
```diff
@@ -0,0 +1,114 @@
...
+from enum import IntFlag
+from autopts.pybtp import btp
+from autopts.ptsprojects.stack import get_stack
+from autopts.pybtp.types import WIDParams, BAS_BATTERY_PRESENT
+from autopts.wid import generic_wid_hdl
+
+log = logging.debug
+
+_step = 0
+
```
Thalley, line 27:
```suggestion

```

autopts/wid/bas.py
```diff
@@ -0,0 +1,114 @@
...
+    btp.bas_set_battery_level(battery_level=10)
+    return True
+
+def hdl_wid_34(_: WIDParams):
+    """
+        Please change the value of Battery Level characteristic.
+    """
+    btp.bas_set_battery_level(battery_level=1)
+    btp.bas_set_battery_level(battery_level=10)
+    return True
```
Thalley, line 83: Why are these changing the battery level twice?

autopts/wid/bas.py
```diff
@@ -0,0 +1,114 @@
...
+        2) Please send notifications for Characteristic 'Battery Level' to the PTS.
+    """
+
+    # This test calls wid_20108 multiple times, requiring a differnt behaviour each time
+    if param.test_case_name in ['BAS/SR/CN/BV-20-C']:
+        if 'Battery Level Status' in param.description:
+            btp.bas_set_battery_present(battery_present=BAS_BATTERY_PRESENT.BATTERY_PRESENT)
+        elif 'Battery Level' in param.description:
+            btp.bas_set_battery_level(battery_level=69)
+        else:
```
Thalley, line 107: Is there a plan to modify this later so that we don't depend on hardcoded values, but rather are aware of the current state and can use that to ensure that the state is modified?

Thalley (changes requested):

autopts/wid/bas.py
MicImbriani: Similar reasoning as before - to ensure that the value is indeed changed.

autopts/wid/bas.py
MicImbriani: There is no plan, at least not for me.
If you think that merging these changes can wait a bit longer, I can modify to implement your suggestion in this PR.

Thalley: I'll defer to the maintainer @sjanc 

Personally I'd keep a copy of the state in autopts. 
If we don't get events from the IUT about changes, then that can be done by setting a specific value as part of the `pre_conditions` and then store the IUTs state in the `STACK` object as we do for some other profiles.

Thalley: @sjanc any input? Keep as is, or implement a future proof version where the autopts client keeps track of the battery level (and status) to ensure that we trigger notifications?

sjanc: I agree that long term we should start avoiding hardcoded values and take those from client (this gives a bit more flexibility for various IUT implementations) but if @MicImbriani is short on time I think we can merge this as is, and fix in followup

sjanc (commented):
docs should be fixed and updated to follow up convention used

doc/btp_bas.txt
```diff
@@ -0,0 +1,210 @@
...
+		Each bit in response is a flag indicating if command with
+		opcode matching bit number is supported. Bit set to 1 means
+		that command is supported. Bit 0 is reserved and shall always
+		be set to 0. If specific bit is not present in response (less
+		than required bytes received) it shall be assumed that command
+		is not supported.
+
+		In case of an error, the error response will be returned.
+
+Events:
```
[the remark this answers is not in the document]
MicImbriani: Good spotting, thanks! I must have taken the starting template from the wrong doc file.

Thalley (approved):

doc/btp_bas.txt
sjanc: this is marked as resolved but address is stil there...

doc/btp_bas.txt
```diff
@@ -0,0 +1,234 @@
...
+					Address (6 octets)
+					Battery Present (1 octet)
+		Response parameters:	<None>
+
+		This command instruct the IUT to change the Power State field of 
+		the Battery Level Status characteristic.
+
+		BTP Values:
+			/** Battery is not present. */
+			BTP_BAS_BLS_BATTERY_NOT_PRESENT = 0,
```
sjanc: also I'd prefer if those follow style in other services (ie no C comments, formatting )

MicImbriani: should be good now.

doc/btp_bas.txt
```diff
@@ -0,0 +1,183 @@
...
+				is not supported.
+
+				In case of an error, the error response will be returned.
+
+		Opcode 0x02 - BAS Set Battery Level
+
+				Controller Index:		<controller id>
+				Command parameters:		Address_Type (1 octet)
+										Address (6 octets)
+										Battery Level (1 octet)
```
Thalley: Some indentation is off here and several other places

doc/btp_bas.txt
```diff
@@ -0,0 +1,183 @@
...
+										Address (6 octets)
+										Battery Present (1 octet)
+				Response parameters:	<None>
+
+				This command instruct the IUT to change the Power State field of 
+				the Battery Level Status characteristic.
+
+				BTP Values:
+						BTP_BAS_BLS_BATTERY_NOT_PRESENT = 0,
+						BTP_BAS_BLS_BATTERY_PRESENT = 1
```
Thalley: I think the format should be more like
```suggestion
				Battery Present:
					0 = Battery not present
					1 = Baterry present
```
or
```suggestion
				Battery Present:
					0	Battery not present
					1	Baterry present
```

To match most other commands. We don't usually refer to the values defined in Python (but rather use text), and the value always seems to come first (but there are quite a few different formats)

This applies to most other these commands

Thalley (changes requested):

PR #1243, opened 2024-09-06
bot: Add ptsgui mode

mkasenberg:
The ptsgui mode allows to run a test case with PTS in GUI mode. A user has to manually enter MMI(WID) ID and its description into AutoPTS dialog and deliver the response to the PTS dialog.

Useful while debugging issues with test cases like the GAP/SEC/SEM/BV-26-C, where 2 x MMI windows appear at once in PTS GUI mode, but in PTS automation mode they arrive in random order, because the PTS automation interface can handle only one at a time,

### DIFF
--- a/autopts/bot/common_features/report.py
+++ b/autopts/bot/common_features/report.py
@@ -28,8 +28,9 @@ import yaml
 
 from autopts.bot import common
 from autopts.bot.common_features import github
-from autopts.client import PtsServer
+from autopts.client import PtsDirectClient
 from autopts.config import AUTOPTS_ROOT_DIR
+from autopts.types import AutoPTSMode
 
 log = logging.debug
 
@@ -473,9 +474,9 @@ def pull_server_logs(args, tmp_dir, xml_folder):
             except BaseException as e:
                 logging.exception(e)
 
-    if args.server_args:
+    if args.autopts_mode == AutoPTSMode.AUTO_CLIENT_ONLY:
         # Logs available locally
-        _pull_logs(PtsServer)
+        _pull_logs(PtsDirectClient)
     else:
         # Use xmlrpc proxy to pull logs
         server_addr = args.ip_addr
--- a/autopts/client.py
+++ b/autopts/client.py
@@ -36,6 +36,8 @@ import time
 import traceback
 import xml.etree.ElementTree as ElementTree
 import xmlrpc.client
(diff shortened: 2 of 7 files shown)


sjanc (approved):
cool :)

autopts/bot/common_features/report.py
```diff
@@ -473,9 +473,9 @@ def _pull_logs(_pts):
             except BaseException as e:
                 logging.exception(e)
 
-    if args.server_args:
+    if args.autopts_mode == 'autopts_direct_client':
```
Thalley: Let's not use string literals for comparison :) Suggest to define the supported options and then use those defines (possibly as an StrEnum)

autopts/client.py
```diff
@@ -507,42 +522,49 @@ def init_pts_thread_entry(proxy, args, exceptions, finish_count):
...
 
 
-def init_pts(args, ptses):
-    """Initialization procedure for PTS instances"""
+PTS_FACTORY = PtsServerProxy
 
-    proxy_list = ptses
+
+def setup_pts_factory(args):
+    global PTS_FACTORY
```
Thalley: How is this a factory? Isn't it just an argparser?

autopts/client.py
```diff
@@ -560,6 +582,24 @@ def init_pts(args, ptses):
     if exeption_msg:
         raise Exception(exeption_msg)
 
+
+def init_pts(args, ptses):
+    """Initialization procedure for PTS instances"""
```
Thalley: Suggest to proper document all new functions and classes (preferably also using type hinting)

autopts/ptsguiproxy.py
```diff
@@ -0,0 +1,244 @@
...
+
+from autopts.client import ClientCallback
+from autopts.ptsprojects.ptstypes import MMI_STYLE_STRING
+
+
+log = logging.debug
+
+
+class WIDDialog:
+    def __init__(self, parent, callback):
```
Thalley, line 13: Ditto; please properly document new functions and classes

Thalley (dismissed):

Thalley (commented):
No blockers from me, but have not reviewed it in detail. Feel free to merge

mkasenberg:
Thanks!

@szymon-czapracki, does the current GUI layout and behavior work for you?

PR #1244, opened 2024-09-06
zephyr: sysbuild configuration for 53/54h

AyturkDuzen:
Using sysbuild, cpuapp build automatically triggers building hci_ipc build for cpurad. Preserve the old behaviour for 53 with no sysbuild.


sjanc:
Hi,

Could this be made configurable?  On occasions I find it useful to be able to bisect only controller or host, or is this possible with sysbuild?

AyturkDuzen:
> Hi,
> 
> Could this be made configurable? On occasions I find it useful to be able to bisect only controller or host, or is this possible with sysbuild?
Sysbuild takes the complexity of building/flashing multi-image to its own configs. There is a way to turn-off sysbuild but it also affects how we structure the tester app in zephyr by having Kconfig.sysbuild+sysbuild folder for hci_ipc so multi-core series might as default have sysbuild use.

sjanc (approved):

PR #1265, opened 2024-10-02
wid: btp: bap: Add BIS stopped event BPT callback

Frodevan:
Adds BTP callback for BIS stopped, and uses this to ensure a BIS stream has truly been stopped before advancing during BAP WID 379 and 380. These have been problematic if the controller is not fast enough closing the stream, since there were previously no check that the controller was truly ready to proceed.

This PR is synchronized with https://github.com/zephyrproject-rtos/zephyr/pull/79329


autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -107,6 +108,19 @@ def wait_bis_synced_ev(self, broadcast_id, bis_id, timeout, remove=True):
             lambda ev: (broadcast_id, bis_id) == (ev['broadcast_id'], ev['bis_id']),
             timeout, remove)
 
+    def wait_bis_stopped_ev(self, broadcast_id, bis_id, reason, timeout, remove=True):
```
Thalley, line 111: Since `reason` is seemingly optional to set, suggest to make it
```suggestion
    def wait_bis_stopped_ev(self, broadcast_id, bis_id, timeout, reason=None, remove=True):
```

Frodevan: I was thinking about it, but I want @sjanc's input on this since that breaks with the argument-ordering of many of the other functions in this file.

PR #1268, opened 2024-10-04
bot: imrpove rtt log

piotrnarajowski:
Work in progress.
This should stop rtt read process only if we are sure that test case has ended and there is no more data to send


autopts/rtt.py
```diff
@@ -233,9 +243,23 @@ def start(self, buffer_name, log_filename, device_core, debugger_snr):
...
+    def stop_timer(self):
+        # timeout reached
+        self.rtt_reader.rtt_stop = True
+
+    def stop(self, reset, timeout=0.1):
         log("%s.%s", self.__class__, self.stop.__name__)
-        self.rtt_reader.stop()
+
+        if reset:
+            self.rtt_reader.stop()
```
koffes, line 254: Should this wait even when when this is true?

autopts/rtt.py
```diff
@@ -57,14 +59,22 @@ def _get_buffer_index(self, buffer_name):
...
                 byte_list = jlink.rtt_read(buffer_index, 1024)
                 if len(byte_list) > 0:
+                    count = 0
                     user_callback(bytes(byte_list), user_data)
-        except BaseException:
+                elif self.test_end:
+                    count += 1
+                    if count >= threshold:
+                        # we have reached the threshold for consecutive empty reads, we can stop rtt_read
+                        self.rtt_stop = True
```
koffes, line 75: Seems to me that this loop of checking for 5 iterations may be too fast. 
Should there be a delay in here as well?

autopts/rtt.py
```diff
@@ -233,9 +243,23 @@ def start(self, buffer_name, log_filename, device_core, debugger_snr):
...
+
+        if reset:
+            self.rtt_reader.stop()
+        else:
+            self.rtt_reader.test_end = True
+            # let's wait until stopping rtt reader; timeout is configurable
+            timer = threading.Timer(timeout, self.stop_timer)
+            timer.start()
+            while not self.rtt_reader.rtt_stop:
+                time.sleep(0.1)
```
Thalley, line 261: Even with `timeout=1` I still didn't all the logs when running locally. Had to do
```suggestion
            while not self.rtt_reader.rtt_stop:
                time.sleep(1)
```

As well

autopts/rtt.py
piotrnarajowski: This could be implemented, but from what I understood the rtt closing issue occured at the end of the test case which is why I omitted this case. Ideally IUT should be reseted only at the beginning of test case, If you encounter unexpected IUT reset while in the middle/end of test case please let me know.

autopts/rtt.py
piotrnarajowski: We will enter this loop only after timeout provided with -rtt_logger_param is reached, so maybe incrementing it will help?

autopts/rtt.py
piotrnarajowski: So changing this helped to get all the logs?

Thalley: Yes, but I don't really understand why we would need a time.sleep here at all, if there's a timeout elsewhere.

piotrnarajowski: Please see updated version. made a few changes. Timer will be started when first empty packet is received and then cancelled if non-empty packet is received.

autopts/rtt.py
piotrnarajowski: Got rid of 5 iterations check. Instead,  Timer will be started when first empty packet is received and then cancelled if non-empty packet is received.

PR #1271, opened 2024-10-10
zephyr: Fix cap.hdl_wid_312 to use cap unicast stop

Thalley:
Modified the hdl_wid_312 handler to call cap_unicast_audio_stop with release = False, so that it performs a disable (and stop) without release.

Why PTS wants us to disable first is unclear, as the test spec states to release, which would be handled by hdl_wid_309.

This is required by https://github.com/zephyrproject-rtos/zephyr/pull/79674

### DIFF
--- a/autopts/pybtp/btp/cap.py
+++ b/autopts/pybtp/btp/cap.py
@@ -122,10 +122,18 @@ def cap_unicast_audio_update(metadata_tuple):
     cap_command_rsp_succ()
 
 
-def cap_unicast_audio_stop(cig_id):
+BTP_CAP_UNICAST_AUDIO_STOP_FLAG_RELEASE = 0b00000001
+
+def cap_unicast_audio_stop(cig_id, release):
     logging.debug(f"{cap_unicast_audio_stop.__name__}")
 
     data = struct.pack('B', cig_id)
+
+    flags = 0x00
+    if release:
+        flags |= BTP_CAP_UNICAST_AUDIO_STOP_FLAG_RELEASE
+
+    data += struct.pack('?', flags)
 
     iutctl = get_iut()
     iutctl.btp_socket.send(*CAP['unicast_audio_stop'], data=data)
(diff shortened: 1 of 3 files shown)


sjanc:
this breaks BTP command compatibility,  I checked and it doesn't affect any of our internal setup so I'm fine with this, but should probably be synced with @mringwal

mringwal:
@sjanc thanks! It would be nice to get notified about changes to the BTP spec. I just found this https://app.github-file-watcher.com/ and set-up a subscription for the doc/ folder. Let's see if that works as expected  :)

Thalley:
@sjanc what's the plan on merging the 2 PRs?

sjanc:
I think we can merge this as is (I'll merge it once zephyr PR is merged,  just approved it)

I'll also try to find time to figure our on how we should handle such things in long term,  there is BTP versioning issue created some time ago to cover this,  comments would be appriaciated

at very minimum we could have command for reading BTP version just fail early with big fat error (so that IUT can be updated),  alternatively we could follow what BT Core spec is not doing with HCI commands and have V1/V2/etc based on mentioned version,  that would make transition easier (maybe)

sjanc (approved):

Thalley:
> I'll also try to find time to figure our on how we should handle such things in long term, there is BTP versioning issue created some time ago to cover this, comments would be appriaciated

Yeah, I think we need something. 
The LE Audio BTP commands and events are quite Zephyr specific, so chances are that they will need refactoring when used by other platforms. 

As long as we _add_ and do not modify fields to events and commands, it should be possible for applications to handle multiple versions. 

For example in this case where the flags were added, any existing application using BTP could still just read the expected bytes and discard the rest, but not sure how the applications are implemented in other platforms than Zephyr

mringwal:
Good question how to address BTP changes. To me, a quick crash in the IUT is better than having a failing test cases that might have worked earlier before a command was modified. In this context, I'd prefer additional / v2 commands for changes that will cause existing implementations to fail. Not sure, if that's a an easy to assess criterium. 

Another option could be to have the IUT report a version per profile and the autopts clients aborts if the version is lower than expected. By this, I would know that I need to check the changes to BTP and get warned that I need to update our BTP implementation.

PR #1201, opened 2024-07-10
Add PBP Test Suite

mringwal:
This PR adds the PBP layer and WIDs for the PBP Test suite

It also fixes smaller details in other areas we've stumbled upon using one commit per fix.

PTS Project PBP code: https://github.com/bluekitchen/auto-pts/blob/develop-le-audio-tmap/autopts/ptsprojects/btstack/pbp.py

PR #1278, opened 2024-10-21
Add GMAP Test Suite

mringwal:
This PR adds the GMAP layer and WIDs for the GMAP Test suite

It uses BTP service ID 31 as our PBP PR (https://github.com/auto-pts/auto-pts/pull/1201) uses 30.

PTS Project GMAP code: https://github.com/bluekitchen/auto-pts/blob/develop-le-audio-gmap/autopts/ptsprojects/btstack/gmap.py


sjanc (commented):
also commit message could be less sparse ;-)

mringwal:
Ok, will do. It did apply cleanly in October and did follow the naming conventions back then.. 
The commit message is rather precise :)

sjanc:
> Ok, will do. It did apply cleanly in October and did follow the naming conventions back then..

yeah, sorry for that,  in general commands should follow BTP_<SERVICE>_CMD_FOO and events BTP_<SERVICE>_EV_BAR

This is for BTP trace parsing, which are now much more user friendly :)

PR #1291, opened 2024-11-04
errata: Add Request ID 147314 for CSIP and more

Thalley:
The Request ID 147314 applies to nearly all tests that uses multiple lower testers, as it affects how IRKs are handled by PTS.


sjanc (commented):
we could at least track it for CSIP tests (we don't have ***that*** many of them enabled:)

Thalley:
> we could at least track it for CSIP tests (we don't have _**that**_ many of them enabled:)

Sure. I can add the ones with 3 lower testers to the list at least :) That's a fairly small subset

PR #1312, opened 2024-12-02
server: Check that COM port exists

koffes:
- Check has been added to ensure that the selected com port(s) exist. pyserial was used as it was much faster than wmi

### DIFF
--- a/autoptsserver.py
+++ b/autoptsserver.py
@@ -51,6 +51,8 @@ from time import sleep
 import pythoncom
 import wmi
 
+import serial.tools.list_ports
+
 from autopts import ptscontrol
 from autopts.config import SERVER_PORT
 from autopts.utils import CounterWithFlag, get_global_end, exit_if_admin, ykush_replug_usb, ykush_set_usb_power, \
@@ -62,6 +64,13 @@ log = root_logging.debug
 log_inited = False
 PROJECT_DIR = dirname(abspath(__file__))
 
+def _com_port_exists(port_name):
+    """Check if the COM port exists."""
+    ports = serial.tools.list_ports.comports()
+    for port in ports:
+        if port.device == port_name:
+            return True
+    return False
 
 def init_logging(_args):
     """Initialize server logging"""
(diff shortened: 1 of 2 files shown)


autoptsserver.py
```diff
@@ -623,6 +632,10 @@ def shutdown_pts_bpv(self):
     autoptsservers = []
     server_count = len(_args.srv_port)
     finish_count = CounterWithFlag(init_count=0)
+    for i in range(server_count):
+        if not _com_port_exists(_args.dongle[i]):
+            sys.exit(f"Port {_args.dongle[i]} not found")
```
Thalley, line 637: I'm unsure whether this is the best place/way to do this. 

I'm not a Python expert, but the `_args.dongle` is effectively used to instantiate a PyPTS object with `device = args.dongle`. 

Given that the class allows `None` as the type, it seems that it's possible to actually instantiate a PyPTS without a dongle. Perhaps if the dongle is inserted later? 

Perhaps this check should be moved to `PyPTS` and add a check when instantiate a `PyPTS` with `device != None` where `_com_port_exists` can be called. This could then be an internal method for the class, e.g.

```python
if device is not None and not self._com_port_exists(device):
    raise ValueError(...)
```

or something similar to that

mkasenberg:
The legacy dual mode PTS dongles are not visible as COM ports. Could you add a check like this:
```python
if port_name.startswith("COM"):
```

autoptsserver.py
koffes: Made some changes. It should only check the dongles arguments and not server.

Thalley:
> The legacy dual mode PTS dongles are not visible as COM ports. Could you add a check like this:
> 
> ```python
> if port_name.startswith("COM"):
> ```

What about the new Intel ones?

sjanc:
> What about the new Intel ones?

We don't use those so not sure, but I'd assume similar to old PTS dongles (ie BT USB transport)

autoptsserver.py
Thalley: > However, I'd want the check here to ensure it is caught before the servers are started.

One argument for moving it to the place that actually uses them, is that this check will only be applied if `autoptsserver.py` is run as the main script, whereas moving the check would apply the check to any uses of `PyPTS`

Thalley: That's also a good argument. Maybe we can add it to both places then?

koffes: @sjanc what do you think?

Thalley (approved):

koffes:
I added `if _args.dongle is not None:` as I am unsure if your proposed change may impact other parts of the system. Now the type is unchanged.

Thalley (approved):

mkasenberg (approved):

sjanc (approved):

PR #1313, opened 2024-12-03
zephyr: boards: fix nrf53 downstream

AyturkDuzen:
Pass '-DSNIPPET=bt-ll-sw-split' for building zephyr LL


sjanc:
why is this needed? nothing in upstream zephyr seems to use it (moreover it seems to be out of sync now since CONFIG_BT_CTLR=y is now renamed...)

AyturkDuzen:
It is not required in the upstream as ZLL is already the default. However, for sdk-zephyr downstream, this is required for fixing the build. We either patch this to our fork or make the update here on main so we don't need to diverge. Since the change doesn't affect the behavior upstream, thought it best to merge here.

sjanc:
https://github.com/zephyrproject-rtos/zephyr/pull/82157#event-15506142248

what if it gets removed from upstream? west will silently ignore non-existing snipped?

jhedberg:
All that snippet does now is `CONFIG_BT=y`. The way to control which LL gets built in your downstream tree should be through devicetree. I.e. if you want to disable the SoftDevice controller and enable the ll_sw_split one, you need to disable the SDC DT node, enable the ll_sw_split one, and update your `zephyr,bt-hci` chosen property to point at the right node.

PR #1328, opened 2024-12-20
wid: MCP: Rework some handlers for more explicit actions

sjanc:
Do discovery only when asked to do it and avoid hardcoded ATT handles in WIDs.

### DIFF
--- a/autopts/wid/mcp.py
+++ b/autopts/wid/mcp.py
@@ -221,10 +221,6 @@ def hdl_wid_20100(params: WIDParams):
     btp.gap_conn()
     stack.gap.wait_for_connection(timeout=5, addr=addr)
     stack.gap.gap_wait_for_sec_lvl_change(level=2, timeout=30)
-    btp.mcp_discover(addr_type, addr)
-    ev = stack.mcp.wait_discovery_completed_ev(addr_type, addr, 20, remove=False)
-    if ev is None:
-        return False
 
     return True
 
@@ -236,7 +232,13 @@ def hdl_wid_20103(params: WIDParams):
     Test (IUT) can send Discover All Characteristics command.
     """
 
-    # Discover and subscribe at hdl_wid_20100
+    addr_type = btp.pts_addr_type_get()
+    addr = btp.pts_addr_get()
+    stack = get_stack()
+    btp.mcp_discover(addr_type, addr)
+    ev = stack.mcp.wait_discovery_completed_ev(addr_type, addr, 20, remove=False)
+    if ev is None:
+        return False
 
     return True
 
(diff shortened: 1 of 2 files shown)


autopts/wid/mcp.py
```diff
@@ -341,24 +350,29 @@ def hdl_wid_20110(params: WIDParams):
     addr_type = btp.pts_addr_type_get()
     addr = btp.pts_addr_get()
 
-    if "0x00F3" in params.description:
+    """ PTS is providing handle value instead of expected action so we map it here based on
```
Thalley, line 353: Is the description of this WID correct? Seems more related to MICP than MCP; MCP isn't related to audio input control point

autopts/wid/mcp.py
```diff
@@ -341,24 +350,29 @@ def hdl_wid_20110(params: WIDParams):
...
+    if 'MCP/CL/CGGIT/CHA/BV-07-C' == params.test_case_name:
         btp.mcp_track_position_set(100, addr_type, addr)
         ev = stack.mcp.wait_track_position_ev(addr_type, addr, 10)
-    elif "0x00F6" in params.description:
+    elif 'MCP/CL/CGGIT/CHA/BV-08-C' == params.test_case_name:
         btp.mcp_playback_speed_set(64, addr_type, addr)
         ev = stack.mcp.wait_playback_speed_ev(addr_type, addr, 10)
-    elif "0x0101" in params.description:
+    elif 'MCP/CL/CGGIT/CHA/BV-11-C' == params.test_case_name:
+        btp.mcp_current_track_obj_id_set(16777499, addr_type, addr)
```
Thalley, line 363: Where does the literal ints (e.g. `16777499`) come from?

autopts/wid/mcp.py
sjanc: I think this is copy'n'paste error in those descriptions

autopts/wid/mcp.py
sjanc: from what I see those are hardcoded value returned by PTS....  I'll see if we can track this and avoid hardcoded values

autopts/wid/mcp.py
```diff
@@ -287,6 +298,8 @@ def hdl_wid_20107(params: WIDParams):
     elif "Current Track Object" in params.description:
         btp.mcp_current_track_obj_id_read(addr_type, addr)
         ev = stack.mcp.wait_current_track_obj_id_ev(addr_type, addr, 10)
+        if ev is not None:
+            stack.mcp.object_id = ev[3]
```
Thalley, line 302: Why type of object is `ev` here? Where is it documented that the ID is in `ev[3]`?

sjanc: this is just python tuple, there are no dedicated classes for all those events

https://github.com/auto-pts/auto-pts/blob/master/autopts/pybtp/btp/mcp.py#L729

Thalley: Thanks :)

piotrnarajowski (approved):
Can't test these changes because of PTS issue (PTS disconnects right after IUT performs discovery) but everything looks good to me.

PR #1332, opened 2024-12-27
TMAP: Fix various issues

piotrnarajowski:
- Adjust TMAP preconditions - Calling BTP Commands with uninitialized svcs causes BTP ERROR in TMAP/CG/VRC/ test cases. 
- Adjust wid handlers - wid_504_settings dictionary is adjusted accordingly to newest test case naming. WID 20107 which handles reading characteristics should use profile api.
related - https://github.com/zephyrproject-rtos/zephyr/pull/83456

### DIFF
--- a/autopts/wid/tmap.py
+++ b/autopts/wid/tmap.py
@@ -370,31 +370,31 @@ def hdl_wid_503(params: WIDParams):
     if params.test_case_name.endswith('LT2'):
         return True
 
-    btp.tbs_terminate_call(0)
+    btp.tbs_terminate_call(1)
 
     return True
 
 
 wid_504_settings = {
     # test_case_name: (lt count, iut as audio source: streams + channels, sink locations (0 - don't care), iut as audio sink streams, metadata)
-    'TMAP/CG/VRC/BV-01-I': (1, 1, 1, 0, 1, 1, struct.pack('<BBH', 3, AUDIO_METADATA_STREAMING_AUDIO_CONTEXTS, 0x0200)),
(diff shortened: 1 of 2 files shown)


autopts/wid/tmap.py
```diff
@@ -370,31 +370,31 @@ def hdl_wid_503(params: WIDParams):
     if params.test_case_name.endswith('LT2'):
         return True
 
-    btp.tbs_terminate_call(0)
+    btp.tbs_terminate_call(1)
```
sjanc, line 373: why call index needs to be changed here?

piotrnarajowski: As per TBS spec, "When the service instantiates, the Call_Index shall be a number from 1 to 255. (...) The
Call_Index of 0 is reserved."
In zephyr, Call_index of 0 is defined as BT_TBS_FREE_CALL_INDEX and is reserved for outgoing calls.
It would be better not to use hardcoded index with this btp command, but for now I would leave this as is and modify in the future when it is specified in wid description e.g. "Please make a call ID (1) to the Remotely Held state."

Thalley: We should get the call index from either PTS or from the callbacks from the stack. It should be known and not hardcoded

piotrnarajowski: I think it could be retrieved from callback by reading call state. @Thalley if you have better ideas please share suggestions

Thalley: It's a bit unclear to me whether this is on the TBS server or client side, but regardless when the TBS server allocates a new call, there is (at least in Zephyr) a callback with the call ID. That could be sent to autopts via an event. As of today there is no way for the TBS server application to get the list of calls or their call indexes in Zephyr

sjanc (approved):

PR #1337, opened 2025-01-09
client: Fix logging an exception at exit

mkasenberg:
It seems that the 'code' field is specific for the SystemExit class, because while handling other exception I've got this one:
  File "/home/codecoup/autopts/client.py", line 1378, in start
    if e.code != 0:
AttributeError: 'TimeoutError' object has no attribute 'code'

### DIFF
--- a/autopts/client.py
+++ b/autopts/client.py
@@ -1373,15 +1373,22 @@ class Client:
             signal.signal(signal.SIGINT, sigint_handler)
 
             return self.main(args)
-        except BaseException as e:
-            if not isinstance(e, KeyboardInterrupt):   # Ctrl-C
-                if e.code != 0:
-                    # Exit with traceback
-                    logging.exception(e)
-
+        except KeyboardInterrupt:
+            # Handling Ctrl-C
+            raise
+        except SystemExit as e:
+            if e.code != 0:
+                logging.exception(e)
+            # Handling --help
+            raise
+        except (TimeoutError, BaseException) as e:
+            logging.exception(e)
+            # We have to propagate an exception from the simple client layer
+            # up to the bot layer, so it could handle its own cleanup.
+            raise
+        finally:
             set_global_end()
             self.cleanup()
-            raise
 
     def main(self, _args=None):
         """Main.


autopts/client.py
```diff
@@ -1375,7 +1375,7 @@ def sigint_handler(sig, frame):
             return self.main(args)
         except BaseException as e:
             if not isinstance(e, KeyboardInterrupt):   # Ctrl-C
-                if e.code != 0:
+                if getattr(e, "code", 1) != 0:
                     # Exit with traceback
                     logging.exception(e)
```
Thalley, line 1380: We should consider doing this in a more pythonic way IMO. 

So it should be of the form of
```
try: 
    # do stuff
except KeyboardInterrupt:
    # no-op
    pass
except SystemExit as e:
    if e.code != 0:
        # Exit with traceback
        logging.exception(e)
except TimeoutError as e:
    logging.exception(e)
finally:
    set_global_end()
    self.cleanup()
    raise
```

Generally catching `BaseException` in Python is frowned upon for the reasons here. It catches everything, and makes it harder to determine what errors may happen here and how they are handled.

autopts/client.py
```diff
@@ -1373,12 +1373,15 @@ def sigint_handler(sig, frame):
...
+        except SystemExit as e:
+            if e.code != 0:
+                logging.exception(e)
         except BaseException as e:
-            if not isinstance(e, KeyboardInterrupt):   # Ctrl-C
-                if e.code != 0:
-                    # Exit with traceback
-                    logging.exception(e)
-
+            logging.exception(e)
```
Thalley, line 1383: I think handling `BaseException` as a last effort is OK (but not ideal). We should try to cover all known exceptions specifically, so that the code becomes more obvious. 

I would thus still suggest to add
```
except TimeoutError as e:
    logging.exception(e)
```
before `except BaseException`

or do
```
except (TimeoutError, BaseException) as e:
    logging.exception(e)
```

but the best approach would be to not catch `BaseException` (or `Exception`) at all IMO, which is also the recommendation by PEP8. 

Since it's advised against in PEP8, most Python linters will also complain about a `except BaseException` or `except BaseException`

@sjanc should we consider following PEP8 and/or adding a style guide for autopts, and use a linter in CI?

sjanc: yeap, I think enforcing PEP8 is good thing (maybe with some exceptions, line allow lines longer than 79. eg 120)

there is some cleanup to be done before this can be enabled on default though :)

Thalley: Indeed. Refactoring the entire codebase will take a while :D 

https://github.com/auto-pts/auto-pts/issues/1338

autopts/client.py
```diff
@@ -1373,15 +1373,22 @@ def sigint_handler(sig, frame):
...
+        except SystemExit as e:
+            if e.code != 0:
+                logging.exception(e)
+            # Handling --help
+            raise
+        except (TimeoutError, BaseException) as e:
+            logging.exception(e)
+            # We have to propagate an exception from the simple client layer
+            # up to the bot layer, so it could handle its own cleanup.
+            raise
```
Thalley, line 1388: I think you can put the `raise` in the `finally` clause, right?

mkasenberg: It seems that a lonely `raise` doesn't work in `finally`:
`RuntimeError: No active exception to reraise`

Thalley (approved):

sjanc (approved):

piotrnarajowski (approved):

PR #1345, opened 2025-01-14
bot: report: add test group summary to readme.md

piotrnarajowski:
Previously implemented profile_summary function is now used to write to readme.md file in Github logging repo and AutoPTS
test session result email. Table width is now adjusted to be displayed properly on various screens

### DIFF
--- a/autopts/bot/common_features/mail.py
+++ b/autopts/bot/common_features/mail.py
@@ -81,7 +81,7 @@ class TestGroup:
             self.pass_rate = (self.passed / float(self.total)) * 100
 
 
-def profile_summary_html(tc_results):
+def profile_summary(tc_results):
     """Creates HTML formatted message with summarized profile results"""
 
     """Dictionary containing profile name as key, test group object as value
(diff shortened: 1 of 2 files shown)


autopts/bot/common_features/mail.py
```diff
@@ -107,39 +107,49 @@ def profile_summary_html(tc_results):
...
-                <tbody>
-                    {table_rows}
-                </tbody>
-            </table>
-        </div>
-        """
-    return suite_summary
+        return suite_summary
+    else:
+        # ascii
```
Thalley, line 144: Why ascii instead of markdown?

Thalley, line 144: Why ascii instead of just using the HTML table? 
AFAIK you can embed HTML in markdown for the README.md

Alternatively, the Github Markdown flavor also supports tables IIRC.

piotrnarajowski: @sjanc suggested using ascii formated, but IMO html table looks neat in the email so I decided to leave as is for email and use ascii just for readme

piotrnarajowski: @Thalley HTML table is now used for .md and email

Thalley (approved):

mkasenberg (approved):

PR #1346, opened 2025-01-15
Fix flash failure on 54H20

yuxCai:
In case of a west flash failure due to memory access restriction, a device recover for all cores clears
 the memory allocations and thus the flash should work


theob-pro:
> I'll let Nordic folks to comment on this, but I'd expect that west should be able to recover nrf chip if needed ?

I think it's not actually 😅

But @yuxCai could we use `west flash --recover` instead of `nrfutil`? I guess it does the same, it's more to do it the "Zephyr way"

yuxCai:
Updated the PR to use --recover only when necessary

Thalley (approved):

theob-pro (approved):

PR #1350, opened 2025-01-17
Adding debugging info

yuxCai:
Adding debugging info

### DIFF
--- a/autoptsserver.py
+++ b/autoptsserver.py
@@ -68,6 +68,7 @@ def _com_port_exists(port_name):
     """Check if the COM port exists."""
     ports = serial.tools.list_ports.comports()
     for port in ports:
+        print(f"Port_device: {port.device}, port_name: {port_name}")
         if port.device == port_name:
             return True
     return False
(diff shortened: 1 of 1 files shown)


autoptsserver.py
```diff
@@ -636,10 +637,12 @@ def shutdown_pts_bpv(self):
 
     if _args.dongle is not None:
         for dongle in _args.dongle:
+            print(f"dongle: {dongle}")
             # If dongle does not start with "COM" then it is considered valid port
             # for legacy dual-mode dongle compatibility.
             if dongle.startswith("COM"):
                 if not _com_port_exists(dongle):
+                    print(f"Port {dongle} not found")
                     sys.exit(f"Port {dongle} not found")
```
Thalley, line 646: Since we exit with the same information, is that useful?

autoptsserver.py
```diff
@@ -68,6 +68,7 @@ def _com_port_exists(port_name):
     """Check if the COM port exists."""
     ports = serial.tools.list_ports.comports()
     for port in ports:
+        print(f"Port_device: {port.device}, port_name: {port_name}")
```
Thalley, line 71: Since `_com_port_exists` is called once per port per dongle, then we end up printing this n * m times, which seems very verbose

yuxCai:
Please ignore this PR. For testing purpose only

PR #1351, opened 2025-01-24
wid: MCP: Fix storing object_id in WID 20107

sjanc:
This ID is later used in WID 20121.


piotrnarajowski (approved):
Maybe in the future we should add to stack.mcp variables that refer to specific object ID but for now this works good I guess

PR #1356, opened 2025-01-24
zephyr: wid: btp: Add files for PBP source testing

Frodevan:
This adds code for testing the PBP feature for the Zephyr project on broadcast source. This is done in conjunction with a PR in the Zephyr repository.

See https://github.com/zephyrproject-rtos/zephyr/pull/82654

### DIFF
--- a/autopts/ptsprojects/stack/layers/pbp.py
+++ b/autopts/ptsprojects/stack/layers/pbp.py
@@ -19,6 +19,8 @@ from autopts.pybtp import defs
 
 class PBP:
     def __init__(self):
+        self.program_info = None
+        self.broadcast_name = None
         self.event_queues = {
             defs.BTP_PBP_EV_PUBLIC_BROADCAST_ANNOUNCEMENT_FOUND: [],
         }
@@ -31,3 +33,17 @@ class PBP:
             self.event_queues[defs.BTP_PBP_EV_PUBLIC_BROADCAST_ANNOUNCEMENT_FOUND],
             lambda ev: (addr_type, addr, broadcast_name) == (ev['addr_type'], ev['addr'], ev['broadcast_name']),
             timeout, remove)
+
+    def set_program_info(self, program_info):
+        if not isinstance(program_info, str):
+            raise TypeError('Tried to set improper program info')
+        if len(program_info) == 0:
+            raise ValueError('Tried to set empty program info')
+        self.program_info = program_info
+
+    def set_broadcast_name(self, broadcast_name):
+        if not isinstance(broadcast_name, str):
+            raise TypeError('Tried to set improper broadcast name')
+        if len(broadcast_name) < 4 or len(broadcast_name) > 32:
+            raise ValueError(f'Tried to set broadcast name with invalid length: \'{broadcast_name}\' ({len(broadcast_name)})')
+        self.broadcast_name = broadcast_name
--- a/autopts/ptsprojects/zephyr/pbp.py
+++ b/autopts/ptsprojects/zephyr/pbp.py
@@ -1,0 +1,94 @@
+#
+# auto-pts - The Bluetooth PTS Automation Framework
+#
+# Copyright (c) 2025, Nordic Semiconductor ASA.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms and conditions of the GNU General Public License,
+# version 2, as published by the Free Software Foundation.
+#
+# This program is distributed in the hope it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+
+import struct
+
+from autopts.pybtp import btp
+from autopts.client import get_unique_name
+from autopts.ptsprojects.stack import get_stack
+from autopts.ptsprojects.testcase import TestFunc
+from autopts.ptsprojects.zephyr.pbp_wid import pbp_wid_hdl
+from autopts.ptsprojects.zephyr.ztestcase import ZTestCase
+from autopts.pybtp.types import Addr
+from autopts.utils import ResultWithFlag
+
+PROGRAM_INFO = '00112233445566778899AABBCCDDEEFF'
+BROADCAST_CODE = '0102680553F1415AA265BBAFC6EA03B8'
+BROADCAST_NAME = 'Broadcaster'
+
+def set_pixits(ptses):
+    pts = ptses[0]
+
+    pts.set_pixit("PBP", "TSPX_time_guard", "180000")
+    pts.set_pixit("PBP", "TSPX_use_implicit_send", "TRUE")
+    pts.set_pixit("PBP", "TSPX_broadcast_code", BROADCAST_CODE)
+    pts.set_pixit("PBP", "TSPX_Program_Info", PROGRAM_INFO)
+    pts.set_pixit("PBP", "TSPX_Broadcast_Name", BROADCAST_NAME)
+
+def test_cases(ptses):
+    """
+    Returns a list of PBP test cases
+    ptses -- list of PyPTS instances
+    """
+
+    pts = ptses[0]
+
+    pts_bd_addr = pts.q_bd_addr
+    iut_device_name = get_unique_name(pts)
+    stack = get_stack()
+
+    iut_addr = ResultWithFlag()
+
+    def set_addr(addr):
+        iut_addr.set(addr)
+
+    pre_conditions = [
+        TestFunc(btp.core_reg_svc_gap),
+        TestFunc(stack.gap_init, iut_device_name),
+        TestFunc(btp.gap_read_ctrl_info),
+        TestFunc(btp.core_reg_svc_gatt),
+        TestFunc(btp.set_pts_addr, pts_bd_addr, Addr.le_public),
+        TestFunc(stack.gatt_init),
+        TestFunc(btp.gap_set_conn),
+        TestFunc(btp.core_reg_svc_ascs),
+        TestFunc(btp.core_reg_svc_bap),
+        TestFunc(stack.ascs_init),
+        TestFunc(stack.bap_init),
+        TestFunc(stack.cap_init),
+        TestFunc(btp.core_reg_svc_cap),
+        TestFunc(btp.core_reg_svc_cas),
+        TestFunc(btp.core_reg_svc_pbp),
+        TestFunc(stack.pbp_init),
+        TestFunc(btp.gap_set_extended_advertising_on),
+        TestFunc(lambda: stack.bap.set_broadcast_code(BROADCAST_CODE)),
+        TestFunc(lambda: stack.pbp.set_program_info(PROGRAM_INFO)),
+        TestFunc(lambda: stack.pbp.set_broadcast_name(BROADCAST_NAME)),
+        TestFunc(lambda: set_addr(stack.gap.iut_addr_get_str())),
+
+        TestFunc(lambda: pts.update_pixit_param("PBP", "TSPX_bd_addr_iut",
+                                                stack.gap.iut_addr_get_str()))
+    ]
+
+    test_case_name_list = pts.get_test_case_list('PBP')
+    tc_list = []
+
+    # Use the same preconditions and MMI/WID handler for all test cases of the profile
+    for tc_name in test_case_name_list:
+        instance = ZTestCase('PBP', tc_name, cmds=pre_conditions,
+                             generic_wid_hdl=pbp_wid_hdl)
+
+        tc_list.append(instance)
+
+    return tc_list
--- a/autopts/ptsprojects/zephyr/pbp_wid.py
+++ b/autopts/ptsprojects/zephyr/pbp_wid.py
@@ -1,0 +1,26 @@
+#
+# auto-pts - The Bluetooth PTS Automation Framework
+#
+# Copyright (c) 2025, Nordic Semiconductor ASA.
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms and conditions of the GNU General Public License,
+# version 2, as published by the Free Software Foundation.
+#
+# This program is distributed in the hope it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+
+import logging
+
+from autopts.pybtp.types import WIDParams
+from autopts.wid import generic_wid_hdl
+
+log = logging.debug
+
+
+def pbp_wid_hdl(wid, description, test_case_name):
+    log(f'{pbp_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
+    return generic_wid_hdl(wid, description, test_case_name, [__name__, 'autopts.wid.pbp'])
--- a/autopts/wid/pbp.py
+++ b/autopts/wid/pbp.py
@@ -198,11 +198,15 @@ def hdl_wid_552(params: WIDParams):
     features, bis_count, qos_set_name = wid_552_settings[params.test_case_name]
 
     # default PTS program info, see TSPX_Program_Info
-    program_info_string = "00112233445566778899AABBCCDDEEFF"
-    program_info = bytes.fromhex(program_info_string)
+    stack = get_stack()
+    program_info = bytes.fromhex(stack.pbp.program_info)
     program_info_len = len(program_info)
     metadata = struct.pack('<BB', program_info_len + 1, AUDIO_METADATA_PROGRAM_INFO) + program_info
+
     btp.pbp.pbp_set_public_broadcast_announcement(features, metadata)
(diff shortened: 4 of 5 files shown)


autopts/wid/pbp.py
```diff
@@ -202,7 +202,11 @@ def hdl_wid_552(params: WIDParams):
     program_info = bytes.fromhex(program_info_string)
     program_info_len = len(program_info)
     metadata = struct.pack('<BB', program_info_len + 1, AUDIO_METADATA_PROGRAM_INFO) + program_info
+
     btp.pbp.pbp_set_public_broadcast_announcement(features, metadata)
+    btp.pbp.pbp_set_broadcast_name("Broadcaster")
```
Thalley, line 207: This should use the IXIT value, instead of a hardcoded string

autopts/ptsprojects/zephyr/pbp_wid.py
```diff
@@ -0,0 +1,27 @@
...
+
+import logging
+
+from autopts.pybtp.types import WIDParams
+from autopts.wid import generic_wid_hdl
+
+log = logging.debug
+
+
+def pbp_wid_hdl(wid, description, test_case_name):
```
Thalley, line 25:
```suggestion
log = logging.debug

def pbp_wid_hdl(wid, description, test_case_name):
```

autopts/ptsprojects/zephyr/pbp_wid.py
```diff
@@ -0,0 +1,27 @@
...
+# version 2, as published by the Free Software Foundation.
+#
+# This program is distributed in the hope it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
+# more details.
+#
+
+
+import logging
```
Thalley, line 17:
```suggestion

import logging
```

autopts/ptsprojects/zephyr/pbp_wid.py
```diff
@@ -0,0 +1,27 @@
...
+
+from autopts.pybtp.types import WIDParams
+from autopts.wid import generic_wid_hdl
+
+log = logging.debug
+
+
+def pbp_wid_hdl(wid, description, test_case_name):
+    log(f'{pbp_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
```
Thalley, line 26:
```suggestion
def pbp_wid_hdl(wid, description, test_case_name):
    logging.debug(f'{pbp_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
```

autopts/ptsprojects/zephyr/pbp.py
```diff
@@ -0,0 +1,87 @@
...
+        TestFunc(stack.ascs_init),
+        TestFunc(stack.bap_init),
+        TestFunc(stack.cap_init),
+        TestFunc(btp.core_reg_svc_cap),
+        TestFunc(btp.core_reg_svc_cas),
+        TestFunc(btp.core_reg_svc_pbp),
+        TestFunc(stack.pbp_init),
+        TestFunc(btp.gap_set_extended_advertising_on),
+        TestFunc(lambda: stack.bap.set_broadcast_code(broadcast_code)),
+        TestFunc(lambda: set_addr(stack.gap.iut_addr_get_str())),
```
Thalley, line 71: Might be nice to order these by name, or at least group them by profile

autopts/ptsprojects/zephyr/pbp.py
```diff
@@ -0,0 +1,87 @@
...
+from autopts.pybtp import btp
+from autopts.client import get_unique_name
+from autopts.ptsprojects.stack import get_stack
+from autopts.ptsprojects.testcase import TestFunc
+from autopts.ptsprojects.zephyr.pbp_wid import pbp_wid_hdl
+from autopts.ptsprojects.zephyr.ztestcase import ZTestCase
+from autopts.pybtp.types import Addr
+from autopts.utils import ResultWithFlag
+
+broadcast_code = '0102680553F1415AA265BBAFC6EA03B8'
```
Thalley, line 26: Since this is done almost everywhere else, it is fine for now, but generally we should avoid global variables, and if we do want to use global variables, it is common to use UPPER_CASE for global variables so they are no confused with local variables

Thalley (changes requested):

autopts/ptsprojects/zephyr/pbp.py
Frodevan: Many of these were copy-pasted from the cap.py version, where they appear in this order. I am not sure if there's a reason for that, or if it's not really that important.

autopts/ptsprojects/zephyr/pbp_wid.py
Frodevan: PEP8 rules do suggest 2 blank lines before function definitions

autopts/ptsprojects/zephyr/pbp_wid.py
Frodevan: This is part of code which was autogenerated. I'd rather keep it as the other modules.

autopts/wid/pbp.py
Frodevan: Done

autopts/ptsprojects/zephyr/pbp_wid.py
Frodevan: Done

autopts/ptsprojects/zephyr/pbp.py
Frodevan: Changed to uppercase

autopts/ptsprojects/zephyr/pbp_wid.py
Thalley: Fair point :)

Thalley (changes requested):

autopts/ptsprojects/stack/layers/pbp.py
```diff
@@ -19,6 +19,8 @@
 
 class PBP:
     def __init__(self):
+        self.program_info = ''
+        self.broadcast_name = ''
```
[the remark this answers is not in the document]
Frodevan: If the code which tries to access these later does not test for None, it can generate an exception if it treats the values as strings.

autopts/ptsprojects/zephyr/pbp.py
```diff
@@ -0,0 +1,94 @@
...
+    pts_bd_addr = pts.q_bd_addr
+    iut_device_name = get_unique_name(pts)
+    stack = get_stack()
+
+    iut_addr = ResultWithFlag()
+
+    def set_addr(addr):
+        iut_addr.set(addr)
+
+    pre_conditions = [
```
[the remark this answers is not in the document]
Frodevan: PACS is currently not being used or addressed in any of the PBP test code, so I don't see why we need to ready the bttester to accept BTP commands for PACS.

autopts/ptsprojects/stack/layers/pbp.py
```diff
@@ -31,3 +33,9 @@ def wait_public_broadcast_event_found_ev(self, addr_type, addr, broadcast_name,
             self.event_queues[defs.BTP_PBP_EV_PUBLIC_BROADCAST_ANNOUNCEMENT_FOUND],
             lambda ev: (addr_type, addr, broadcast_name) == (ev['addr_type'], ev['addr'], ev['broadcast_name']),
             timeout, remove)
+
+    def set_program_info(self, program_info):
+        self.program_info = program_info
+
+    def set_broadcast_name(self, broadcast_name):
+        self.broadcast_name = broadcast_name
```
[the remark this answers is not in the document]
Frodevan: Done

autopts/ptsprojects/stack/layers/pbp.py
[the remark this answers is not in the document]
Thalley: Right, but would it be nice to have to ensure that they are set before use? 

I did see somewhere, where e.g. the broadcast code was checked for None, but it was never none as it was initialized to the empty string :|

@sjanc what do you think w.r.t. None vs the empty string?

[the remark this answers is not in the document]
sjanc: using None results in exception so it is easy to catch such unintended use, I'd call this a feature :)    autopts in such case catches exception and fails test (with 'FATAL ERROR' result) and move to next one, so no harm in using None

I'd personally use None for this reason

autopts/ptsprojects/zephyr/pbp.py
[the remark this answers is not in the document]
Frodevan: I think that's the best solution for now.

autopts/ptsprojects/stack/layers/pbp.py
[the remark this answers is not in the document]
Frodevan: Ok, done

autopts/ptsprojects/stack/layers/pbp.py
```diff
@@ -31,3 +33,11 @@ def wait_public_broadcast_event_found_ev(self, addr_type, addr, broadcast_name,
...
             lambda ev: (addr_type, addr, broadcast_name) == (ev['addr_type'], ev['addr'], ev['broadcast_name']),
             timeout, remove)
+
+    def set_program_info(self, program_info):
+        if isinstance(program_info, str) and len(program_info > 0):
+            self.program_info = program_info
+
+    def set_broadcast_name(self, broadcast_name):
+        if isinstance(broadcast_name, str) and len(broadcast_name > 0):
+            self.broadcast_name = broadcast_name
```
Thalley, line 43: Should they raise an exception instead of failing silently?

autopts/ptsprojects/zephyr/pbp.py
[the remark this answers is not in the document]
Thalley: @Frodevan for good measure, please also update the PR title

sjanc (commented):
looks OK to me but lets wait for comment from @mringwal (as change in hdl_wid_552 will likely require update on his side)

autopts/ptsprojects/zephyr/pbp.py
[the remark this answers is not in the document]
Frodevan: ok, done

autopts/ptsprojects/stack/layers/pbp.py
mringwal: Please fail loudly, it makes debugging / adding support for other stacks / ptsprojects much easier.

mringwal:
Thanks for the consideration. Please move on.

We don't have automated PTS testing in place, so I'll look at these later. It would be great to pepper the code with more assertions, e.g. if the WID tries to use a value like program info, but it hasn't been set.

autopts/ptsprojects/stack/layers/pbp.py
```diff
@@ -31,3 +33,15 @@ def wait_public_broadcast_event_found_ev(self, addr_type, addr, broadcast_name,
...
+        if isinstance(program_info, str) and len(program_info > 0):
+            self.program_info = program_info
+        else:
+            raise Exception('Tried to set improper program info')
+
+    def set_broadcast_name(self, broadcast_name):
+        if isinstance(broadcast_name, str) and len(broadcast_name > 0):
+            self.broadcast_name = broadcast_name
+        else:
+            raise Exception('Tried to set improper broadcast name')
```
alexsven, line 47: Missing blank line at end of file

autopts/ptsprojects/stack/layers/pbp.py
```diff
@@ -31,3 +33,15 @@ def wait_public_broadcast_event_found_ev(self, addr_type, addr, broadcast_name,
...
             timeout, remove)
+
+    def set_program_info(self, program_info):
+        if isinstance(program_info, str) and len(program_info > 0):
+            self.program_info = program_info
+        else:
+            raise Exception('Tried to set improper program info')
+
+    def set_broadcast_name(self, broadcast_name):
+        if isinstance(broadcast_name, str) and len(broadcast_name > 0):
```
alexsven, line 44: According to the spec, the broadcast_name shall be minimum 4, maximum 32 characters

autopts/wid/pbp.py
```diff
@@ -235,7 +239,6 @@ def hdl_wid_552(params: WIDParams):
     broadcast_code = None
```
alexsven, line 239: Can't comment on the lines above, but the broadcast_id should also be part of the TSPX and be fetched the same way, instead of being hard coded here

alexsven (changes requested):

autopts/wid/pbp.py
Thalley: The comment, `# Zephyr stack generates Broadcast_ID itself`, is also no longer true :)

autopts/ptsprojects/stack/layers/pbp.py
Frodevan: Fixed

autopts/ptsprojects/stack/layers/pbp.py
Frodevan: Fixed

autopts/ptsprojects/stack/layers/pbp.py
Frodevan: Done

autopts/wid/pbp.py
Frodevan: I can remove the comment, but this should in the case be done in a whole bunch of the wid modules.

autopts/ptsprojects/stack/layers/pbp.py
```diff
@@ -31,3 +33,15 @@ def wait_public_broadcast_event_found_ev(self, addr_type, addr, broadcast_name,
...
+        if isinstance(program_info, str) and len(program_info > 0):
+            self.program_info = program_info
+        else:
+            raise Exception('Tried to set improper program info')
+
+    def set_broadcast_name(self, broadcast_name):
+        if isinstance(broadcast_name, str) and len(broadcast_name >= 4) and len(broadcast_name <= 32):
+            self.broadcast_name = broadcast_name
+        else:
+            raise Exception('Tried to set improper broadcast name')
```
Thalley, line 47: Just to make this a bit more explicit: 

If the `isinstance(broadcast_name, str)` check fails, we should raise a TypeError, and if the length check fails, we should raise a ValueError. 

We should not rely on the basic Exception

Thalley (changes requested):

autopts/wid/pbp.py
Frodevan: Broadcast ID is otherwise only an IXIT value for BAP and TMAP tests, so I think it should be ok to define it in the wid call as done here (similar to the CAP wid file)

autopts/ptsprojects/stack/layers/pbp.py
Frodevan: Fixed

alexsven (approved):

autopts/wid/pbp.py
Frodevan: Done

autopts/ptsprojects/stack/layers/pbp.py
Frodevan: Done

autopts/wid/pbp.py
Thalley: > Broadcast ID is otherwise only an IXIT value for BAP and TMAP tests

PTS being PTS :upside_down_face:

autopts/ptsprojects/stack/layers/pbp.py
```diff
@@ -31,3 +33,17 @@ def wait_public_broadcast_event_found_ev(self, addr_type, addr, broadcast_name,
...
+            raise TypeError('Tried to set improper program info')
+        if len(program_info) == 0:
+            raise ValueError('Tried to set empty program info')
+        self.program_info = program_info
+
+    def set_broadcast_name(self, broadcast_name):
+        if not isinstance(broadcast_name, str):
+            raise TypeError('Tried to set improper broadcast name')
+        if len(broadcast_name) < 4 or len(broadcast_name) > 32:
+            raise ValueError('Tried to set broadcast name with invalid length')
```
Thalley, line 48:
```suggestion
            raise ValueError(f'Tried to set broadcast name with invalid length: {broadcast_name} ({len(broadcast_name)})')
```

autopts/wid/pbp.py
```diff
@@ -229,13 +233,11 @@ def hdl_wid_552(params: WIDParams):
     btp.cap_broadcast_source_setup_subgroup(source_id, subgroup_id, coding_format, vid, cid,
                                             codec_ltvs_bytes, metadata)
 
-    # Zephyr stack generates Broadcast_ID itself
```
Thalley, line 232: If there is not IXIT value for broadcast ID for PBP (...) then I suggest that we replace this comment with stating that, and that any broadcast_id will work.

autopts/ptsprojects/stack/layers/pbp.py
Frodevan: Added

Thalley (approved):

sjanc (approved):

PR #1357, opened 2025-01-27
rtt: Fix memory leak at JLink init

mkasenberg:
Pylink loads a new cache of J-Link DLL at its __init__, but its __del__ does not unload the cache. If RTT logging was used, there was a huge memory leak, about 20MB per test case. Let's reuse the cached lib to prevent the leak.

### DIFF
--- a/autopts/utils.py
+++ b/autopts/utils.py
@@ -457,6 +457,14 @@ def exit_if_admin():
         sys.exit("Administrator rights are not required to run this script!")
 
 
+def log_memory_usage():
+    process = psutil.Process(os.getpid())
+    mem_info = process.memory_info()
+
+    mem_usage = mem_info.rss / (1024 ** 2)
+    logging.debug(f"Memory usage: {mem_usage:.2f} MB")
+
+
 def main():
     """Main."""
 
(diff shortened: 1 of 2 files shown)


autopts/utils.py
```diff
@@ -457,6 +457,14 @@ def exit_if_admin():
         sys.exit("Administrator rights are not required to run this script!")
 
 
+def log_memory_usage():
+    process = psutil.Process(os.getpid())
+    mem_info = process.memory_info()
+
+    mem_usage = mem_info.rss / (1024 ** 2)
+    logging.debug(f"Memory usage: {mem_usage:.2f} MB")
```
Thalley, line 465: Function defined but never called?

mkasenberg: Yes, you can use this if you need to debug a memory leak in one of the autopts scripts.

Thalley: Right, so you have to make source code modifications in order to use this? 

Should this be possible to enable/log automatically via a CLI option?

mkasenberg: > Right, so you have to make source code modifications in order to use this?

Yes.

> Should this be possible to enable/log automatically via a CLI option?

Might be useful.

PR #1358, opened 2025-01-27
bot: report: adjust profile_summary

piotrnarajowski:
Profile_summary function is split to two functions that can return ascii formatted table or html formatted. Html version printed in readme.md was incorrectly formatted by github.

### DIFF
--- a/autopts/bot/common.py
+++ b/autopts/bot/common.py
@@ -574,7 +574,7 @@ class BotClient(Client):
         """Creates README.md for Github logging repo
         """
         readme_file = readme_md_path
-        profile_summary = mail.profile_summary(report_data['tc_results'])
+        profile_summary = report.ascii_profile_summary(report_data['tc_results'])
 
         Path(os.path.dirname(readme_file)).mkdir(parents=True, exist_ok=True)
 
@@ -586,6 +586,8 @@ class BotClient(Client):
     End time: {report_data["end_time_stamp"]}
 
     PTS version: {report_data["pts_ver"]}
+    
+    Test Group/Profile Summary:
     
     {profile_summary}
 
@@ -707,7 +709,7 @@ class BotClient(Client):
         mail_ctx = {'project_name': report_data['project_name'],
                     'repos_info': report_data['repo_status'],
                     'summary': [mail.status_dict2summary_html(report_data['status_count'])],
-                    'profile_summary': mail.profile_summary(report_data['tc_results']),
+                    'profile_summary': mail.html_profile_summary(report_data['tc_results']),
                     'log_url': [],
                     'board': self.bot_config['auto_pts']['board'],
                     'platform': report_data['platform'],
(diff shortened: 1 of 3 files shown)


autopts/bot/common_features/mail.py
```diff
@@ -106,38 +106,49 @@ def profile_summary(tc_results):
...
-    for suite, stats in test_groups.items():
-        table_rows += f"""
-            <tr>
-                <td>{suite}</td>
-                <td>{stats.total}</td>
-                <td>{stats.passed}</td>
-                <td>{stats.failed}</td>
-                <td>{stats.pass_rate:.2f} %</td>
-            </tr>
+    if target == 'mail':
```
Thalley, line 109: Suggest to either define the targets as an string enum or something else, instead of relying on comparing literal values

autopts/bot/common_features/mail.py
```diff
@@ -81,7 +81,7 @@ def get_pass_rate(self):
             self.pass_rate = (self.passed / float(self.total)) * 100
 
 
-def profile_summary(tc_results):
+def profile_summary(tc_results, target):
     """Creates HTML formatted message with summarized profile results"""
```
Thalley, line 85: The docstring needs to be updated, as this now may create an ASCII table that I presume works with the GFM https://github.github.com/gfm/#tables-extension-? 

The arguments should also be documented

Thalley: Alternatively we can create a another function specific for this purpose. It doesn't make sense to use such a small part of a function from `mail.py` to generate an ASCII table specifically not used for mail

piotrnarajowski: Decided to create another function as you suggested. Docstrings are updated

autopts/bot/common.py
```diff
@@ -574,7 +574,7 @@ def make_readme_md(self, readme_md_path, report_data):
         """Creates README.md for Github logging repo
         """
         readme_file = readme_md_path
-        profile_summary = mail.profile_summary(report_data['tc_results'])
+        profile_summary = mail.ascii_profile_summary(report_data['tc_results'])
```
Thalley, line 577: Consider moving `ascii_profile_summary` out from the `mail.py` file, as it's not related to emails. 

It is OK to keep for now, as it has a dependency for the (also not-email specific `get_tc_res_data` function), but ideally none of those should be in a file called `mail.py`.

Thalley (approved):

autopts/bot/common.py
piotrnarajowski: Yeah right. `get_tc_res_data` is now in `common.py`, `ascii_profile_summary` has been moved to `report.py`

sjanc (approved):

Thalley:
@piotrnarajowski looks like this broke the table: 
image

piotrnarajowski:
@Thalley Gonna fix this in few hours

Issue #1251, opened 2024-09-11
Start using 'Read supported commands'


sjanc:
We should have generic framework that on service registration will also read supported commands, store them and provide simple API for querying it.

PR #1366, opened 2025-02-05
Add test BAP/BSRC/SCC/BV-38-C

alexsven:
- Add test BAP/BSRC/SCC/BV-38-C


autopts/pybtp/btp/bap.py
```diff
@@ -128,14 +128,16 @@ def bap_send(ase_id, data_ba, bd_addr_type=None, bd_addr=None):
 
 
 def bap_broadcast_source_setup(
-        streams_per_subgroup, subgroups, coding_format, vid, cid,
+        source_id, streams_per_subgroup, subgroups, coding_format, vid, cid,
         codec_ltvs, sdu_interval, framing, max_sdu, retransmission_number,
-        max_transport_latency, presentation_delay):
+        max_transport_latency, presentation_delay, broadcast_id):
```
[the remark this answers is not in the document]
alexsven: Because we don't necessarily want to keep track of which broadcast_id is what source. Source_id in this case would be wid_counter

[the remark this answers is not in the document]
alexsven: I did change all places that uses this function though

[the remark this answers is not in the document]
Thalley: You could also just use `broadcast_id=0` and `broadcast_id=1` and achieve the same :)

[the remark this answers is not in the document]
alexsven: Good point, I can make them be the same

autopts/wid/bap.py
```diff
@@ -276,6 +277,9 @@ def hdl_wid_114(params: WIDParams):
         'BAP/BSRC/STR/BV-33-C': '48_6_1',
     }
 
+    broadcast_ids = [0x123456, 0x234567]
```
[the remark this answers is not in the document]
alexsven: Isn't it the other way around? That we set the ixit-values through pts.set_pixit? I couldn't find a function for reading out values

autopts/pybtp/btp/bap.py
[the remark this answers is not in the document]
Thalley: > Because we don't necessarily want to keep track of which broadcast_id is what source. Source_id in this case would be wid_counter

Was this a comment for the autoPTS client or the BT tester? 

Either way, keeping track of the broadcast IDs aren't really much of a hassle in any case if you ask me :)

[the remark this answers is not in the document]
alexsven: Both I guess. Source_id is more useful, so I'll stick with that one. Examples are when starting and stopping advertising, stopping streams etc

autopts/wid/bap.py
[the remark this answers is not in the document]
Thalley: You did set them in the workspace here: https://github.com/auto-pts/auto-pts/pull/1366/files#diff-6494b026983b850773e7a8dfa1809288ca59bd9f58773c4a7681e7e62a6f0414R5460-R5468

But yeah, we probably should use `set_pixit` to set the values. I had expected that we could also read the values from the set PIXIT values (like some sort of `get_pixit`, but I don't see any way of doing that. @sjanc What's the best way of setting a IXIT value and then also using it in the client?

autopts/pybtp/btp/bap.py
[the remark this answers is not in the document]
alexsven: Nvm, I see that broadcast_id is already used everywhere, so will set ids to 0 and 1 and use that instead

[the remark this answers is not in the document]
Thalley: `broadcast_id` should really be the ID used to identify a broadcast (as the name implies) :D 

It is an ID, and not an index, so we should ideally not use it as an index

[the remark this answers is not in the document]
Thalley: > I did change all places that uses this function though

It is a protocol change, so all IUT devices will need be updated as well, which is the tricky part. It's not just in the autoPTS client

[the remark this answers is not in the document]
alexsven: Done

autopts/pybtp/defs.py
```diff
@@ -639,6 +639,8 @@ def BIT(bit):
 BTP_ASCS_EV_CHARACTERISTIC_SUBSCRIBED = 0x81
 BTP_ASCS_EV_ASE_STATE_CHANGED = 0x82
 
+BTP_BAP_BROADCAST_SOURCE_ID_BYTES = 1
+BTP_BAP_BROADCAST_ID_BYTES = 3
```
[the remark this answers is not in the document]
alexsven: Removed all together

autopts/ptsprojects/mynewt/bap_wid.py
```diff
@@ -54,17 +54,17 @@ def hdl_wid_380(_: WIDParams):
                                              audio_locations, octets_per_frame,
                                              frames_per_sdu)
 
+    broadcast_id = 0x123456
```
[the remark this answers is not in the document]
alexsven: Done

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -26,7 +26,9 @@ def __init__(self):
         self.peers = {}
         self.ase_configs = []
         self.broadcast_id = 0x1000000  # Invalid Broadcast ID
+        self.broadcast_id_2 = 0x1000000  # Invalid Broadcast ID 2
```
Thalley, line 29: Hmm, why we are initializing them to an invalid value in the first place? 
We are never checking this value, so perhaps we should initialize them to `None` so that we break loudly if they are not set before use?

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -26,7 +26,9 @@ def __init__(self):
...
         self.broadcast_code = ''
+        self.wid_counter = 0
```
Thalley, line 31: I would rename this to 
```suggestion
        self.hdl_wid_114 = 0
```

It is specifically used for this WID and if multiple WID handles uses this with the generic name, then we may end up with an unexpected count

autopts/ptsprojects/zephyr/bap.py
```diff
@@ -25,6 +25,8 @@
 from autopts.utils import ResultWithFlag
 
 broadcast_code = '0102680553F1415AA265BBAFC6EA03B8'
+broadcast_id = 0x123456
+broadcast_id_2 = 0x234567
```
Thalley, line 29: I know this is following the existing format, but generally in python constants are always in capital letters
```suggestion
BROADCAST_ID = 0x123456
BROADCAST_ID_2 = 0x234567
```

autopts/pybtp/btp/bap.py
```diff
@@ -130,12 +130,13 @@ def bap_send(ase_id, data_ba, bd_addr_type=None, bd_addr=None):
 def bap_broadcast_source_setup(
         streams_per_subgroup, subgroups, coding_format, vid, cid,
         codec_ltvs, sdu_interval, framing, max_sdu, retransmission_number,
-        max_transport_latency, presentation_delay):
+        max_transport_latency, presentation_delay, broadcast_id=0):
```
Thalley, line 133: Did PTS have an issue with broadcast_id == 0? If so, then I don't think we should default the braodcast_id to 0, and expect always expect the caller to provide a value here

autopts/wid/bap.py
```diff
@@ -276,6 +277,16 @@ def hdl_wid_114(params: WIDParams):
...
 
+    stack = get_stack()
+
+    if stack.bap.wid_counter == 0:
+    	broadcast_id = stack.bap.broadcast_id
+    elif stack.bap.wid_counter == 1:
+    	broadcast_id = stack.bap.broadcast_id_2
+    else:
+    	# Invalid broadcast ID
+    	broadcast_id = 0x1000000
```
Thalley, line 288: Suggest to raise an exception here. We want to fail loudly and not silently :)

autopts/wid/bap.py
```diff
@@ -305,26 +316,27 @@ def hdl_wid_114(params: WIDParams):
...
+                                                  broadcast_id)
 
     btp.bap_broadcast_adv_start(broadcast_id)
 
     btp.bap_broadcast_source_start(broadcast_id)
 
     data = bytearray([j for j in range(0, 41)])
 
-    for i in range(1, 100):
+    for i in range(1, 10):
```
Thalley, line 331: This change won't affect Zephyr, so suggest to revert

autopts/wid/bap.py
```diff
@@ -2015,11 +2027,12 @@ def hdl_wid_380(_: WIDParams):
                                              audio_locations, octets_per_frame,
                                              frames_per_sdu)
 
+    broadcast_id = 0x123456
     presentation_delay = 40000
     streams_per_subgroup = 2
     subgroups = 1
     broadcast_id = btp.bap_broadcast_source_setup(
-        streams_per_subgroup, subgroups, coding_format, vid, cid,
+        0, streams_per_subgroup, subgroups, coding_format, vid, cid,
```
Thalley, line 2035:
```suggestion
        streams_per_subgroup, subgroups, coding_format, vid, cid,
```

and then add `broadcast_id` as the last parameter

autopts/pybtp/btp/bap.py
Thalley, line 133: Given that you've  positioned `broadcast_id` as the first field in the packet, I suggest to also keep it as the first argument to this function

Thalley (changes requested):

autopts/pybtp/btp/bap.py
alexsven: Ok, changed to '0x123456'

autopts/wid/bap.py
alexsven: Done, only changed it temporary for quicker runs while testing, but forgot to change it back

autopts/pybtp/btp/bap.py
Thalley: > Wouldn't it be better to keep it at the end of the packet as well? That way it doesn't affect those that doesn't look for it in the packet? Meaning they only put the first bytes up to the broadcast ID into a struct

Perhaps. My assumption would be that, since it's a static size command, that IUTs would do a `==` size check before parsing it (that's what Zephyr does at least), which would break regardless. 

But you are right that there's a higher chance (even if little) that it will be less disruptive if put at last (that's also the approach the BT SIG uses when releasing new versions of existing commands and events). 

I'd like to get input from @sjanc on this

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -25,8 +25,10 @@ def __init__(self):
     def __init__(self):
         self.peers = {}
         self.ase_configs = []
-        self.broadcast_id = 0x1000000  # Invalid Broadcast ID
+        self.broadcast_id = None
+        self.broadcast_id_2 = None
         self.broadcast_code = ''
+        self.hdl_wid_114 = 0
```
Thalley, line 31:
```suggestion
        self.hdl_wid_114_cnt = 0
```
Should indicate that this is a counter :)

autopts/pybtp/btp/bap.py
```diff
@@ -130,7 +130,7 @@ def bap_send(ase_id, data_ba, bd_addr_type=None, bd_addr=None):
 def bap_broadcast_source_setup(
         streams_per_subgroup, subgroups, coding_format, vid, cid,
         codec_ltvs, sdu_interval, framing, max_sdu, retransmission_number,
-        max_transport_latency, presentation_delay):
+        max_transport_latency, presentation_delay, broadcast_id=0x123456):
```
Thalley, line 133:
```suggestion
        max_transport_latency, presentation_delay, broadcast_id):
```

I would suggest not to provide any default values for the `broadcast_id`

alexsven: But then we would have to add them in every function that calls this, meaning the same value would be added regardless. So it was a means of reducing the breaking changes

Thalley: > So it was a means of reducing the breaking changes

If you are updating all calls to this, there are no breaking changes :) 

I think `bap_broadcast_source_setup` was originally defined as it is because Zephyr would generate the broadcast ID automatically. Since that is no longer the case, someone has to assign a broadcast ID. 

You are right that changing it so that it doesn't default doesn't really change anything if everyone would just call this with the same value, but now that broadcast_id is an IXIT value (at least for some profiles), I think it makes sense to set this explicitly, and to the broadcast ID set in the IXIT value. 

If PTS starts using the IXIT value for more tests/profiles, then it's easier to maintain that way

autopts/pybtp/btp/bap.py
alexsven: I see now why I put it at the start of the packet, there is an LTV at the end with an unknown size, meaning you would need to implement a handler for it, similar to `btp_gap_start_advertising_cmd`

Thalley: > I see now why I put it at the start of the packet, there is an LTV at the end with an unknown size, meaning you would need to implement a handler for it, similar to `btp_gap_start_advertising_cmd`

Ah, yeah that makes sense. Then it cannot be at the very end at least

sjanc: changing  command is always a problem unfortunately,  sometimes we were able to workaround this with extra commands that set a specific state in IUT (which could be used here),   very seldom we just broke BTP (but this was done only for newly added commands afair where only single [kown] user were present)

alternative would be to just add v2 command (with new opcode) and use it (if IUT claimed support for it)  [2]
yet, with this approach we should also be conservative since BTP supports only 127 commands per service...   this might not feel as big issue but... mesh services already have 100+ commands.. ;-)

Since this is likely to start becoming more of an issue, one solution could be to start thinking about versioning BTP,  we knew it would a thing at some point [1].  I think versioning (probably might be an easiest and simplest approach,   when IUT is using too old version of BTP we could simply fail early with dedicated error.  This would at least give very clear hint to update IUT BTP implementation instead of odd regressions in (random) tests..

[1] https://github.com/auto-pts/auto-pts/issues/337
[2] https://github.com/auto-pts/auto-pts/issues/1251

autopts/pybtp/btp/bap.py
```diff
@@ -74,6 +74,8 @@
     'set_broadcast_code': (defs.BTP_SERVICE_ID_BAP, defs.BTP_BAP_CMD_SET_BROADCAST_CODE,
                            CONTROLLER_INDEX),
     'bap_send_past': (defs.BTP_SERVICE_ID_BAP, defs.BTP_BAP_CMD_SEND_PAST, CONTROLLER_INDEX),
+    'broadcast_source_setup_v2': (defs.BTP_SERVICE_ID_BAP, defs.BTP_BAP_CMD_BROADCAST_SOURCE_SETUP_V2,
+                               CONTROLLER_INDEX),
```
Thalley, line 78:
```suggestion
    'broadcast_source_setup_v2': (defs.BTP_SERVICE_ID_BAP, defs.BTP_BAP_CMD_BROADCAST_SOURCE_SETUP_V2,
                                  CONTROLLER_INDEX),
```

autopts/wid/bap.py
```diff
@@ -305,12 +317,18 @@ def hdl_wid_114(params: WIDParams):
...
-    broadcast_id = btp.bap_broadcast_source_setup(
-        streams_per_subgroup, subgroups, coding_format, vid, cid,
-        codec_ltvs_bytes, *qos_config, presentation_delay)
 
-    stack = get_stack()
-    stack.bap.broadcast_id = broadcast_id
+    iut = get_iut()
+
+    if iut.name == "Zephyr":
+        log(f"Zephyr: bap_broadcast_source_setup_v2")
```
Thalley, line 324: @sjanc How much effort would it be to implement reading and using `BTP_BAP_READ_SUPPORTED_COMMANDS` instead of checking the iut.name here?

doc/btp_bap.txt
```diff
@@ -49,7 +49,8 @@ Commands and responses:
 
 	Opcode 0x04 - Broadcast Source Setup
 		Controller Index:	<controller id>
-		Command parameters:	Streams Per Subgroup (1 octet)
+		Command parameters:	Broadcast ID (3 octets)
+					Streams Per Subgroup (1 octet)
```
Thalley, line 53:
```suggestion
		Command parameters:	Streams Per Subgroup (1 octet)
```

doc/btp_bap.txt
```diff
@@ -317,6 +318,31 @@ Commands and responses:
...
+					LTVs length (1 octet)
+					LTVs (varies)
+		Response parameters:	Current_Settings (4 Octets) (see btp_gap.txt)
+					Broadcast ID (3 Octets)
+
+        This command is used to set up and configure a Broadcast Source
+        with the given Codec Config, with the addition of a fixed broadcast ID.
+        In case of an error, the error status response will be returned.
+        In case of a success, the IUT continues processing the command
+        asynchronously.
```
Thalley, line 344: Please spaces instead of tabs for indentation

Thalley (changes requested):

doc/btp_bap.txt
alexsven: As far as I can see, there is a mix in this file

Thalley: Also after https://github.com/auto-pts/auto-pts/pull/1390 ?

alexsven:
Needs this: https://github.com/zephyrproject-rtos/zephyr/pull/85377 to be merged first

doc/btp_bap.txt
```diff
@@ -338,6 +338,31 @@ Commands and responses:
...
+                                        Retransmission Num (1 octet)
+                                        Max Transport Latency (2 octets)
+                                        Presentation Delay (3 octets)
+                                        Coding Format (1 octet)
+                                        VID (2 octets)
+                                        CID (2 octets)
+                                        LTVs length (1 octet)
+                                        LTVs (varies)
+                Response parameters:    Current_Settings (4 Octets) (see btp_gap.txt)
+                                        Broadcast ID (3 Octets)
```
sjanc, line 358: why broadcast ID is needed in response?

autopts/wid/bap.py
sjanc: don't know, probably not that much (even if one would do it in generic way for all services)

doc/btp_bap.txt
alexsven: That was just because v1 had it, I guess it's a good way to verify that it hasn't been changed by Zephyr?

Thalley (approved):

sjanc (approved):

PR #1368, opened 2025-02-10
doc: Proposal for multiple advertising set support

jrintaha:
Adds the following commands and events:
 - Create Advertising Set
 - Delete Advertising Set
 - Update Advertising Set Parameters
 - Update Advertising Set Data
 - Start Advertising Set
 - Stop Advertising Set
 - Advertising Set Terminated event

These allow the BTP host device to manage multiple advertising sets, exposing almost all of the functionality of the Zephyr host stack APIs. This includes things such as legacy vs extended advertising, secondary PHY selection, and advertising timings.

I've attempted to make the BTP API as user-friendly as possible, grouping logically connected things into distinct parameters rather than using giant bitfields, while at the same time not removing any relevant functionality that the Zephyr host stack APIs provide. If anyone should spot any omissions, I'll be happy to fix them.

This proposal does not modify any existing definitions, so backwards compatibility should not be an issue.

The background for this proposal is that we've prototyped using the BTP protocol for continuous integration testing, and in that context it would be very useful to extend the BTP protocol to handle multiple advertising sets, as the current advertising API that BTP offers is rather limited. It could be useful for projects like https://github.com/JuulLabs-OSS/BTPTesterCore as well.

I'll link the corresponding implementation of the tests/bluetooth/tester app once its PR is up.

### DIFF
--- a/doc/btp_gap.txt
+++ b/doc/btp_gap.txt
@@ -192,12 +192,12 @@ Commands and responses:
 
 		Possible values for the Flags parameter are a bit-wise or
 		of the following bits:
-					0 = LE scan
-					1 = BR/EDR scan
-					2 = Use limited discovery procedure
+					0 = LE scan (deprecated, always enabled)
+					1 = BR/EDR scan (deprecated, not supported)
+					2 = Use limited discovery procedure (deprecated, ignored)
 					3 = Use active scan type
-					4 = Use observation procedure
-					5 = Use own ID address
+					4 = Use observation procedure (deprecated, ignored)
+					5 = Use own ID address (deprecated, ignored)
 
 		This command is used to start discovery.
 
@@ -637,6 +637,180 @@ Commands and responses:
 
 		In case of an error, the error response will be returned.
 
+	Opcode 0x2a - Create Advertising Set
+
+		Controller Index:	<controller id>
+		Command parameters:	<none>
+		Return Parameters:	Created Advertising Set ID (1 octet)
+
+		This command is used to create an advertising set, which can be used
+		for both legacy and extended advertisements. The set is created in
+		the stopped state.
+
+		The set is created with the default parameters of:
+			SID (1 octet) = 0
+			Address Type (1 octet) = 0x00 (Identity)
+			Advertising Type (1 octet) = 0x00 (Legacy Not Connectable, Not Scannable)
+			Filter Accept List (1 octet) = 0x00 (Do not use the filter accept list)
+			Directed Advertising (1 octet) = 0x00 (Do not use directed advertising)
+			Options (2 octets) = 0x0000
+			Minimum Interval (4 octets) = 160 (100 ms)
+			Maximum Interval (4 octets) = 320 (200 ms)
+			Peer Address Type (1 octet) = 0x00 (Public)
+			Peer Address (6 octets) = 00:00:00:00:00:00
+
+		The parameters can be changed with the Update Advertising Set
+		Parameters command.
+
+		The set can be deleted with Delete Advertising Set.
+
+		The return value is the ID of the newly created set.
+
+		Note that the state of advertising sets created with this command do
+		not affect the Current_Settings state.
+
+		In case of an error, the error response will be returned.
+
+	Opcode 0x2b - Delete Advertising Set
+
+		Controller Index:	<controller id>
+		Command parameters:	Advertising Set Id (1 octet)
+		Return Parameters:	<none>
+
+		This command is used to delete a previously created advertising set.
+		If the advertising set is currently active, it will be stopped.
+
+		In case of an error, the error response will be returned.
+
+	Opcode 0x2c - Update Advertising Set Parameters
+
+		Controller Index:	<controller id>
+		Command parameters:	Advertising Set Id (1 octet)
+					SID (1 octet)
+					Address Type (1 octet)
+					Advertising Type (1 octet)
+					Filter Accept List (1 octet)
+					Directed Advertising (1 octet)
+					Options (2 octets)
+					Minimum Interval, in 0.625ms units (4 octets)
+					Maximum Interval, in 0.625ms units (4 octets)
+					Peer Address Type (1 octet)
+					Peer Address (6 octets)
+		Return Parameters:	<none>
+
+		This command is used to configure the parameters of a previously created
+		advertising set. The advertising set must be stopped when the command
+		is given, otherwise an error will be emitted.
+
+		The allowed values for SID are 0...15.
+
+		The allowed values for Address Type are:
+			0x00 = Identity Address
+			0x01 = Private Address (resolvable if connectable)
+
+		The allowed values for Advertising Type are:
+			0x00 = Legacy Not Connectable, Not Scannable
+			0x01 = Legacy Connectable
+			0x02 = Legacy Scannable
+			0x03 = Legacy Connectable and Scannable
+			0x04 = Extended Not Connectable, Not Scannable
+			0x05 = Extended Connectable
+			0x06 = Extended Scannable
+			0x08 = Coded Not Connectable, Not Scannable
+			0x09 = Coded Connectable
+			0x0a = Coded Scannable
+
+		The allowed values for Filter Accept List are:
+			0x00 = Do not use the filter accept list
+			0x01 = Use the filter accept list for connections
+			0x02 = Use the filter accept list for scan requests
+			0x03 = Use the filter accept list for connections and scan requests
+
+		The allowed values for Directed Advertising are:
+			0x00 = Do not use directed advertising
+			0x01 = Use high duty directed advertising
+			0x02 = Use low duty directed advertising
+
+		The Options bitfield is a bit-wise OR of the following bits:
+			0 = Advertise with device name
+			1 = Disable 2M PHY for extended advertising
+			2 = Anonymous advertising for extended advertising
+			3 = Include TX Power in extended advertising
+			4-15 = Reserved
+
+		Minimum Interval must be less than or equal to Maximum Interval.
+
+		The allowed values for Peer Address Type are:
+			0x00 = Public
+			0x01 = Random
+			0x02 = Resolvable
+
+		The Peer Address and Peer Address Type fields will be ignored if
+		directed advertising is not used.
+
+		In case of an error, the error response will be returned.
+
+	Opcode 0x2d - Update Advertising Set Data
+
+		Controller Index:	<controller id>
+		Command parameters:	Advertising Set Id (1 octet)
+					Adv Data Length (2 octets)
+					Scan Response Length (2 octets)
+					Adv Data (0-1650 octets)
+					Scan Response Data (0-1650 octets)
+		Return Parameters:	<none>
+
+		This command is used to set the advertising and scan response data
+		for an advertising set.
+
+		The advertising set must have been previously configured appropriately
+		for the data. For example, if more than 31 octets of advertising or scan
+		response data is given, the set must have been configured to use extended
+		PDUs.
+
+		If scan response data is given but the set is not configured as scannable,
+		then the scan response data is ignored.
+
+		Note that if extended PDUs are used, and the set is scannable, the set
+		can only have scan response data and no advertising data.
+
+		If the advertising set is currently advertising, the data will be
+		updated at the next advertising interval.
+
+		In case of an error, the error response will be returned.
+
+	Opcode 0x2e - Start Advertising Set
+
+		Controller Index:	<controller id>
+		Command parameters:	Advertising Set Id (1 octet)
+					Timeout (2 octets)
+					Number of Events (1 octet)
+		Return Parameters:	<none>
+
+		This command is used to start a previously defined advertising set.
+
+		The timeout parameter is in units of 10ms. Set to zero for no timeout.
+
+		Set the Number of Events to zero for unlimited.
+
+		Note that starting advertising sets with this command does not affect
+		the Current_Settings state.
+
+		In case of an error, the error response will be returned.
+
+	Opcode 0x2f - Stop Advertising Set
+
+		Controller Index:	<controller id>
+		Command parameters:	Advertising Set Id (1 octet)
+		Return Parameters:	<none>
+
+		This command is used to stop a previously started advertising set.
+
+		Note that stopping advertising sets with this command does not affect
+		the Current_Settings state.
+
+		In case of an error, the error response will be returned.
+
 Events:
 	Opcode 0x80 - New Settings event
 
@@ -823,3 +997,13 @@ Events:
 
 		This event can be sent after IUT received Periodic Advertising Sync
 		Transfer was received.
+
+	Opcode 0x91 - Advertising Set Terminated event
+
+		Controller Index:	<controller id>
+		Event parameters:	Advertising Set ID (1 octet)
+
+		This event is sent when an advertising set is terminated by the timeout
+		or maximum events mechanism. This event is not sent when a set is
+		explicitly stopped, or when a connection establishment causes the set to
+		stop.


doc/btp_gap.txt
```diff
@@ -192,12 +192,12 @@ Commands and responses:
 
 		Possible values for the Flags parameter are a bit-wise or
 		of the following bits:
-					0 = LE scan
-					1 = BR/EDR scan
-					2 = Use limited discovery procedure
+					0 = LE scan (deprecated, always enabled)
```
jrintaha, line 195: I documented this command to match the current implementation of the tester application. Most of these bits don't actually do anything.

Thalley:
@alexsven This is somewhat related to one of the changes you've done for multiple advertising sets. Perhaps the two approaches can be aligned?

jrintaha:
https://github.com/zephyrproject-rtos/zephyr/pull/85509 here's the corresponding implementation

doc/btp_gap.txt
```diff
@@ -637,6 +637,180 @@ Commands and responses:
...
+			2 = Anonymous advertising for extended advertising
+			3 = Include TX Power in extended advertising
+			4-15 = Reserved
+
+		Minimum Interval must be less than or equal to Maximum Interval.
+
+		The allowed values for Peer Address Type are:
+			0x00 = Public
+			0x01 = Random
+			0x02 = Resolvable
```
sjanc, line 746: how 0x02 differs from 0x01 here?

doc/btp_gap.txt
```diff
@@ -637,6 +637,180 @@ Commands and responses:
...
+		advertising set. The advertising set must be stopped when the command
+		is given, otherwise an error will be emitted.
+
+		The allowed values for SID are 0...15.
+
+		The allowed values for Address Type are:
+			0x00 = Identity Address
+			0x01 = Private Address (resolvable if connectable)
+
+		The allowed values for Advertising Type are:
```
sjanc, line 711: I'd prefer to have this defined as flags, not values (easier to extend if needed), eg:
connectable, scannable, legacy_pdu, directed (although this could be implicit based on peer addr), anonymous, etc

also some options seems to be to zephyr's API specific (eg disabling 2M phy or using device name)

doc/btp_gap.txt
```diff
@@ -823,3 +997,13 @@ Events:
...
+
+	Opcode 0x91 - Advertising Set Terminated event
+
+		Controller Index:	<controller id>
+		Event parameters:	Advertising Set ID (1 octet)
+
+		This event is sent when an advertising set is terminated by the timeout
+		or maximum events mechanism. This event is not sent when a set is
+		explicitly stopped, or when a connection establishment causes the set to
+		stop.
```
sjanc, line 1009: hmm should this also be sent on connection establishment?  if purpose of this BTP API is to support multiple instances, not sending this event on connection could result in unknown state

doc/btp_gap.txt
jrintaha: Ah yes, I'll revert these changes.

doc/btp_gap.txt
jrintaha: You're right; although the connection events themselves can usually be used to determine the state of an advertising set, there could be an implementation that does not stop advertising upon connection establishment, or you could have multiple connectable advertising sets. I'll amend the description.

doc/btp_gap.txt
jrintaha: Plain old mistake here. Copypasted a few lines, edited halfway through. I'll remove this line.

doc/btp_gap.txt
```diff
@@ -637,6 +637,180 @@ Commands and responses:
 
 		In case of an error, the error response will be returned.
 
+	Opcode 0x2a - Create Advertising Set
```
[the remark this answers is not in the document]
Thalley: I would tend to agree with @jrintaha - Some of the issues we see today with BTP is that it has tried to do very specific things with many parameters, which makes it easier to use, but also less flexible and thus harder to maintain when PTS (or spec updates) changes the behavior. 

I think that it's counter-intuitive is easier to main the autopts client when it does more, and there's less IUT-specific behavior that we need to rely on. The downside of this is of course that it may lead to more changes in the autopts client that needs to be handled by the IUTs if not already.

doc/btp_gap.txt
```diff
@@ -645,6 +645,192 @@ Commands and responses:
...
+		be used for both legacy and extended advertisements.
+
+		If the Advertising Set ID is 0xFF, a new advertising set will be created.
+
+		Otherwise, the command will configure the parameters of the specified
+		advertising set. The advertising set must be stopped when the command
+		is given, otherwise an error will be emitted.
+
+		The allowed values for Advertising Set ID are:
+			0x00...0xef = Use the specified advertising set
```
jrintaha, line 678: This corresponds to the range allowed by the HCI spec, so we're not losing any functionality here by having the special value 0xff as the create flag.

doc/btp_gap.txt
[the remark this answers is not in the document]
jrintaha: I've attempted to resolve the issues here. There were a few missing bits and pieces such as channel map configuration and Coded PHY coding selection. There is still room in the Options bitfield so if new things must be added, that can hopefully accommodate them.

jhedberg:
@Thalley @sjanc is this good to go in?

doc/btp_gap.txt
```diff
@@ -645,6 +645,189 @@ Commands and responses:
...
+					Minimum Interval, in 0.625ms units (4 octets)
+					Maximum Interval, in 0.625ms units (4 octets)
+					Peer Address Type (1 octet)
+					Peer Address (6 octets)
+		Return Parameters:	Configured Advertising Set ID (1 octet)
+
+		This command is used to create and configure advertising sets, which can
+		be used for both legacy and extended advertisements.
+
+		If the Advertising Set ID is 0xFF, a new advertising set will be created.
```
Thalley, line 671: Why not treat it similar to HCI? 
i.e. if the SID is not already create, it will create a set with that ID, else it will (attempt) to reconfigure an existing advertising set

doc/btp_gap.txt
```diff
@@ -645,6 +645,189 @@ Commands and responses:
 
                 In case of an error, the error response will be returned.
 
+	Opcode 0x2a - Configure Advertising Set
+
+		Controller Index:	<controller id>
+		Command parameters:	Advertising Set ID (1 octet)
+					SID (1 octet)
```
Thalley, line 652: SID = Advertising Set ID

Suggest to use `Advertising Handle` instead of Set ID (will be more similar to HCI). 

The downside is that not all IUT may support setting the advertising handle, but that argument could probably be said for the SID as well

doc/btp_gap.txt
```diff
@@ -645,6 +645,189 @@ Commands and responses:
...
+
+	Opcode 0x2c - Update Advertising Set Data
+
+		Controller Index:	<controller id>
+		Command parameters:	Advertising Set Id (1 octet)
+					Adv Data Length (2 octets)
+					Scan Response Length (2 octets)
+					Adv Data (0-1650 octets)
+					Scan Response Data (0-1650 octets)
+		Return Parameters:	<none>
```
Thalley, line 778: Consider splitting this in 2: One for advertising data, and one for scan response data. 

For extended advertising only one can usually be set anyhow

Thalley (changes requested):

doc/btp_gap.txt
```diff
@@ -645,6 +645,189 @@ Commands and responses:
...
+		advertising set. The advertising set must be stopped when the command
+		is given, otherwise an error will be emitted.
+
+		The allowed values for Advertising Set ID are:
+			0x00...0xef = Use the specified advertising set
+			0xff = Create a new advertising set
+
+		The allowed values for SID are 0...15.
+
+		The allowed values for Own Address Type are:
```
sjanc, line 683: on advertising command we have     ID, RPA, NRPA for own_addr

what address would be used when selected random here?

doc/btp_gap.txt
```diff
@@ -645,6 +645,189 @@ Commands and responses:
...
+			0x00 = Identity Address
+			0x01 = Random Address
+
+		The allowed values for Advertising Type are a bit-wise OR of the
```
sjanc, line 687: I'd suggest to make this bigger and name it flags and combine more options here

doc/btp_gap.txt
```diff
@@ -645,6 +645,189 @@ Commands and responses:
...
+			0x04 = S=8 coding required
+
+			Ignored if Coded PHY is not used.
+
+		The allowed values for Filter Accept List are a bit-wise OR of the
+		following bits:
+			0 = Use the filter accept list for scan requests
+			1 = Use the filter accept list for connections
+
+		The allowed values for Directed Advertising are:
```
sjanc, line 728: this could be HD selection flag only, and if directed to be used could be based on value of peer_addr

doc/btp_gap.txt
```diff
@@ -645,6 +645,189 @@ Commands and responses:
...
+
+		The return value is the ID of the newly created set, if the given
+		set ID was 0xFF. Otherwise, the return value is the given set ID.
+
+		Note that the state of advertising sets created with this command do
+		not affect the Current_Settings state.
+
+		In case of an error, the error response will be returned.
+
+	Opcode 0x2b - Delete Advertising Set
```
sjanc, line 759: why is this needed?  stopping advertising should do

doc/btp_gap.txt
```diff
@@ -645,6 +645,189 @@ Commands and responses:
...
+		This command is used to set the advertising and scan response data
+		for an advertising set.
+
+		The advertising set must have been previously configured appropriately
+		for the data. For example, if more than 31 octets of advertising or scan
+		response data is given, the set must have been configured to use extended
+		PDUs.
+
+		If scan response data is given but the set is not configured as scannable,
+		then the scan response data is ignored.
```
sjanc, line 789: I'd prefer explicit error than silently ignoring

doc/btp_gap.txt
```diff
@@ -645,6 +645,189 @@ Commands and responses:
...
+
+		Note that if extended PDUs are used, and the set is scannable, the set
+		can only have scan response data and no advertising data.
+
+		If the advertising set is currently advertising, the data will be
```
sjanc, line 794: this (at least on HCI) allowed only if al data fits in single HCI command

doc/btp_gap.txt
```diff
@@ -842,3 +1025,13 @@ Events:
 
                 This event can be sent after IUT received Periodic Advertising Sync
                 Transfer was received.
+
+	Opcode 0x91 - Advertising Set Terminated event
+
+		Controller Index:	<controller id>
+		Event parameters:	Advertising Set ID (1 octet)
```
sjanc, line 1032: should we have peer_addr in case of termination due to connection?

sjanc (commented):
it feels very "HCI",   at least some of those could be done in more BTP (GAP) way by using flags instead of mappings from HCI

(note that I'm much against following HCI, just wondering if this is intention here,  if one wants to test HCI he should use... well HCI, not BTP)

doc/btp_gap.txt
pekozar: If something can be created, it is natural to have a delete counterpart.
The LE Remove Advertiser Set HCI command does exist, so does the bt_le_ext_adv_delete() API.
What's wrong with using it?

doc/btp_gap.txt
pekozar: The problem using the Advertiser Handle is that it does not exist from our perspective.
The advertiser is allocated somewhere and then the "handle" in use is the opaque pointer to that structure.
Due to the internal header definitions, the actual handle cannot be fetched as far as I've seen even if it does exist within.
Basically the tester application can only define a static  array of advertiser objects it wants to create and use, and the ID is in the index within this array, and that is the only ID that can be serialised over BTP. 
We could call it handle, but that would be probably confusing as well, just for other reasons.

doc/btp_gap.txt
pekozar: Creation and configuration is merged into a single command as suggested, but the ID is the index within the static array of advertiser as commented above. 0xFF is can be used to create a new one if there is a free slot, a specific index will create/configure the given one.

doc/btp_gap.txt
pekozar: Can return EINVAL rather than ignoring or could be split as described above, but consensus is needed about the way forward.

doc/btp_gap.txt
pekozar: There are all sorts of limitations, so probably the easiest way is to remove the description and just rely on the advertiser behaviour, whatever it is.

doc/btp_gap.txt
Thalley: I see

In that case, I suggest to remove one of the 2 IDs here. It doesn't really help to use have 2 IDs in a command (especially not when they have the same name ;) )

I'm fine with just using either `SID` or `Advertising Set ID`

pekozar: The naming isn't too convenient, but while we need some serialisable id for the set, the SID is the Advertising SID subfield in the ADI field of the PDU, so that is actual value that would be transmitted.

doc/btp_gap.txt
pekozar: The bt_le_ext_adv_set_data() has both parameters and it is in line with the legacy implementation.
Can be split but there are two ways:
- separate functions to input one of the values, and call bt_le_ext_adv_set_data with the appropriate value and NULL for the other
- one function with one input and an addition parameter to indicate the type

Which one would be preferred?

doc/btp_gap.txt
Thalley: > The naming isn't too convenient, but while we need some serialisable id for the set, the SID is the Advertising SID subfield in the ADI field of the PDU, so that is actual value that would be transmitted.

True, and I guess that, unlike the advertising handle, a single device may use the same SID for different advertising sets assuming they have different advertising addresses, so it's not _as unique_ as the advertising handle, or the advertising set ID you are proposing. 

For connections we rely on the remote address as an identifier, so that's how we got around with no having the connection handle. 

When you say that you don't have the advertising handle, is that with Zephyr or another stack?

doc/btp_gap.txt
Thalley: Hmm, rethinking it, then, at least for the Zephyr API, it probably makes more sense to keep them both in the same command

doc/btp_gap.txt
pekozar: @Thalley: It is Zephyr that does not provide a numeric handle

Thalley: Zephyr does have `bt_hci_get_adv_handle` that can provide the handle in cases where it's needed, so it's not a blocker for Zephyr

pekozar: Agreed, it's not a blocker per se, the numeric handle can be queried for the adv set struct pointer, and there is a lookup for the set based on a numeric value. You are also correct regarding the fact that multiple sets can use the same SID within the PDUs so only the handle is the real differentiator.

However, the API functions all work solely based on the pointer.
The tester application until recently had just this definition:

static struct bt_le_ext_adv *ext_adv;

With the last upstream changes, now it has an array as follows:
#if CONFIG_BT_EXT_ADV
static struct bt_le_ext_adv *ext_adv_sets[CONFIG_BT_EXT_ADV_MAX_ADV_SET];
#else
static struct bt_le_ext_adv *ext_adv_sets[1];
#endif

The implementation relies on the indexed elements within this array.
This is why the proposed changes included the numeric handle, which would track the sets within this array,
otherwise the tester app must have the two way mapping between the btp handle and the set pointer.

The current tester implementation is a bit spaghetti-ish, but to minimize the necessary changes, the array indices could simply be exposed as handles via btp.

doc/btp_gap.txt
pekozar: There's a misalignment here between possible HCI level configuration and Zephyr, for example.
Zephyr seems to only support public identity or "something else", which is then managed entirely by the privacy subsystem if enabled by the configuration.
If you have any suggestions for a possible common definition, I would be very keen to hear about it.
The other option is that we define some subset that is workable right now and let it evolve based on future needs. Let me know what you think.

doc/btp_gap.txt
pekozar: Could you please clarify what flags are missing here based on the Core spec, or what do you mean by more and combined options?

doc/btp_gap.txt
pekozar: HCI defines the Advertising Event Properties as having separate Directed Advertising and HD flag.
This parameter would just combine the options of having directed advertising and if so, what kind.
These flags are defined in HCI regardless of the peer address, so I'm not sure I understand the intent of sometimes defining things as is in HCI, and going against it in other cases.

doc/btp_gap.txt
pekozar: peer address is not part of the LE Advertising Set Terminated event, only the connection handle.
That can be certainly added, or with a bit of additional code, the lookup can be performed for the address based on the connection handle and inserted here. I'm assuming that should be supported by all platforms and HCI events race conditions between termination and connection would not apply to querying the value.

sjanc: in BTP peer are identified by address (as it is not HCI level for various reasons) so this will have to be handled on IUT side

doc/btp_gap.txt
sjanc: BTP doesn't have to be direct mapping for zephyr APIs (nor HCI) so it may also be used by other IUTs (and zephyr APIs also changes occasionally),  it should allow for configuring IUT to suit test purposes and than BTP app on IUT (tester app in zephyr) can map this as needed.

For own address here (since this is not for qualification but e2e testing) I'd suggest to also add random address to command (used when 0x01 own addr is used, this way upper tester may configure whatever is needed).

doc/btp_gap.txt
```diff
@@ -892,6 +892,189 @@ Commands and responses:
...
+					SID (1 octet)
+					Own Address Type (1 octet)
+					Advertising Type (1 octet)
+					Primary Advertising PHY (1 octet)
+					Secondary Advertising PHY (1 octet)
+					Secondary Advertising Maximum Skip (1 octet)
+					Coded PHY Options (1 octet)
+					Filter Accept List (1 octet)
+					Directed Advertising (1 octet)
+					Options (2 octets)
```
sjanc, line 908: I'd suggest to combine coded phy options, filer accept list options, directed options and options into single 32bit options (or flags) that would allow to tune IUT behavior,  this way we could extend BTP more easily if eg new options (eg CTE) or PHYs are introduced  (as mentioned, BTP doesn't have to map 1-1 to HCI or specific IUT apis)

doc/btp_gap.txt
```diff
@@ -892,6 +892,189 @@ Commands and responses:
 
                 In case of an error, the error response will be returned.
 
+	Opcode 0x34 - Configure Advertising Set
+
+		Controller Index:	<controller id>
+		Command parameters:	Advertising Set ID (1 octet)
+					SID (1 octet)
+					Own Address Type (1 octet)
+					Advertising Type (1 octet)
```
sjanc, line 901: I'd suggest to make this uint32 and add also directed adv flags here  (this and combined 32bit options should give enough future proof if we need to extend this)

PR #1383, opened 2025-02-19
doc: btp_gap: Adds set rpa timeout

kedMertens:
Proposal to add gap_set_rpa_timeout command, which would enable dynamically changing rpa timeout.
I've used 0x30 opcode to respect  #1368.

### DIFF
--- a/doc/btp_gap.txt
+++ b/doc/btp_gap.txt
@@ -637,6 +637,17 @@ Commands and responses:
 
 		In case of an error, the error response will be returned.
 
+        Opcode 0x30 - Set RPA Timeout command/response
+
+                Controller Index: <controller id>
+                Command parameters: RPA Timeout in seconds (2 octets)
+                Return Parameters: <none>
+
+                This command is used to set RPA timeout. The new RPA timeout value will be
+                used for the next RPA rotation. Valid range is 1s - 3600s.
+
+                In case of an error, the error response will be returned.
+
 Events:
 	Opcode 0x80 - New Settings event
 


doc/btp_gap.txt
```diff
@@ -637,6 +637,17 @@ Commands and responses:
 
 		In case of an error, the error response will be returned.
 
+  Opcode 0x30 - Set RPA Timeout command/response
+
+    Controller Index: <controller id>
+    Command parameters: RPA Timeout (2 octets)
```
jhedberg, line 643: Perhaps this should state the unit as well? Is it in seconds?

kedMertens: It is in seconds, mentioned in description. I've checked other timeout parameters and noticed that they don't provide units, so omitted here to match style. However might be useful to have it.

jhedberg: The description specifies the valid range, but it still doesn't explicitly say what the unit of the parameter is. The unit could be milliseconds and the valid range would still make sense. Point taken about this missing elsewhere, but I think it's better to be very clear about these details in a specification document.

Thalley (approved):

jhedberg (approved):

kedMertens:
Would appreciate if someone could merge, @sjanc

sjanc (commented):
looks ok, just nitpick if you could fix whitespacing

kedMertens:
> looks ok, just nitpick if you could fix whitespacing

My bad

sjanc (approved):

Issue #1377, opened 2025-02-18
Log an info about BadPixitException into an error.txt file


mkasenberg:
A BadPixitException like this one:
```bash
xmlrpc.client.Fault: <Fault 1: "<class 'autopts.ptscontrol.BadPixitException'>:L2CAP TSPX_inmtu 02A0:\nPTSCONTROL_E_PIXIT_PARAM_NOT_FOUND">
```

is raised when we try to set/update PIXIT using an invalid name or value. The autopts client exits as expected, but the autopts cron restarts the test setup indefinitely. Let's log an info that BadPixitException was raised into the ERROR_TXT_FILE, like we do here:
https://github.com/auto-pts/auto-pts/blob/6490950a01b0d99e310c914159ffddae9575009c/autopts/bot/zephyr.py#L144-L145

and check the existence of the ERROR_TXT_FILE in the cron loop:
https://github.com/auto-pts/auto-pts/blob/6490950a01b0d99e310c914159ffddae9575009c/tools/cron/common.py#L601

Full stack trace of the BadPixitException for the reference:
```bash
Traceback (most recent call last):
  File "/home/codecoup/workspace/autopts/auto-pts/autoptsclient_bot.py", line 151, in <module>
    rc = main()
  File "/home/codecoup/workspace/autopts/auto-pts/autoptsclient_bot.py", line 125, in main
    bot_client.start()
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/bot/mynewt.py", line 150, in start
    super().start(args)
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/bot/common.py", line 502, in start
    stats = self.run_tests()
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/bot/common.py", line 572, in run_tests
    return super().start()
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/client.py", line 1375, in start
    return self.main(args)
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/client.py", line 1420, in main
    self.setup_project_pixits(self.ptses)
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/client.py", line 1439, in setup_project_pixits
    setup_project_pixits(ptses)
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/client.py", line 1608, in setup_project_pixits
    mod.set_pixits(ptses)
[...]

PR #1384, opened 2025-02-19
tools: cron: stop cron in case of error

piotrnarajowski:
ERROR_TXT_FILE will be written to when invalid PIXIT name or value is set. We check for this file in _run_test function from cron to stop it from restarting indefinitely.

Fixes https://github.com/auto-pts/auto-pts/issues/1377

### DIFF
--- a/autopts/bot/common.py
+++ b/autopts/bot/common.py
@@ -500,6 +500,9 @@ class BotClient(Client):
 
         try:
             stats = self.run_tests()
+        except:
+            report.make_error_txt(traceback.format_exc(), self.file_paths['ERROR_TXT_FILE'])
+            raise
         finally:
             release_device(self.args.tty_file)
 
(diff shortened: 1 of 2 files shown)


autopts/client.py
```diff
@@ -1436,7 +1436,7 @@ def init_iutctl(self, args):
         autoprojects.iutctl.init(args)
 
     def setup_project_pixits(self, ptses):
-        setup_project_pixits(ptses)
+        setup_project_pixits(self.file_paths, ptses)
```
mkasenberg, line 1439: `make_error_txt()` doesn't do much now, so let's write directly to the error file to avoid the import from autopts.bot module.
```suggestion
        try:
            setup_project_pixits(ptses)
        except xmlrpc.client.Fault:
            with open(self.file_paths['ERROR_TXT_FILE'], "w") as f:
                f.write('Bad Pixit Exception')
            raise
```

mkasenberg: After second thought, if we could handle all exceptions in bot layer in this way:
```python
try:
    stats = self.run_tests()
except:
    with open(self.file_paths['ERROR_TXT_FILE'], "w") as f:
        f.write(traceback.format_exc())
        raise
finally:
    release_device(self.args.tty_file)
```

mkasenberg: It is here https://github.com/auto-pts/auto-pts/blob/b2f98b6155b371a46a49e53d005886d7ac683624/autopts/bot/common.py#L501-L504

piotrnarajowski: thanks, applied your idea

PR #1386, opened 2025-02-20
autoptsserver: add --dongle_init_retry param

piotrnarajowski:
autoptsserver will try to launch for specified amount of times with unplugged dongle before terminating the process. Use Param value 0 to disable this


autoptsserver.py
```diff
@@ -313,6 +313,11 @@ def __init__(self, description):
                           nargs="+", default=[SERVER_PORT],
                           help="Specify the server port number")
 
+        self.add_argument("--dongle_init_retry", type=int, default=5,
+                          help="autoptsserver will try to launch for specified amount of times "
+                               "with unplugged dongle before terminating the process. Use Param "
+                               "value 0 to disable this")
```
mkasenberg, line 319:
```suggestion
                               "value 0 to retry indefinitely")
```

mkasenberg (approved):

PR #1387, opened 2025-02-21
CAP: Only adv as precondition when CAP/ACC

alexsven:
- Add ternary operation to start adv when tc_name includes CAP/ACC
- Move bd_addr IXIT update to pre_conditions as this was done for all tests
- Fixes regression for CAP/INI/BST tests

### DIFF
--- a/autopts/ptsprojects/zephyr/cap.py
+++ b/autopts/ptsprojects/zephyr/cap.py
@@ -168,20 +168,18 @@ def test_cases(ptses):
         TestFunc(btp.gap_set_extended_advertising_on),
         # Gives a signal to the LT2 to continue its preconditions
         TestFunc(lambda: set_addr(stack.gap.iut_addr_get_str())),
+        TestFunc(lambda: pts.update_pixit_param("CAP", "TSPX_bd_addr_iut",
+                                                stack.gap.iut_addr_get_str()))
     ]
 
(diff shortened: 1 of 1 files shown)


Thalley (changes requested):
Please almost add information in the git message why the IXIT was moved, and which tests are affected by this

alexsven:
> Please almost add information in the git message why the IXIT was moved, and which tests are affected by this

Done

Thalley (changes requested):
Suggest to rewrite 
`Fixes regression for CAP/INI/BST tests` 
as
`Fixes regression for CAP/INI/BST tests where broadcast advertisement would not start in Zephyr`

Thalley (approved):

autopts/ptsprojects/zephyr/cap.py
```diff
@@ -638,7 +636,7 @@ def set_addr(addr):
     for tc_name in test_case_name_list:
 
         instance = ZTestCase("CAP", tc_name,
-                             cmds=pre_conditions + targeted_conditions,
+                             cmds=pre_conditions + (targeted_conditions if "CAP/ACC" in tc_name else []),
```
sjanc, line 639: typically we do this with if-else but this is OK too I guess

sjanc (approved):

PR #1390, opened 2025-02-24
docs: Clean up whitespaces and unify convention used

sjanc:
This unify documentation format, initial idea was to use tabs (8 spaces wide) for indentation but this seems to be causing discrepancies in editors and with time it was not really coherent (various places asuming 4 spaces tabs, or indentation with spaces only).

Just unify this and use spaces for indentation (8 spaces for level).

This also fixes some convention discrepancies in BTP descriptions so it is easier to follow.

### DIFF
--- a/doc/btp_aics.txt
+++ b/doc/btp_aics.txt
@@ -3,289 +3,289 @@ Audio Input Control Service (ID 10)
 
 Commands and responses:
 
-	Opcode 0x00 - Error response
+        Opcode 0x00 - Error response
 
-	Opcode 0x01 - Read supported commands
+        Opcode 0x01 - Read supported commands
 
-		Controller Index:	<controller id>
-		Command parameters:	<none>
-		Response parameters:	<supported commands> (variable)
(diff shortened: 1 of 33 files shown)


Thalley:
Can we somehow define this convention in a file somewhere? 

Is AutoPTS married to using .txt for documentation? 
Since AutoPTS is Python based, we could in theory (huge refactor) define each command and event as a Python class and use Python's docstring to handle documentation in the future, while also ensuring that documentation is always added and all fields are documented. 

If we want to keep BTP and Python separated, then that's also fine :) (we could also consider other formats like rst or md)

sjanc:
it was txt for historical reasons  (initially it was stored in zephyr's repo, but moved here in 2019 when more projects started using autoPTS)

sjanc:
I'm fine to move it to python in long term and have generated documention for it based on that (ie nice md or html that can be viewed in browser)

sjanc:
no,  but I'll add simple script that checks at least for tabs not being used

PR #1404, opened 2025-03-04
TMAP: Allow time for broadcast to stop

alexsven:
- Add sleep to allow time between stop and release
- Fixes TMAP/BMS/ASC/BV-01-C

### DIFF
--- a/autopts/wid/tmap.py
+++ b/autopts/wid/tmap.py
@@ -16,6 +16,7 @@
 import logging
 
 import re
+import time
 from argparse import Namespace
 
 from autopts.ptsprojects.stack import get_stack
@@ -562,6 +563,7 @@ def hdl_wid_506(params: WIDParams):
     source_id = 0x00
     btp.cap_broadcast_adv_stop(source_id)
     btp.cap_broadcast_source_stop(source_id)
+    time.sleep(10) # Allow some time for the stack to stop the broadcast
     btp.cap_broadcast_source_release(source_id)
 
     # Get Audio Locations from description


Thalley:
#AutoPTS run zephyr nrf53 TMAP/BMS/ASC

codecoup-tester:
Scheduled PR https://github.com/auto-pts/auto-pts/pull/1404#issuecomment-2697325173, board: nrf53, estimated start time: 13:08:04, test case count: 2, estimated duration: 0:10:25<details><summary>Test cases to be run</summary>TMAP/BMS/ASC/BV-01-C<br>TMAP/BMS/ASC/BV-02-C</details>

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (2)</summary>TMAP    TMAP/BMS/ASC/BV-01-C            PASS<br>TMAP    TMAP/BMS/ASC/BV-02-C            PASS</details>

alexsven:
> why this is needed in first place? this doesn't require any radio operations so should be instantaneous

It seems the state change takes some time. Before this fix I got TMAP/BMS/ASC/BV-01-C passing 6/10 times, after it is passing 10/10 times. When it's failing I get the response -EBADMSG that comes from [here](https://github.com/zephyrproject-rtos/zephyr/blob/14f382b3b61c893fcf0302b822f6d14f80f547e9/subsys/bluetooth/audio/bap_broadcast_source.c#L1145)

autopts/wid/tmap.py
```diff
@@ -562,6 +563,7 @@ def hdl_wid_506(params: WIDParams):
     source_id = 0x00
     btp.cap_broadcast_adv_stop(source_id)
     btp.cap_broadcast_source_stop(source_id)
+    time.sleep(10) # Allow some time for the stack to stop the broadcast
```
Thalley, line 566: The right solution would be to either implement an BTP event, or for the IUT to not send the response before the BIG is fully stopped. We did something similar for a BAP test. 

Adding a sleep in the AutoPTS code is, IMO, not the right solution.

Thalley (changes requested):

Thalley:
> yes, I understand that, but I wonder why that command is not synchronous (ie it does return current_settings which suggests that state changes), how long does it take to disable broadcasting?

Stopping a broadcast source is not a synchronous operations over HCI. We had a similar issue for a BAP test where we added a semaphore in Zephyr to wait with the response until the BIG was fully stopped.

autopts/wid/tmap.py
Thalley: The PR that fixed this for BAP: https://github.com/zephyrproject-rtos/zephyr/pull/79329

alexsven: Agree, see https://github.com/zephyrproject-rtos/zephyr/pull/86625 instead

alexsven:
Will be fixed by making cap_broadcast_source_stop synchronous instead: https://github.com/zephyrproject-rtos/zephyr/pull/86625

PR #1408, opened 2025-03-10
CAP: Implement CAP/CL/ADV/BV-04-C

alexsven:
- Add wid_421
- Add custom test case for CAP/CL/ADV/BV-04-C
- Test requires available contexts to be 0 in both adv and PACS

### DIFF
--- a/autopts/ptsprojects/zephyr/cap.py
+++ b/autopts/ptsprojects/zephyr/cap.py
@@ -15,7 +15,6 @@
 #
 
 """CAP test cases"""
-from enum import IntEnum, IntFlag
 import struct
 
 from autopts.pybtp import btp
@@ -25,17 +24,9 @@ from autopts.ptsprojects.testcase import TestFunc
 from autopts.ptsprojects.zephyr.cap_wid import cap_wid_hdl
 from autopts.ptsprojects.zephyr.ztestcase import ZTestCase, ZTestCaseSlave
 from autopts.pybtp.defs import PACS_AUDIO_CONTEXT_TYPE_CONVERSATIONAL, PACS_AUDIO_CONTEXT_TYPE_MEDIA
-from autopts.pybtp.types import Addr, AdType, Context
+from autopts.pybtp.types import Addr, Context
+from autopts.pybtp.btp.cap import announcements
 from autopts.utils import ResultWithFlag
-
-
-class Uuid(IntEnum):
-    ASCS = 0x184E
-    BASS = 0x184F
-    PACS = 0x1850
-    BAAS = 0x1852
-    CAS  = 0x1853
-
 
 def set_pixits(ptses):
     """Setup CAP profile PIXITS for workspace. Those values are used for test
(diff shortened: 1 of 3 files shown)


autopts/ptsprojects/zephyr/cap.py
```diff
@@ -173,12 +173,17 @@ def set_addr(addr):
...
     ]
 
     targeted_conditions = [
-        TestFunc(announcements, adv_data, rsp_data, True),
+        TestFunc(announcements, adv_data, rsp_data, True, SINK_CONTEXTS, SOURCE_CONTEXTS),
+        TestFunc(btp.gap_adv_ind_on, ad=adv_data, sd=rsp_data),
+    ]
+
+    idle_targeted_conditions = [
+        TestFunc(announcements, adv_data, rsp_data, True, 0, 0),
```
Thalley, line 186: Did you consider the approach of updating the advertising data in `hdl_wid_421` instead of setting it as a pre-condition?

Thalley, line 186:
```suggestion
        TestFunc(announcements, adv_data, rsp_data, True, Context.PROHIBITED, Context.PROHIBITED),
```

The fact that `0 == Context.PROHIBITED` makes the new requirement from CAP (1.0.1) even worse...

alexsven: Yes, the problem is that the wid is platform agnostic, whereas the format of the ad probably isn't. I ran into that issue when I tried to import the `announcement` function from `ptsproject.zephyr.cap`

alexsven: I deliberately refrained from using the define of zero here because that would look worse... :P

Thalley: :D 

Ideally we'd use some defined value here, but I'm not happy about introducing a `Context.NONE` either, as that is not a BT SIG defined value... 

In hindsight, they should have defined these to use `NONE = 0xFFFF` or something, but alas...

Thalley: The advertising data should be the same as what BT SIG defines, so should be platform agnostic. 

@sjanc may have some input here

autopts/pybtp/btp/cap.py
```diff
@@ -17,12 +17,21 @@
...
 from autopts.pybtp.btp.gap import __gap_current_settings_update
-from autopts.pybtp.types import addr2btp_ba, BTPError
+from autopts.pybtp.types import addr2btp_ba, BTPError, AdType
+
+class Uuid(IntEnum):
+    ASCS = 0x184E
+    BASS = 0x184F
+    PACS = 0x1850
+    BAAS = 0x1852
+    CAS  = 0x1853
```
Thalley, line 34: This is OK for now since it lived in another CAP file before, but ideally these would be moved to https://github.com/auto-pts/auto-pts/blob/master/autopts/pybtp/types.py but I see that file already has a string version of these... 

So we should probably defined all the UUIDs in a way that we can get both a string and integer value, but that's a separate topic

Thalley (approved):

sjanc:
#AutoPTS run zephyr CAP/CL/ADV/BV-04-C

codecoup-tester:
Scheduled PR https://github.com/auto-pts/auto-pts/pull/1408#issuecomment-2721378479, board: nrf53, estimated start time: 15:04:40, test case count: 1, estimated duration: 0:10:28<details><summary>Test cases to be run</summary>CAP/CL/ADV/BV-04-C</details>

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (1)</summary>CAP     CAP/CL/ADV/BV-04-C              PASS</details>

sjanc (approved):

PR #1412, opened 2025-03-14
doc/btp_gap: add filter accept list flag to discovery

jrintaha:
Add "use filter accept list" flag to Start Discovery.

Previously the BTP specification did not really say whether or not the filter accept list should be used for scan results.

This caused some confusion at least on our end, as we expected the accept list to apply for scanning as well. Now the application can explicitly say whether it wants to use the filtering or not. Any existing implementations that don't parse this flag will continue to behave the same way they used to, so this shouldn't cause backwards compatibility problems.

### DIFF
--- a/doc/btp_gap.txt
+++ b/doc/btp_gap.txt
@@ -202,6 +202,7 @@ Commands and responses:
                         3 = Use active scan type
                         4 = Use observation procedure
                         5 = Use own ID address
+                        6 = Use filter accept list
 
                 This command is used to start discovery.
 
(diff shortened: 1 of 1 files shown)


Thalley (approved):
Not sure what the correct approach is, but we should either add it here or follow up with another PR that adds this to the defined flags:
```
GAP_DISCOVERY_FLAG_LE = 0x01
GAP_DISCOVERY_FLAG_BREDR = 0x02
GAP_DISCOVERY_FLAG_LIMITED = 0x04
GAP_DISCOVERY_FLAG_LE_ACTIVE_SCAN = 0x08
GAP_DISCOVERY_FLAG_LE_OBSERVE = 0x10
```

Looks like it is also missing bit 5

sjanc (approved):

PR #1413, opened 2025-03-14
Add bot conf overlay for BAP/BSRC/SCC/BV-38-C

alexsven:
- Due to an encryption bug in ZLL this needs to be added only for this test

### DIFF
--- a/autopts/bot/iut_config/zephyr.py
+++ b/autopts/bot/iut_config/zephyr.py
@@ -121,6 +121,15 @@ iut_config = {
         ]
     },
 
+    "multiple_iso_adv_sets.conf": {
+        "overlay": {
+            'CONFIG_BT_CTLR_ADV_ISO_SET': '2',
+	},
+	"test_cases": [
+	    'BAP/BSRC/SCC/BV-38-C',
+	]
+    },
+
     "csip_privacy.conf": {
         "pre_overlay": ["overlay-le-audio.conf", "privacy.conf"],
         "test_cases": [


alexsven:
When fixing the regression for PBP I discovered that ZLL would crash if encryption is set for a broadcast source if CONFIG_BT_CTLR_ADV_ISO_SET=2. This config is only needed for BAP/BSRC/SCC/BV-38-C, so as a temporary solution it can be set specifically for that test.

autopts/bot/iut_config/zephyr.py
```diff
@@ -121,6 +121,15 @@
...
     },
 
+    "multiple_iso_adv_sets.conf": {
+        "overlay": {
+            'CONFIG_BT_CTLR_ADV_ISO_SET': '2',
+	},
+	"test_cases": [
+	    'BAP/BSRC/SCC/BV-38-C',
+	]
+    },
```
Thalley, line 131: Is the plan to keep this as a long-term solution, or temporary? If long term, then I suggest that we set all the Kconfig options for the test here, and not just this one), so keep the hci_ipc_cpunet.conf file as small as possible for the the other tests. 

If a temporary solution until the bug in the ZLL is fixed, we can keep this as is

alexsven: Hard to say. So far I've only encountered this one test that requires two adv sets, but there might be others either in the future or simply not in the ICS. I would say the best solution would be to fix the bug in ZLL and put it back in the hci_ipc_cpunet.conf, so that all tests are using the same controller config

Thalley: Will let @sjanc determine how to best approach this :)

alexsven: Just realized that this needs to be set in the net-core and not app-core. This would set it in the app-core, right? @sjanc

sjanc: yeap, this affects only bttester app build (ie not netcore)
(overlays are generally meant for tuning IUT behavior to match test requirements, not to workaround bugs in IUT)

I think the right solution would be to create issue for this in zephyr and point that in errata/zephyr.yml so that it is listed test results

Thalley:
The changes here are also applied by default in https://github.com/zephyrproject-rtos/zephyr/pull/88300. 

@alexsven did we ever create a Zephyr GH issue for the ZLL where we saw the issue with encrypted broadcast and with `CONFIG_BT_CTLR_ADV_ISO_SET=2`?

alexsven:
> The changes here are also applied by default in [zephyrproject-rtos/zephyr#88300](https://github.com/zephyrproject-rtos/zephyr/pull/88300).
> 
> @alexsven did we ever create a Zephyr GH issue for the ZLL where we saw the issue with encrypted broadcast and with `CONFIG_BT_CTLR_ADV_ISO_SET=2`?

If we did, I can't seem to find it, so I assume not. I had it on my todo-list, but it drowned in release work for NCS 3.0.0

PR #1423, opened 2025-03-25
autopts: doc: fix documentation

JakubMarciniak93:
This Pull Request improves the documentation related to Opcode 0x01 – Read Supported Commands.
Change include correctly value for Controller Index

### DIFF
--- a/doc/btp_aics.txt
+++ b/doc/btp_aics.txt
@@ -5,9 +5,9 @@ Commands and responses:
 
         Opcode 0x00 - Error response
 
-        Opcode 0x01 - Read supported commands
+        Opcode 0x01 - Read Supported Commands command/response
 
-                Controller Index:       <controller id>
+                Controller Index:       <non-controller>
                 Command parameters:     <none>
                 Response parameters:    <supported commands> (variable)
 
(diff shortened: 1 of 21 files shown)


doc/btp_aics.txt
```diff
@@ -5,9 +5,9 @@ Commands and responses:
 
         Opcode 0x00 - Error response
 
-        Opcode 0x01 - Read supported commands
+        Opcode 0x01 - Read Supported Commands command/response
 
-                Controller Index:       <controller id>
```
Thalley, line 10: A general comment: While I'm not 100% sure how the `controller id` is used in AutoPTS, if you can send commands to different controllers, then there's a decent chance that each of those have different supported commands, right? 

Can you elaborate on why it's being changed to `non-controller`?

sjanc: this is overlook from my side when doing BTP docs review..

Those should all be non-controller as those were (are) meant to check what is supported by service in general (ie if service can understand command in first place).  If there are discrepancies due to multiple controllers (note that this is leftover from very early times of autopts when things were heavily influenced by Linux BlueZ:P) supporting different features than command may simply fail.

sjanc (approved):

PR #1425, opened 2025-03-26
Make sure CSIP discovery events are unique

Frodevan:
CSIP sirk events were not cleared after discovery, causing some test-cases to incorrectly assume all lower-testers were discovered when in fact they were not. This commit fixes this issue.


autopts/wid/csip.py
Thalley, line 1: What about the one in `hdl_wid_20107`?

Thalley: @Frodevan can you please expand the commit message that this last one with `remove=False` is required for `CSIP/CL/SP/BV-06-C` because that test requires accessing the SIRK later?

Frodevan: Allright

Frodevan: Done

Thalley:
#AutoPTS run zephyr nrf53 CSIP/CL

sjanc (approved):

PR #1427, opened 2025-03-27
boards: Remove uses of overlay-le-audio-ctlr

Thalley:
The overlay doesn't exist in Zephyr. If any specific overlay file should be used, e.g. the overlay-le-audio.conf, then that should be supplied as the `conf_file` without this extra `audio` check.


sjanc (approved):
yeap, sorry, somehow I've miss this...

PR #1429, opened 2025-03-28
Add GATT service initialization for L2CAP test cases.

szymon-czapracki:
L2CAP needs GATT service initialized for some tests.


szymon-czapracki:
> I assume this is for L2CAP/TIM/BV-03-C, so lets just make it only for that test

Done

sjanc (approved):

Issue #1430, opened 2025-03-28
Add Port Parameter to BTP Socket Initialization and it should default zero


jignesh-kumar:
https://github.com/auto-pts/auto-pts/blob/master/autopts/pybtp/iutctl_common.py#L225C51-L225C52

PR #1431, opened 2025-03-28
Port should be pass as argument and it’s should default zero in class BTPSocketSrv(BTPSocket) #1430

jignesh-kumar:


autopts/pybtp/iutctl_common.py
```diff
@@ -215,14 +215,14 @@ def __init__(self, log_dir=None):
...
-    def open(self, addres=BTP_ADDRESS):
+    def open(self, addres=BTP_ADDRESS, port=0):
         """Open BTP socket for IUT"""
         if os.path.exists(addres):
             os.remove(addres)
 
         if sys.platform == "win32":
             self.sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            self.sock.bind((socket.gethostname(), 0))
+            self.sock.bind((socket.gethostname(), port))
```
Thalley, line 225: Will there be a follow-up PR that actually uses this?

jignesh-kumar: Nope

Thalley: Then why are you requesting this change if it has no uses? :s

jignesh-kumar: I want to use this repo to run the pts test on my device. My IUT communication pts client using TCP Socate where no required Socat.exe. Hence these changes required non-zero ports so.

jignesh-kumar: @Thalley 
More inputs required on this?

Thalley: No, I'm overall OK with this change, but the issue and/or the commit message should have _why_ this is needed, but will defer to @sjanc to determine whether to approve this

jignesh-kumar: @sjanc
Please review this

jignesh-kumar:
@sjanc 
My IUT device uses the ADB console instead of UART, making socat.exe unsuitable for my setup. However, I need socat communication with a specific port, so I require an adjustment or an alternative approach to ensure compatibility.

sjanc:
@jignesh-kumar alright,  just put that info in commit message and we are good to go

jignesh-kumar:
> @jignesh-kumar alright, just put that info in commit message and we are good to go

Done

jignesh-kumar:
@sjanc can you review now?

sjanc (approved):

PR #1432, opened 2025-03-31
zephyr: mesh: PNID advertisement must stop immediately

m-alperen-sener:
Private Node Identity advertisements must stop immediately when ordered by PTS.

To do so;
- Adding enable parameter to BTP_MESH_CMD_PROXY_PRIVATE_IDENTITY to enable or disable the proxy private identity advertisements.
- Updated documentation for the command.


sjanc (approved):
I guess we can break API it since this is zephyr is only mesh mode service user for now

m-alperen-sener:
> I guess we can break API it since this is zephyr is only mesh mode service user for now

I did not understand this point. Zephyr changes here:
https://github.com/zephyrproject-rtos/zephyr/pull/87891

sjanc:
what I mean is that you should consider BTP a "stable" protocol and try not making changes that breaks if both sides are not updated

PR #1437, opened 2025-04-03
Fix `hdl_wid_52` decode error

theob-pro:
The `value_read` variable is converted to a string in the if branch but not in the else one where it's still `bytes`. The `decode` method later called on `value_read` is trying to convert it to a string. But in the case we were in the if branch, `value_read` is already a string. This lead to an `AttributeError` because the `decode` method doesn't exist for string.

Introduce `value_read_str` variable to fix the issue. That variable will be used to convert `value_read` in each branch to a string.

### DIFF
--- a/autopts/wid/gatt.py
+++ b/autopts/wid/gatt.py
@@ -704,15 +704,15 @@ def hdl_wid_52(params: WIDParams):
 
     if attr.uuid == UUID.CEP:
         (value_read,) = struct.unpack("<H", attr.value)
-        value_read = '{0:04x}'.format(value_read)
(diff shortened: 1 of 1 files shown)


sjanc:
#AutoPTS run zephyr GATT/SR/GAR/BV-06-C

codecoup-tester:
Scheduled PR https://github.com/auto-pts/auto-pts/pull/1437#issuecomment-2777745862, board: nrf52, estimated start time: 09:15:29, test case count: 1, estimated duration: 0:10:24<details><summary>Test cases to be run</summary>GATT/SR/GAR/BV-06-C</details>

sjanc (approved):

PR #1438, opened 2025-04-04
[DRAFT] Poetry and pep8

JakubMarciniak93:


tools.py
```diff
@@ -0,0 +1,31 @@
...
+                command,
+                stdout=log_file,
+                stderr=subprocess.STDOUT,
+                text=True
+            )
+        return result.returncode
+
+    ruff_code = run_tool("Ruff", ["poetry", "run", "ruff", "check", ".", "--fix"], logs_dir / "ruff.log")
+    isort_code = run_tool("isort", ["poetry", "run", "isort", "."], logs_dir / "isort.log")
+    black_code = run_tool("Black", ["poetry", "run", "black", "."], logs_dir / "black.log")
```
Thalley, line 21: Why both `ruff` and `black`? Don't we risk cases where these 2 disagree?

README.md
```diff
@@ -1,3 +1,119 @@
+# Getting Started
+Follow the steps below to set up the project, install dependencies, 
+and enable automatic code formatting and linting.
+
+### Install Poetry
+If you don’t have Poetry installed:
```
Thalley, line 6: What's the rationale behind using Poetry instead of pip for this?

tools.py
JakubMarciniak93: I'm looking for the best solution, for now it's a draft. It's not the final version. Thanks for your attention

Thalley: If you are looking for input, I have a strong preference for `ruff`. Black does a good job, but it's quite slow and I've had many issues with formatting things with Black. `ruff` is extremely fast, and I've no issues with it yet

PR #1442, opened 2025-04-09
cron: pr_job: Add startup_fail_max_count option

mkasenberg:
If a PR causes the test run startup to fail and no ERROR_TXT_FILE is generated, cron will be restarted indefinitely and block other jobs.

### DIFF
--- a/tools/cron/common.py
+++ b/tools/cron/common.py
@@ -590,6 +590,7 @@ def _run_test(config):
     test_cases_completed = False
     backup = config['auto_pts'].get('use_backup', False)
     timeguard = config['cron']['test_run_timeguard']
+    startup_fail_count = config['cron'].get('startup_fail_max_count', 2)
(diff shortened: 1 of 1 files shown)


tools/cron/common.py
```diff
@@ -626,11 +627,17 @@ def _run_test(config):
 
         if not test_cases_completed and not os.path.exists(results_file_path):
             if timedelta(seconds=current_time - last_check_time) > timedelta(seconds=timeguard):
+                if startup_fail_count == 0:
+                    log(f"Test run has not been started on time. No more retries...")
```
Thalley, line 631:
```suggestion
                    log("Test run has not been started on time. No more retries...")
```

The `f` is superfluous if not doing any formatting

sjanc (approved):

Thalley (approved):

PR #1452, opened 2025-05-09
log: Add "|" symbol to identify header in raw data

szymon-czapracki:
Add | symbol between packet header and data for
BTP command. This could prove useful, as user
can see command data without counting bytes.


szymon-czapracki:
obraz

szymon-czapracki:
Packet structure (from overview.txt file) for reference.
obraz

sjanc (approved):

PR #1453, opened 2025-05-14
Add new command for scan pareamete and device found flags

petrip-slab:
Add new command that can be used for setting scan interval and scan window.

### DIFF
--- a/doc/btp_gap.txt
+++ b/doc/btp_gap.txt
@@ -695,6 +695,29 @@ Commands and responses:
 
                 In case of an error, the error response will be returned.
 
+        Opcode 0x31 - Set Discovery Parameters command/response
+
+                Controller Index:       <controller id>
+                Command parameters:     Interval (2 octets)
+                                        Window (2 octets)
+                                        Flags (1 octet)
+                Return Parameters:      <none>
+
+                This command is used to set scanning parameters for discovery.
+                Interval: The time interval when the device starts its last scan
+                until it begins the subsequent scan.
+                Valid range is from 0x0004 to 0xffff.
+                Time = 0.625 ms x Interval
+
+                Window: Duration of single scan, must less than or equal
+                to the Interval.
+                Valid range is from 0x0004 to 0xffff.
+                Time = 0.625 ms x Interval
+
+                Default values are implementation specific.
+
+                In case of an error, the error response will be returned.
+
 Events:
         Opcode 0x80 - New Settings event
 


doc/btp_gap.txt
```diff
@@ -695,6 +695,29 @@ Commands and responses:
 
                 In case of an error, the error response will be returned.
 
+        Opcode 0x31 - Set Discovery Parameters command/response
```
sjanc, line 698: lets start filling gaps between 0x2a and 0x30 ;-)
so 0x2b should be used for this

Also, I'd name it "Set LE Discovery Parameters" as those are LE only params

doc/btp_gap.txt
```diff
@@ -695,6 +695,29 @@ Commands and responses:
...
+                                        Flags (1 octet)
+                Return Parameters:      <none>
+
+                This command is used to set scanning parameters for discovery.
+                Interval: The time interval when the device starts its last scan
+                until it begins the subsequent scan.
+                Valid range is from 0x0004 to 0xffff.
+                Time = 0.625 ms x Interval
+
+                Window: Duration of single scan, must less than or equal
```
szymon-czapracki, line 712:
```suggestion
                Window: Duration of single scan, must be less than or equal
```

doc/btp_gap.txt
```diff
@@ -695,6 +695,29 @@ Commands and responses:
 
                 In case of an error, the error response will be returned.
 
+        Opcode 0x31 - Set Discovery Parameters command/response
+
+                Controller Index:       <controller id>
+                Command parameters:     Interval (2 octets)
+                                        Window (2 octets)
+                                        Flags (1 octet)
```
Thalley, line 703: What are the possible flags?

doc/btp_gap.txt
```diff
@@ -695,6 +695,29 @@ Commands and responses:
...
+                Return Parameters:      <none>
+
+                This command is used to set scanning parameters for discovery.
+                Interval: The time interval when the device starts its last scan
+                until it begins the subsequent scan.
+                Valid range is from 0x0004 to 0xffff.
+                Time = 0.625 ms x Interval
```
Thalley, line 710: This seems like it's leaning a lot on the HCI spec :D I'm fine with that, but @sjanc have previously mentioned a dislike for being too HCI centric in BTP

petrip-slab: I did thing putting bare milliseconds here. easier to understand BUT then it could lead confusion. Like 2 milliseconds but it advertising got about 1.8 miliseconds. 

But  I am fine with both.

doc/btp_gap.txt
petrip-slab: Sorry about late answer I will add the missing explantion. One bit 0x01 would active scan and 0x00 would passive 
Scan

I could have called in mode and if preferred I can change the name but just in case someone wants to add new parameters. And mode might be better still.

doc/btp_gap.txt
petrip-slab: done

doc/btp_gap.txt
petrip-slab: done

doc/btp_gap.txt
sjanc: active/passive is already controlled  in start discovery,  this is just for future in case some additional tuning is needed (to avoid need to V2 command)

doc/btp_gap.txt
sjanc: using HCI units in hosts for this is non uncommon so I'm fine with this case

doc/btp_gap.txt
```diff
@@ -695,6 +695,33 @@ Commands and responses:
...
+
+                Window: Duration of single scan, must be less than or equal
+                to the Interval.
+                Valid range is from 0x0004 to 0xffff.
+                Time = 0.625 ms x Interval
+
+                Possible values for the Flags parameter are a bit-wise or
+                of the following bits:
+                        0 = Active Scan
```
sjanc, line 719: this is already controlled in start discovery

petrip-slab: Ok I will change comment to reserved for future use.

doc/btp_gap.txt
```diff
@@ -695,6 +695,32 @@ Commands and responses:
 
                 In case of an error, the error response will be returned.
 
+        Opcode 0x2b - Set LE Discovery Parameters command/response
```
Thalley: One issue I've had with BTP for a while, is that the descriptions are not very explicit. 

This command is defined as `This command is used to set scanning parameters for discovery.` but what does that mean? It seems to lean a lot on the `HCI_LE_Set_Scan_Parameters` command, except that it only does a subset of the parameters. 

Should the IUT send the HCI command if it receives this? Should it just store the parameters until another command? The IUT may not be able to complete the former if already scanning, but could do the latter regardless of the scanning state. 

Without more explicit documentation on what the IUT is supposed to do, we end up with a protocol that is so implementation specific, that it becomes hard to consider it a protocol for multiple IUTs.

petrip-slab: I wonder how we could make it specific. We can add a line to documenetion like "for all future discovery procedures"

As for all parameters yes. For instance privacy is closely related to scanning but not only to it and has already command. And not all IUT's implement it. For instance Zephyr   does not have set privacy as it controlled build time.

As ACTIVE/PASSIVE BTP has it already in command that actually starts the procedure.
I did consider making new start discovery with parameters but then we would have two of them. and still not perhaps complete

And mixing whole white list and filtering policies to simple things interval/window seem quite. Is there anything other scanning paremaete I missed?

petrip-slab: some parts are implementation specific anyway. and yes I did not duplicate the active/passive as removal of it was requested. Only other relevant addition would be the 'filter policy'  field. But not all implementation support it so implementations would differ on this as well 

Obviously can the one byte field but goal of similar implementation is probably not reachable. Not on just this issue.

sjanc: I think things like PHY and filtering (other than what is already possible with start discovery command) can be handled by flags here.   My suggestion would be to also make this 'stateless', ie fail if discovery is already ongoing, and maybe add note (in discovery command) that if this command is never called IUT specific defaults are used.

doc/btp_gap.txt
```diff
@@ -695,6 +695,42 @@ Commands and responses:
...
+                Time = 0.625 ms x Interval
+
+                Window: Duration of single scan, must be less than or equal
+                to the Interval.
+                Valid range is from 0x0004 to 0xffff.
+                Time = 0.625 ms x Interval
+
+                Phy: Select Phy to be used in the discovery. 
+                        0: 1 mbps
+                        1: 2 2mbps
```
sjanc: this is for primary advertising so 1M and Coded only

doc/btp_gap.txt
```diff
@@ -695,6 +695,42 @@ Commands and responses:
...
+                        3: Coded Phy
+
+		Flags: Possible values for the Flags parameter are a bit-wise OR.
```
sjanc: nitpick: use only spaces for indentation

doc/btp_gap.txt
```diff
@@ -695,6 +695,42 @@ Commands and responses:
...
+
+                Window: Duration of single scan, must be less than or equal
+                to the Interval.
+                Valid range is from 0x0004 to 0xffff.
+                Time = 0.625 ms x Interval
+
+                Phy: Select Phy to be used in the discovery. 
+                        0: 1 mbps
+                        1: 2 2mbps
+                        3: Coded Phy
```
Thalley: The HCI spec allows a device to scan for either 1M or Codec, or both. This does not seem to match this command

Image

doc/btp_gap.txt
petrip-slab: Fixed

doc/btp_gap.txt
petrip-slab: Done
And rebasing caused some empty lines before the 'Events:' removed those as well

doc/btp_gap.txt
Thalley: @petrip-slab shouldn't this be a bitfield? 

Or is the intention that this command will be called twice if wanting to use both PHYs?

petrip-slab: Yes you are correct I will update it

doc/btp_gap.txt
```diff
@@ -802,6 +802,44 @@ Commands and responses:
...
+                to the Interval.
+                Valid range is from 0x0004 to 0xffff.
+                Time = 0.625 ms x Window
+
+                Phy: Bitfield tp select Phys to be used in the discovery.
+
+                Possible values for the Flags parameter are a bit-wise OR
+                of the following bits: 
+                        0 = PHY 1M
+                        1 = PHY Coded
```
Thalley, line 833: Since this is now a bitfield, it will allow us to set the interval and window for both PHYs, but won't allow us to set different values for each PHY. 

Is that something we want this to support? The HCI spec allows for different values for the scan type (should that be part of this command as well, or is it always just "active"?), the interval and window

Image

sjanc: TBH I was thinking on PHY being either value or flag in flags (ie only coded flag) and that command would have to be called twice if one wants to set both 1M and coded)

petrip-slab: Thinking of implementation and such. having separate window and interval might be quite challenging. Because coded phy and 1M are same modulation and can be captured in same go having different scan window 

But both solution are fine. as @sjanc said if bitfield and two call to set both it up user to set intervals and windows that bot can served even device cannot get them from same reception.

Thalley: If we put the PHY in the flags, we just need to make sure to mark them as mutually exclusive. 

Either way really works, since this won't trigger an HCI command, so the IUT just need to store the data for each PHY (or reject the 2nd request if only 1 PHY can be used).

petrip-slab: Then I would proposet to have it as originally a field with two values and IUT stores them and given error message if unsupported is given. Either by not supporting simultaneous or not supporting (the coded) at all.

Thalley: Might be a bit tricky with flags actually; if we do it over 2 commands, then we should make sure that the flags are useful for both PHYs (although not a strict requirement), and that each PHY may have a different set of flags than the other (although not a strict requirement)

petrip-slab: True, no one want intepret same field in two ways in same command.  But flags specific to PHY is not that serious as flag having two interpretations.

Thalley: So 2 solutions: 
1) PHY is a value:

- Command will be called for each PHY
- Flags may be the same for both types of PHY, but can be documented if some flags are specific to a PHY
- IUT will need to store parameters for each PHY if supported
- Command is simpler

2) PHY is a bitfield

- Command is called once
- Either need to support interval and window (and type?) per PHY in the command, or to always use the same for both
- Flags are specific to the command, and not the PHY (although PHY specific flags can still be added if properly documented)
- IUT will need to store parameters for each PHY if supported
- Command is slightly more complex

Either approach really works IMO, and any issues with flags can be handled with proper documentation. If we decide on 1), we may want to use 16-bits for the flags if 8 bits are too few if we have PHY specific flags

petrip-slab: Option 2 I would prefer both being the same. simpler command and also some devices are capable scan both at the same time when same parameters are pretty much required.

But I have no strong preferences here. All solution provide sufficient functionality for all use cases I can think of.

Thalley: > Option 2 I would prefer both being the same. simpler command and also some devices are capable scan both at the same time when same parameters are pretty much required.

For this, would you have 2 intervals and 2 windows (or perhaps keep it as an array similar to the HCI Command in case more PHYs are added in the future), or always use the same for both PHYs?

petrip-slab: I would thing same parameters for both. Thinking of future option 1 is more flexiple? I mean implementation could support easily expansion phy's allow implementation to force parameters to be same or not. protocol would not limit the IUT has extra or missed some common. And in typical case having just on PHY activated no need to feed the extra parameters 

Allow having different or same parameters without complicating the command definition . Just need document this somehow and increase flags to 16bit to avoid future versions.

Thalley: > I would thing same parameters for both. Thinking of future option 1 is more flexiple? I mean implementation could support easily expansion phy's allow implementation to force parameters to be same or not. protocol would not limit the IUT has extra or missed some common. And in typical case having just on PHY activated no need to feed the extra parameters

It does limit the BTP command compared to the HCI command though. 

@sjanc what are you thoughts?

petrip-slab: @sjanc So what is your opinion which way I should do it? I prefer one set of parameters as it more flexiple and one call will server most use cases.

sjanc: Hi,

Sorry for late reply, miss this comment.

I'm fine with current approach, if one needs to have different params per PHY he may call this twice with only single PHY flag set

petrip-slab: Okay since no further comment exist I squashed the commit and verified the it will nicely rebase on top of 'master'

doc/btp_gap.txt
```diff
@@ -802,6 +802,44 @@ Commands and responses:
...
+
+                Flags: Possible values for the Flags parameter are a bit-wise OR.
+                Currently this field is reserved for future use and shall be set to 0.
```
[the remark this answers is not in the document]
petrip-slab: done

sjanc (approved):

PR #1458, opened 2025-05-16
mynewt/gap: Add dummy LE supported features

szymon-czapracki:
Add dummy value LE supported features to
GAP initialization.


szymon-czapracki:
Fix for failing GAP/ADV/BV-19-C test case in nimble.

sjanc (approved):

PR #1459, opened 2025-05-19
gatt_client: Add missing WIDs

szymon-czapracki:
Add missing WIDs.
WID 130 - deleting pairing information
WID 135 - writing to wid-specific handle

This is aimed for fixing GATT/CL/GAS/BV-03-C test case.


Thalley (changes requested):

szymon-czapracki:
Update - added 2 more WIDs for the almost the same purpose as in previous version of this PR

autopts/wid/gatt_client.py
```diff
@@ -1289,6 +1289,48 @@ def hdl_wid_109(params: WIDParams):
...
+    # First read the existing value in Client Supported Features.
+    btp.gatt_cl_read(btp.pts_addr_type_get(), btp.pts_addr_get(), handle)
+
+    stack = get_stack()
+    stack.gatt_cl.wait_for_verify_values()
+    client_support_features = btp.get_verify_values()
+
+    # Set Robust Caching bit in features.
+    robust_caching_bit = 1
+    value = int(client_support_features[1]) | robust_caching_bit
```
Thalley, line 1324: Let's try to avoid magic numbers
Suggest to define the `client_support_features` bits and use define values. 
Additionally, why is this only setting `client_support_features[1]` and not both `[0]` and `[1]`?

autopts/wid/gatt_client.py
```diff
@@ -1354,6 +1396,49 @@ def hdl_wid_150(params: WIDParams):
...
+
+    stack = get_stack()
+    stack.gatt_cl.wait_for_verify_values()
+    client_support_features = btp.get_verify_values()
+
+    # Set Robust Caching bit in features.
+    robust_caching_bit = 1
+    eatt_bit = 2
+
+    value = int(client_support_features[1]) | robust_caching_bit | eatt_bit
```
Thalley, line 1424: Ditto; the bits should be defined in e.g. `defs.py` or something, and used here, instead of being defined multiple times

autopts/wid/gatt_client.py
```diff
@@ -1289,6 +1289,48 @@ def hdl_wid_109(params: WIDParams):
...
+    stack = get_stack()
+    stack.gatt_cl.wait_for_verify_values()
+    client_support_features = btp.get_verify_values()
+
+    # Set Robust Caching bit in features.
+    robust_caching_bit = 1
+    value = int(client_support_features[1]) | robust_caching_bit
+
+    btp.gatt_cl_write(btp.pts_addr_type_get(), btp.pts_addr_get(), handle,
+                      f'{value:02x}', 1)
```
Thalley, line 1327:
```suggestion
    btp.gatt_cl_write(btp.pts_addr_type_get(), btp.pts_addr_get(), handle,
                      f'{value:02x}')
```

Superfluous

autopts/wid/gatt_client.py
```diff
@@ -1354,6 +1396,49 @@ def hdl_wid_150(params: WIDParams):
...
+    client_support_features = btp.get_verify_values()
+
+    # Set Robust Caching bit in features.
+    robust_caching_bit = 1
+    eatt_bit = 2
+
+    value = int(client_support_features[1]) | robust_caching_bit | eatt_bit
+
+    btp.gatt_cl_write(btp.pts_addr_type_get(), btp.pts_addr_get(), handle,
+                      f'{value:02x}', 1)
```
Thalley, line 1427:
```suggestion
    btp.gatt_cl_write(btp.pts_addr_type_get(), btp.pts_addr_get(), handle,
                      f'{value:02x}')
```
Superfluous

Thalley (changes requested):

szymon-czapracki:
UPDATE:

After an internal design review and offline discussion, we have concluded that
writing to the Client Supported Features characteristic must occur during the 
initial setup phase of the BLE connection.

The purpose of this write is to explicitly inform the GATT server which optional 
features the client supports — such as Robust Caching (bit 0), Enhanced ATT (bit 1), etc. 
This allows the server to optimize behavior based on the client’s capabilities.

Delaying this write or performing it on-demand (i.e., later in the session) may lead 
to undefined behavior or compatibility issues.

autopts/wid/gatt_client.py
szymon-czapracki: ditto

autopts/wid/gatt_client.py
szymon-czapracki: ditto

autopts/wid/gatt_client.py
```diff
@@ -1289,6 +1289,27 @@ def hdl_wid_109(params: WIDParams):
...
+
+
+def hdl_wid_135(params: WIDParams):
+    """
+    Please write to client support feature handle = 'XXXX'O to enable Robust Caching.
+    Discover all characteristics if needed.
+    """
+
+    # This write should be done as part of the initial GATT setup phase,
+    # not later on demand.
```
Thalley, line 1308:
```suggestion
    # This write is done as part of the initial GATT setup phase
```

autopts/wid/gatt_client.py
```diff
@@ -1355,6 +1376,27 @@ def hdl_wid_150(params: WIDParams):
...
 
 
+def hdl_wid_153(params: WIDParams):
+    """
+    Please write to client support feature handle = 'XXXX'O to enable EATT and Robust Caching.
+    Discover all characteristics if needed.
+    """
+
+    # This write should be done as part of the initial GATT setup phase,
+    # not later on demand.
```
Thalley, line 1386:
```suggestion
    # This write is done as part of the initial GATT setup phase
```

autopts/wid/gatt_client.py
```diff
@@ -1355,6 +1376,27 @@ def hdl_wid_150(params: WIDParams):
     return True
 
 
+def hdl_wid_153(params: WIDParams):
+    """
+    Please write to client support feature handle = 'XXXX'O to enable EATT and Robust Caching.
+    Discover all characteristics if needed.
+    """
```
Thalley, line 1383: Since this is supposed to be done during initialization, is that being done somewhere?

I cannot find the code that enables EATT and Robust caching, or is it assumed that the IUT will do this without instructions from autopts?

autopts/wid/gatt_client.py
szymon-czapracki: Done

autopts/wid/gatt_client.py
szymon-czapracki: Done

autopts/wid/gatt_client.py
szymon-czapracki: IUT will do this without instructions from autopts.

Thalley: What is the rationale for expecting the IUT to do this autonomously?

The GATT test suite states:
> The Upper Tester orders the IUT to enable the Robust Caching feature.

Shouldn't the AutoPTS client here, as the Upper Tester, instruct the IUT to enable the feature?

szymon-czapracki: The IUT is acting as the GATT client in this scenario. During the initial setup phase, we correctly inform the PTS of our supported features, with the Robust Caching bit explicitly set to 0—indicating that this feature is not supported by the IUT. (It is unclear why this test case is enabled given our lack of support for robust caching, but that is a separate matter.)

However, we have observed that the PTS subsequently requests the IUT to write to the PTS’s Client Supported Features characteristic, despite the IUT not supporting the robust caching feature. In our understanding, this behavior does not fully align with the intent of the Bluetooth Core Specification.

image

szymon-czapracki: I've submitted a request in PTS support - will update here when I'll get any feedback.
I agree that we should follow the instructions that WID provides us with - I'll revert those changes to previous state minding your previous comments about defining and not using magic numbers.

Thalley: Can you please add reference to the created erratum in the code, so that we can search for them once they are fixed? 

Does the erratum prevent us from passing any tests? (i.e. should the erratum be added to the errata list?)

szymon-czapracki: I've checked in weekly runs and this test is disabled for Zephyr, it's enabled for mynewt-nimble but I think that it will be disabled in the near future. There is one more test for mynewt-nimble (GATT/CL/GPM/BV-11-C) which is affected by this, but It will also be disabled IMHO.
Anyway as the request is active and tests are still in weekly runs I'll expand the errata list with it.

autopts/wid/gatt_client.py
```diff
@@ -1289,6 +1290,49 @@ def hdl_wid_109(params: WIDParams):
...
+    btp.gatt_cl_read(btp.pts_addr_type_get(), btp.pts_addr_get(), handle)
+    stack = get_stack()
+    stack.gatt_cl.wait_for_verify_values()
+    client_support_features = int(btp.get_verify_values()[1])
+
+    # Set Robust Caching bit in features.
+    value = (
+        client_support_features
+        | defs.GATT_CLIENT_SUPPORTED_FEATURES_ROBUST_CACHING
+    )
```
Thalley, line 1326: `get_verify_values` isn't documented, so I'm not sure what it does, but is there a reason why `GATT_CLIENT_SUPPORTED_FEATURES_ROBUST_CACHING` isn't added to that?

Thalley: Thanks :) 

I think I'm missing some overall design knowledge here, so I don't think I can approve this PR (simply because I don't know if it's correctly implemented) - So I defer to someone else to approve :)

PR #1465, opened 2025-05-23
doc: Update GATT Server BTP service

sjanc:
This adds command for explicit database initialization.

### DIFF
--- a/doc/btp_gatt_server.txt
+++ b/doc/btp_gatt_server.txt
@@ -20,9 +20,25 @@ Commands and responses:
 
                 In case of an error, the error response will be returned.
 
-        Opcode 0x02 - Get Attributes
+        Opcode 0x02 - Initialize GATT database
 
-                Controller Index:       <controller id>
+                Controller Index:       <non-controller>
+                Command parameters:     Database_ID (1 octet)
+                                        Flags (4 octets)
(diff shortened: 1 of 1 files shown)


Thalley (commented):
I'm a bit confused here. The commit message states that it adds a new command to initialize a GATT database, but the change seems to modify several other commands as well (changing the opcode). 

Additionally, if the database is IUT specific, how would the BTP client know which database_ID to provide? This looks more like a RPC than a protocol design

sjanc:
this service is unused yet so no issues with changing it,  the idea is to allow IUT option to select alternative database is specific test requires it  (ID will be platform specific as this is selected in platform specific code anyway)

Thalley:
> this service is unused yet so no issues with changing it, the idea is to allow IUT option to select alternative database is specific test requires it (ID will be platform specific as this is selected in platform specific code anyway)

Fair point :) Suggest to add in the commit message why it was added before the other commands, thus causing all of those to get new opcodes. 

Do we have any way to mark this service as WIP or experiemental or something? You say that it isn't in use, but you are defining a protocol, and you have no idea who uses this protocol outside of this repository :)

Thalley:
Either way, you are the maintainer, and I'm not blocking this change :)

PR #1475, opened 2025-06-03
Improve btmon logging via rtt

szymon-czapracki:
Improve general btmon funcionality & stability.

### DIFF
--- a/autopts/rtt.py
+++ b/autopts/rtt.py
@@ -197,7 +197,7 @@ class BTMON:
 
             self.rtt_reader.start(buffer_name, device_core, debugger_snr, self._on_line_read_callback, (sock,))
         else:
-            cmd = f"btmon -C 130 -J {device_core},{debugger_snr} -w {log_filename} > {plain_log_filename}"
+            cmd = f"btmon -C 130 -J {device_core},{debugger_snr} -w {log_filename} | grep -v '^= bt:' > {plain_log_filename}"
             self.btmon_process = subprocess.Popen(cmd, cwd=log_filecwd,
                                                   shell=True,
                                                   stdout=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


Thalley (approved):

autopts/rtt.py
```diff
@@ -196,17 +197,19 @@ def start(self, buffer_name, log_file, device_core, debugger_snr):
...
-                                                  stderr=subprocess.PIPE)
+                                                  stderr=subprocess.PIPE,
+                                                  preexec_fn=os.setsid)
 
     def stop(self):
         log("%s.%s", self.__class__, self.stop.__name__)
         self.rtt_reader.stop()
 
         if self.btmon_process and self.btmon_process.poll() is None:
+            os.killpg(os.getpgid(self.btmon_process.pid), signal.SIGTERM)
```
mkasenberg: It seems that there is no os.killpg in Windows, so let's add a check:
```suggestion
            if sys.platform != 'win32':
                os.killpg(os.getpgid(self.btmon_process.pid), signal.SIGTERM)
```

mkasenberg (changes requested):

autopts/rtt.py
Thalley: There isn't a platform agnostic method?

mkasenberg: In the Windows + WSL2 setup the btmon is started differently and I personally have not encountered any hanging of btmon processes.

szymon-czapracki: Done

szymon-czapracki: > In the Windows + WSL2 setup the btmon is started differently and I personally have not encountered any hanging of btmon processes.

Maybe it's due to calling killall  at the beginning - this kills all hanging btmon processes.

mkasenberg (approved):

PR #1480, opened 2025-06-09
zephyr: Update workspace to PTS 8.9.1

sjanc:
This includes update to TCRL 2025-2 and sync ICS with corresponding Zephyr qualification workspace.


autopts/ptsprojects/zephyr/l2cap.py
Thalley, line 1: I assume that the removed IXIT values here no longer exists? 

Last I looked some IXIT values up, I do recall seeing quite a lot of values set by AutoPTS that weren't defined by the IXIT specs, but not sure if PTS may still use some that aren't defined?

sjanc: yeap, PTS 8.9.0 removed bunch of unused IXITs

(in long term I'll be working on getting rid of most of those set_pixit() calls,  most of them are set to default values anyway and are causing exceptions if PTS removes/renames any of those)

PR #1481, opened 2025-06-10
bot: Add no_copy parameter

piotrnarajowski:
This option can be used to skip creating and using copy of workspace file. This might be useful if you need to run tests in automation and later use this workspace file to generate report in PTS, or you need to repeat some test cases manually and    keep verdicts from automated test session.

### DIFF
--- a/autopts/bot/common.py
+++ b/autopts/bot/common.py
@@ -72,6 +72,10 @@ class BotCliParser(CliParser):
         self.add_argument('--simple', action='store_true',
                           help='Skip build and flash in bot mode.', default=False)
 
+        self.add_argument('--nc', dest='copy', action='store_false',
+                          help='Do not copy workspace, open original one. '
+                               'Warning: workspace file might be modified', default=True)
+
     def add_positional_args(self):
         self.add_argument("config_path", nargs='?', default='config.py',
                           help="Path to config.py to use for testing.")
@@ -125,6 +129,7 @@ class BotConfigArgs(Namespace):
         self.no_build = args.get('no_build', False)
         self.dongle_init_retry = args.get('dongle_init_retry', 5)
         self.build_env_cmd = args.get('build_env_cmd', None)
+        self.copy = args.get('copy', True)
 
         if self.ykush or self.active_hub_server:
             self.usb_replug_available = True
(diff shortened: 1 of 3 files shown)


autopts/ptscontrol.py
```diff
@@ -696,23 +696,27 @@ def open_workspace(self, workspace_path):
         # Workaround CASE0044114 PTS issue
         # Do not open original workspace file that can become broken by
         # TestCase. Instead use a copy of this file
```
Thalley, line 698: If this is still the case, isn't it "dangerous" to use this new option?

piotrnarajowski: It is and I probably should note this somewhere

Thalley: I assume there's a usecase for not doing the copy?

piotrnarajowski: Sure, please see updated commit message

autopts/bot/common.py
```diff
@@ -72,6 +72,10 @@ def __init__(self, **kwargs):
         self.add_argument('--simple', action='store_true',
                           help='Skip build and flash in bot mode.', default=False)
 
+        self.add_argument('--nc', dest='no_copy', action='store_true',
+                          help='Do not copy workspace, open original one. '
+                               'Warning: workspace file might be modified', default=False)
```
Thalley: There's not a right and wrong way, but what's the rationale for doing a `no_copy` argument that defaults to `False`, instead of a `copy` argument that defaults to `True`?

Thalley (approved):

autopts/bot/common.py
piotrnarajowski: There is no rationale and your proposed idea makes more sense, so I changed it

Thalley (approved):

Atm-Sara:
It caused the exception "Exception: <Fault 1: "<class 'TypeError'>:PyPTS.open_workspace() missing 1 required positional argument: 'copy'">" when running without the bot.

sjanc:
> It caused the exception "Exception: <Fault 1: "<class 'TypeError'>:PyPTS.open_workspace() missing 1 required positional argument: 'copy'">" when running without the bot.

I think this is what @szymon-czapracki was seeing locally too,  fix is pending

PR #1486, opened 2025-06-12
wid: Implement GATT WID 154

sjanc:
This WID requires context from previous MMIs and is used to pass size of characteristic value to LT.

It is used in following tests:
GATT/SR/GAW/BI-32-C
GATT/SR/GAW/BI-33-C
GATT/SR/GAW/BI-39-C
GATT/SR/GAW/BI-40-C

### DIFF
--- a/autopts/wid/gatt.py
+++ b/autopts/wid/gatt.py
@@ -1959,7 +1959,7 @@ def hdl_wid_151(_: WIDParams):
             stack.gatt.value_len = val_len
             return f'{handle:04x}'
 
-    return False
+    return 0
 
 
 def hdl_wid_152(_: WIDParams):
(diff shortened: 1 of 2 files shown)


autopts/wid/gatt.py
```diff
@@ -1992,11 +1997,23 @@ def hdl_wid_152(_: WIDParams):
...
 
+def hdl_wid_154(_: WIDParams):
+    # description: Enter size of handle value based on previous mmi requirement.
+
+    stack = get_stack()
+
+    if stack.gatt.value_len != 0:
+        return stack.gatt.value_len
+
+    return False
```
Thalley: Probably outside the scope of this PR, but do we really have functions that have dynamic return types? This seems to return either an `int` or `bool` :o

sjanc: nah,  we should not do that,  I guess False would be simply converted 0.  I'll simplify this  (and clean those few others that do same)

PR #1488, opened 2025-06-16
ptscontrol: Fix AttributeError exception

mkasenberg:
The restart_pts method can be called without args provided. Fix exception:
...
    if args.dongle_init_retry == 0:
AttributeError: 'NoneType' object has no attribute 'dongle_init_retry'


autopts/ptscontrol.py
```diff
@@ -513,6 +513,8 @@ def restart_pts(self, args=None):
 
         log("restart_pts")
         exception = 0
+        dongle_init_retry = args.dongle_init_retry if args else 5
+
```
Thalley: Suggest to put this into the exception scope

mkasenberg: Changed to use `getattr` instead, as it handles cases when args is None and `dongle_init_retry` does not exist in args.

Thalley: The local variable, `dongle_init_retry`, should still be moved to the inner scope where it's used. 

Would it make more sense to default the value of `dongle_init_retry` to 5, instead of using `getattr` and doing this manual handling of a hidden default value if not set?

sjanc (approved):

PR #1489, opened 2025-06-17
wid: Add missing WID for GAP layer

zhaynxp:
Added wid 246 and 247 to return handle of a characteristic with specified permissions which required by new MMIs.

They're used in following tests:
GAP/SEC/SEM/BV-40-C
GAP/SEC/SEM/BV-22-C

### DIFF
--- a/autopts/wid/gap.py
+++ b/autopts/wid/gap.py
@@ -1339,6 +1339,76 @@ def hdl_wid_243(_: WIDParams):
     return True
 
 
+def hdl_wid_246(_: WIDParams):
+    attrs = btp.gatts_get_attrs(type_uuid='2803')
+    bd_addr = btp.pts_addr_get()
+    bd_addr_type = btp.pts_addr_type_get()
+
+    perm = Perm.write_authn
+
+    for attr in attrs:
+        if not attr:
+            continue
+
+        (handle, permission, type_uuid) = attr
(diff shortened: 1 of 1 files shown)


autopts/wid/gap.py
```diff
@@ -1339,6 +1339,76 @@ def hdl_wid_243(_: WIDParams):
...
+        chrc_value_attr = btp.gatts_get_attrs(start_handle=handle,
+                                              end_handle=handle)
+        if not chrc_value_attr:
+            continue
+
+        (handle, permission, type_uuid) = chrc_value_attr[0]
+        if permission & perm:
+            return format(handle, 'x').zfill(4)
+
+    return False
```
sjanc: this was pointed recently in review (I already fixed couple of those),  lets not mix return values types.

So failed case here should just return 0

autopts/wid/gap.py
```diff
@@ -1339,6 +1339,76 @@ def hdl_wid_243(_: WIDParams):
...
```
sjanc: ditto

sjanc (changes requested):
Hi,

Looks OK, some nitpicks about returned value though

autopts/wid/gap.py
zhaynxp: Hi sjanc, 

Thanks a lot for your review and suggestion, I've changed it to return 0 .

sjanc (approved):

PR #1502, opened 2025-07-02
Fix AttributeError

Atm-Sara:
Fixed the exception "Exception: 'Namespace' object has no attribute 'copy'" when running without bot.

### DIFF
--- a/cliparser.py
+++ b/cliparser.py
@@ -4,6 +4,7 @@
 # auto-pts - The Bluetooth PTS Automation Framework
 #
 # Copyright (c) 2017, Intel Corporation.
+# Copyright (c) 2025, Atmosic.
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms and conditions of the GNU General Public License,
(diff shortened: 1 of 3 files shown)


cliparser.py
```diff
@@ -96,6 +97,10 @@ def __init__(self, cli_support=None, board_names=None, add_help=True):
         self.add_argument("--pylink_reset", action='store_true', default=False,
                           help="Use pylink reset.")
 
+        self.add_argument('--nc', dest='copy', action='store_false',
+                          help='Do not copy workspace, open original one. '
+                               'Warning: workspace file might be modified', default=True)
```
Thalley, line 102: Isn't there a way we an reuse the parser, or subparser, instead of duplicating options?

Atm-Sara: Deleted the one in common.py.

Thalley (commented):
Can you squash your 2 commits? :)

Thalley (approved):
LGTM, but perhaps someone with knowledge of the bot should review as well

sjanc (approved):

PR #1517, opened 2025-07-16
General non test-specific improvements to the debugging options.

MicImbriani:


cliparser.py
```diff
@@ -170,6 +170,10 @@ def __init__(self, cli_support=None, board_names=None, add_help=True):
                               "Requires rtt support on IUT.",
                               action='store_true', default=False)
 
+            self.add_argument("--rtt-log-syncto",
+                              help="XXX",
```
sjanc: :-)

autopts/config.py
```diff
@@ -21,7 +21,7 @@
 CLIENT_PORT = 65001
 BTMON_PORT = 65432
 
-MAX_SERVER_RESTART_TIME = 60
+MAX_SERVER_RESTART_TIME = 120
```
sjanc, line 24: while I don't have strong objections on changing defaults,  there is --max_server_restart_time command line option  for tuning this

cliparser.py
MicImbriani: Ah, you are referring to the very unhelpful help message I assume? I will add more words :)

autopts/config.py
MicImbriani: Good point, but that would require needing to remember which tests are affected by this, and which ones are not, so that a user would use the --max_server_restart_time only when running those tests. 
I would still suggest taking thins change in.

autopts/ptsprojects/stack/common.py
```diff
@@ -93,3 +93,55 @@ def wait_for_event(timeout, test, *args, **kwargs):
             return result
 
     return False
+
+
+def wait_event_with_condition(event_queue, condition_cb, timeout, remove):
```
mkasenberg, line 67: This function looks very similar to this one:
https://github.com/auto-pts/auto-pts/blob/8d3d69d10c1c69f552131f88714951eaff338e21/autopts/ptsprojects/stack/common.py#L45-L73
but with the difference that you have fixed the too wide indentation at `sleep(0.5)`. Maybe the `wait_for_queue_event(event_queue, test, timeout, remove)` could be updated? This new function signature `def wait_event_with_condition(event_queue, condition_cb, timeout, remove)` looks better.

autopts/ptsprojects/stack/common.py
```diff
@@ -93,3 +93,55 @@ def wait_for_event(timeout, test, *args, **kwargs):
...
+                return ev
+
+        # TODO: Use wait() and notify() from threading.Condition
+        #  instead of sleep()
+        sleep(0.5)
+
+    return None
+
+
+def wait_for_event_iut(event_queue, timeout, remove):
```
mkasenberg: Are you using any custom event queues? If not, maybe using a call like this:
https://github.com/auto-pts/auto-pts/blob/8d3d69d10c1c69f552131f88714951eaff338e21/autopts/ptsprojects/zephyr/iutctl.py#L247
would be enough?

autopts/ptsprojects/stack/common.py
```diff
@@ -93,3 +93,55 @@ def wait_for_event(timeout, test, *args, **kwargs):
...
+        raise_on_global_end()
+
+        for ev in event_queue:
+            if ev:
+                t.cancel()
+                if ev and remove:
+                    event_queue.remove(ev)
+                return ev
+
+            sleep(0.5)
```
mkasenberg: Is this indentation the right width?
```suggestion
        sleep(0.5)
```

autopts/client.py
```diff
@@ -1439,7 +1447,10 @@ def main(self, _args=None):
 
         os.makedirs(self.file_paths["TMP_DIR"], exist_ok=True)
 
-        self.load_test_case_database()
+        if self.args.test_cases_file:
+            tests = [line.strip() for line in self.args.test_cases_file.readlines() if line.strip()
+                and not line.strip().startswith("#")]
+            self.args.test_cases.extend(tests)
```
mkasenberg: Accidentally deleted `self.load_test_case_database()`. We can also do a few `strip()`s less:
```suggestion
        self.load_test_case_database()

        if self.args.test_cases_file:
            tests = [_line for line in self.args.test_cases_file.readlines()
                if (_line := line.strip()) and not _line.startswith("#")]
            self.args.test_cases.extend(tests)
```

cliparser.py
```diff
@@ -54,6 +54,9 @@ def __init__(self, cli_support=None, board_names=None, add_help=True):
                                "test cases can be specified by profile names."
                                 "Option can be used multiple times.")
 
+        self.add_argument("--test-cases-file",
+                          help="", type=argparse.FileType('r'))
```
mkasenberg: Maybe a comment like this will be helpful in the future:
```suggestion
                          help="A file with names of test cases to run. One test case per line. Use instead of -c option.", type=argparse.FileType('r'))
```

autopts/config.py
mkasenberg: With my laggy virtual machine I had to configure
`'max_server_restart_time': 600,`
so 120 is fine.

autopts/client.py
MicImbriani: Thanks, indeed it looks much cleaner :)

autopts/ptsprojects/stack/common.py
MicImbriani: Done.

autopts/ptsprojects/stack/common.py
MicImbriani: I actually noticed that this function was not used anywhere in my code and instead was kept for "just in case I need it". Your suggestion makes this not needed, so I removed it.

cliparser.py
MicImbriani: Done.

mkasenberg (approved):

PR #1518, opened 2025-07-16
Fix CCP tests edge cases.

MicImbriani:
The current implementation of CCP does not take into account the edge cases addressed in the commits


autopts/wid/ccp.py
```diff
@@ -279,11 +283,12 @@ def hdl_wid_114(params: WIDParams):
...
     ev = stack.ccp.wait_cp_ev(addr_type, addr, 20, remove=True)
-    if ev[2] != 0:
+
+    if ev is not None and ev[2] != 0:
         # Invalid opcode should be returned
-        return True
+        log(f'Invalid opcode {ev[2]}')
 
-    return False
+    return True
```
Thalley: Why are we not returning `False` here if `ev == None`?

autopts/wid/ccp.py
```diff
@@ -477,20 +484,22 @@ def hdl_wid_20206(params: WIDParams):
     """
     stack = get_stack()
 
-    chars = stack.ccp.events[defs.BTP_CCP_EV_CHRC_HANDLES][0]
-    chrc_list = [f'{chrc:04X}' for chrc in chars]
+    if len(stack.ccp.events[defs.BTP_CCP_EV_CHRC_HANDLES]) > 0:
```
Thalley, line 488: This is fine, but one could also have consider a simpler change with early termination:

```
if len(stack.ccp.events[defs.BTP_CCP_EV_CHRC_HANDLES]) == 0:
        return False
```

Thalley (changes requested):
Could you add the test cases fixed by this change? Then we can run the bot and verify it as well

autopts/wid/ccp.py
MicImbriani: Sorry, that was a mistake introduced while cleaning up the commits. It is fixed now. Thanks for the comment!

PR #1454, opened 2025-05-15
Enable test cases for feature `LE security mode 3`

lylezhu2012:
* btp: Add command Broadcast Source Setup v3

Compare with the command ` Broadcast Source Setup v2`, two command parameters are added.

Parameter `Encryption` is used to flag whether encrypt the data that is transmitted in a BIS.

Parameter `Broadcast Code` is used to set the `Broadcast_Code` if the parameter `Encryption` is not zero.

* Enable test cases for feature `LE security mode 3`

Supports following cases,
GAP/SEC/SEM/BV-31-C
GAP/SEC/SEM/BV-32-C
GAP/SEC/SEM/BV-34-C
GAP/SEC/SEM/BV-35-C

Update wid 300 to enable BIS source.

Add wids 352, 352, and 356.

PR #1519, opened 2025-07-17
Add support for ISO

MicImbriani:
Add support for ISO, and fix 2 GAP tests that depend on it.


MicImbriani:
> This seems like a subset of #1454

Seems like you are correct. What would be the suggested approach to include these changes in that commit?

PR #1527, opened 2025-07-28
bot & tools: Generating wid usage report

szymon-czapracki:
Add tool to generate wid report.
Add wid report functionality in bot.

### DIFF
--- a/autopts/utils.py
+++ b/autopts/utils.py
@@ -15,19 +15,32 @@
 #
 
 """Utilities"""
+import csv
 import ctypes
 import logging
 import os
+import re
 import sys
 import threading
 import traceback
 import xmlrpc.client
+from collections import defaultdict
+from pathlib import Path
 from time import sleep
 
 import hid
 import psutil
 
 PTS_WORKSPACE_FILE_EXT = ".pqw6"
+
+# Global paths for wid report
+BASE_DIR = Path(__file__).parent.parent.resolve()
+LOG_DIR = BASE_DIR / "logs"
+OUTPUT_CSV_PATH = BASE_DIR / "wid_usage_report.csv"
+
+# Regex patterns for log field parsing in wid report
+WID_REGEX = re.compile(r"^wid:\s*(\S+)")
+TC_NAME_REGEX = re.compile(r"^test_case_name:\s*(.+)")
 
 # A mechanism for safely terminating threads
 # after interrupt triggered with Ctrl+C
@@ -475,6 +488,81 @@ def log_memory_usage():
 
     mem_usage = mem_info.rss / (1024 ** 2)
     logging.debug(f"Memory usage: {mem_usage:.2f} MB")
+
+
+def extract_wid_testcases_to_csv():
+
+    # Example log block format:
+    # BEGIN OnImplicitSend:
+    # wid: 35
+    # test_case_name: GAP/ADV/BV-01-C
+    # ...
+    # END OnImplicitSend
+
+    profile_wid_map = defaultdict(lambda: defaultdict(set))
+
+    for log_file in LOG_DIR.rglob("*.log"):
+        logging.debug(f"Processing log file: {log_file}")
+        try:
+            with log_file.open(encoding='utf-8', errors='ignore') as f:
+                lines = f.readlines()
+        except Exception as e:
+            logging.warning(f"Failed to read {log_file}: {e}")
+            continue
+
+        # Track wheter inside an OnImplicitSend block
+        in_implicit_send_block = False
+        wid = None
+        test_case_name = None
+
+        for line in lines:
+            stripped = line.strip()
+
+            # Detect start of a new block
+            if not in_implicit_send_block and stripped.startswith("BEGIN OnImplicitSend:"):
+                in_implicit_send_block = True
+                wid = None
+                test_case_name = None
+
+            elif in_implicit_send_block:
+                # Match and extractWID and TC fields
+                wid_match = WID_REGEX.match(stripped)
+                tc_name_match = TC_NAME_REGEX.match(stripped)
+
+                if wid_match:
+                    wid = wid_match.group(1)
+                elif tc_name_match:
+                    test_case_name = tc_name_match.group(1)
+                elif stripped == "END OnImplicitSend":
+                    # End of block - store data if complete
+                    if wid and test_case_name:
+                        if '/' in test_case_name:
+                            profile = test_case_name.split('/')[0]
+                        else:
+                            profile = "UNKNOWN"
+                            logging.warning(f"Missing profile prefix in test_case_name: '{test_case_name}'")
+
+                        profile_wid_map[profile][wid].add(test_case_name)
+                    else:
+                        logging.error(
+                            "INCOMPLETE BLOCK:\n"
+                            f"  File           : {log_file.name}\n"
+                            f"  WID            : {wid!r}\n"
+                            f"  TestCase Name  : {test_case_name!r}"
+                        )
+                    in_implicit_send_block = False
+                    wid = None
+                    test_case_name = None
+
+    # Output results to CSV grouped by profile and wid
+    with OUTPUT_CSV_PATH.open('w', newline='', encoding='utf-8') as csvfile:
+        writer = csv.writer(csvfile)
+        for profile in sorted(profile_wid_map.keys()):
+            writer.writerow([f'{profile}'])
+            for wid in sorted(profile_wid_map[profile], key=lambda x: int(x)):
+                testcases = sorted(profile_wid_map[profile][wid])
+                testcases_combined = " ".join(testcases)
+                writer.writerow([wid, testcases_combined])
 
 
 def main():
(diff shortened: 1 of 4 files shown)


tools/wid_usage.py
Thalley, line 1: The content of the two files change seems quite identical. Sounds like some/most of the functionality should be put into a module or function

szymon-czapracki: Is the current version more acceptable?

Thalley: Yeah, that works much better :)

autopts/utils.py
```diff
@@ -477,6 +479,61 @@ def log_memory_usage():
     logging.debug(f"Memory usage: {mem_usage:.2f} MB")
 
 
+def extract_wid_testcases_to_csv():
+    log_dir = "./logs"
```
Thalley: If `./logs` is a specific path used by AutoPTS, it should be defined as a global value or be retrieved by a getter

Thalley: And since it's a relative path, is it relative to this file, or the CWD of where the script is called?

autopts/utils.py
```diff
@@ -477,6 +479,61 @@ def log_memory_usage():
...
+    output_csv_path = "./wid_usage_report.csv"
+    profile_wid_map = defaultdict(lambda: defaultdict(set))
+
+    for root, _, files in os.walk(log_dir):
+        for filename in files:
+            if filename.endswith(".log"):
```
Thalley: Consider replacing this with 

```suggestion
    for log_file in Path("./logs").rglob("*.log"):
```

To reduce the complexity and number of indendations

(Generally I would recommend using Pathlib over os.path for nearly everything)

tools/wid_usage.py
```diff
@@ -0,0 +1,7 @@
+
+from autopts.utils import extract_wid_testcases_to_csv
+
+# USE: Simply  run python3 -m tools.wid_usage in auto-pts directory
```
Thalley, line 19: I suppose this assumes that some tests have already been run and the logs are in the correct location relative to the CWD?

autopts/utils.py
```diff
@@ -477,6 +479,61 @@ def log_memory_usage():
...
+                filepath = os.path.join(root, filename)
+                with open(filepath, encoding='utf-8', errors='ignore') as f:
+                    lines = f.readlines()
+
+                in_block = False
```
Thalley: Can you elaborate on what `in_block` means?

autopts/utils.py
```diff
@@ -477,6 +479,61 @@ def log_memory_usage():
...
+                wid = None
+                test_case_name = None
+
+                for line in lines:
+                    stripped = line.strip()
+
+                    if stripped.startswith("BEGIN OnImplicitSend:"):
+                        in_block = True
+                        wid = None
+                        test_case_name = None
```
Thalley: Should this be

```suggestion
                    if not in_block and stripped.startswith("BEGIN OnImplicitSend:"):
                        in_block = True
                        wid = None
                        test_case_name = None
```

Otherwise isn't there a small risk that `wid` and `test_case_name` may be overwritten if `lines` for some reason contains multiple `BEGIN OnImplicitSend`?

autopts/utils.py
```diff
@@ -477,6 +479,61 @@ def log_memory_usage():
...
+                                profile_wid_map[profile][wid].add(test_case_name)
+
+                            in_block = stripped.startswith("BEGIN")
+                            if not in_block:
+                                wid = None
+                                test_case_name = None
+
+                if in_block and wid and test_case_name:
+                    profile = test_case_name.split('/')[0] if '/' in test_case_name else "UNKNOWN"
+                    profile_wid_map[profile][wid].add(test_case_name)
```
Thalley: I think it would be useful to add comments with examples of the lines this is parsing, so the code becomes more obvious. For example it's unclear why the lines are being split and then the expected value is in index `[1]`. 

Alternatively, did you consider using a regex for this purpose?

Thalley (changes requested):

autopts/utils.py
szymon-czapracki: Defined a global value for this

autopts/utils.py
szymon-czapracki: Done

tools/wid_usage.py
szymon-czapracki: Yes - if you use the tool, it's better if you have the logs :) If they are not present, this will just output an empty file.

autopts/utils.py
szymon-czapracki: It indicates whether we’re currently parsing inside a "BEGIN OnImplicitSend:" ... "END OnImplicitSend" log block. I also renamed it to in_implicit_send_block for clarity.

autopts/utils.py
szymon-czapracki: Good catch - done.

autopts/utils.py
szymon-czapracki: I’ve added example log lines as comments and switched to using regular expressions for parsing 'wid:' and 'test_case_name'. It’s now clearer what we expect and avoid fragile string-splitting.

tools/wid_usage.py
szymon-czapracki: I can always add a check if the logs folder exists, and if not, exit with log info, wdyt?

autopts/utils.py
```diff
@@ -477,6 +490,70 @@ def log_memory_usage():
...
+        for line in lines:
+            stripped = line.strip()
+
+            # Detect start of a new block
+            if not in_implicit_send_block and stripped.startswith("BEGIN OnImplicitSend:"):
+                in_implicit_send_block = True
+                wid = None
+                test_case_name = None
+
+            elif in_implicit_send_block:
```
Thalley, line 527: Up to you, but it should be possible to replace the loop with a regex. 

Something like 

```python
pattern = r"""
BEGIN OnImplicitSend:
.*?
wid:\s*(?P<wid>\d+)
.*?
test_case_name:\s*(?P<test_case_name>\w+/\w+/\w+-\w+)
.*?
END OnImplicitSend:
"""

result = re.search(pattern, text)

if result:
    wid = result["wid"]
    test_case_name = result["test_case_name"]
    print(f"wid: {wid}")
    print(f"test_case_name: {test_case_name}")
```

If you prefer the loop, that's also fine

autopts/utils.py
```diff
@@ -477,6 +490,70 @@ def log_memory_usage():
...
+
+                if wid_match:
+                    wid = wid_match.group(1)
+                elif tc_name_match:
+                    test_case_name = tc_name_match.group(1)
+                elif stripped == "END OnImplicitSend":
+                    # End of block - store data if complete
+                    if wid and test_case_name:
+                        profile = test_case_name.split('/')[0] if '/' in test_case_name else "UNKNOWN"
+                        profile_wid_map[profile][wid].add(test_case_name)
```
Thalley: Should we log an error if the profile is UNKNOWN?

autopts/utils.py
```diff
@@ -477,6 +490,70 @@ def log_memory_usage():
...
+                wid_match = WID_REGEX.match(stripped)
+                tc_name_match = TC_NAME_REGEX.match(stripped)
+
+                if wid_match:
+                    wid = wid_match.group(1)
+                elif tc_name_match:
+                    test_case_name = tc_name_match.group(1)
+                elif stripped == "END OnImplicitSend":
+                    # End of block - store data if complete
+                    if wid and test_case_name:
```
Thalley, line 538: Should we log an error if wid or test_case_name wasn't found?

autopts/utils.py
szymon-czapracki: I prefer the loop :)

autopts/utils.py
szymon-czapracki: Added log

autopts/utils.py
szymon-czapracki: Added log

tools/wid_usage.py
Thalley: Was wondering if this (simple) tool should perhaps have an argparser as well with a path to the logs (defaulting to e.g. ./logs)? 

But that can also wait until there's an actual request for that feature

tools/wid_usage.py
```diff
@@ -0,0 +1,7 @@
+
```
sjanc, line 16:
```
#!/usr/bin/env python3

#
# auto-pts - The Bluetooth PTS Automation Framework
#
# Copyright (c) 2025, Codecoup
#
# This program is free software; you can redistribute it and/or modify it
# under the terms and conditions of the GNU General Public License,
# version 2, as published by the Free Software Foundation.
#
# This program is distributed in the hope it will be useful, but WITHOUT
# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
# FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for
# more details.
#
```

szymon-czapracki: Done

PR #1531, opened 2025-08-04
Zephyr: Make TSPX_TARGET_LATENCY consistent

Thalley:
Modify 2 occurences of TSPX_TARGET_LATENCY to be
TARGET_BALANCED_LATENCY_RELIABILITY which is the value that the Zephyr BAP Unicast client uses.


sjanc (commented):
I guess this should also be set for BAP, otherwise PTS defaults to TARGE_LOTWER_LATENCY  (would also set PHY to 2M to keep things similar)

Thalley:
> I guess this should also be set for BAP, otherwise PTS defaults to TARGE_LOTWER_LATENCY (would also set PHY to 2M to keep things similar)

`TSPX_TARGET_PHY` is already set to `LE_2M_PHY` for all IXITs. For some reason the IXIT value doesn't exist for BAP as far as I can tell

sjanc (approved):

PR #1539, opened 2025-09-03
doc: Add commands for EAD handling

petrip-slab:
Added command for storing EAD keymaterial and encryptinon/decryption This is order to enable implementin wid's needed by tests GAP/ADV/BV-20-C and GAP/SCN/BV-01-C


petrip-slab:
those command are for the IUT, the device - zephyr for instance - must at step 1 create a valid random payload advertisement. And second one bit later.

"The Upper Tester orders the IUT to start advertising; the IUT enters broadcast mode or a
discoverable mode and begins advertising using Encrypted Advertising Data with a random
payload."

sjanc:
> those command are for the IUT, the device - zephyr for instance - must at step 1 create a valid random payload advertisement. And second one bit later.
> 
> "The Upper Tester orders the IUT to start advertising; the IUT enters broadcast mode or a discoverable mode and begins advertising using Encrypted Advertising Data with a random payload."

yes, this is what I was asking, if it is IUT that is doing encryption/decryption why explicit encrypt/decrypt BTP commands?  I'd rather expect implicit (if EAD key was provided) or explicit configuration when advertising is being started

sjanc:
oh!  I've just checked that both enrypted and unencrypted AD is allowed in same advertising PDU!
in this case I guess we indeed need interface to handle this ontop of IUT

OK, so I think this is good to go (and sorry for noise:P)

doc/btp_gap.txt
```diff
@@ -849,6 +849,49 @@ Commands and responses:
 
                 In case of an error, the error response will be returned.
 
+        Opcode 0x36 - Set EAD keymaterial
```
sjanc: any particular reason for not using 0x31 0x32 and 0x33 opcodes?

petrip-slab: No I will change them. Mayve there was some other work that planned to use one or more of them 

I will fix this

PR #1545, opened 2025-09-10
wid: Fix BAP WID 100 handling

sjanc:
BAP/BSNK/STR and BAP/BSRC/STR tests > 35 are new tests with single BIS.


autopts/wid/bap.py
```diff
@@ -158,7 +158,7 @@ def hdl_wid_100(params: WIDParams):
     if params.test_case_name.startswith('BAP/BSNK/STR') or \
             params.test_case_name.startswith('BAP/BSRC/STR'):
         tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-        if tc_num >= 18:
+        if tc_num >= 18 and tc_num <= 34:
```
Thalley, line 161: Could consider having a similar lookup table or something as we have for unicast

PR #1547, opened 2025-09-10
wid: add/modify GAP WIDs for new ISO tests

ardo-nordic:
Adding new and modifying some existing WIDs for new tests:
- GAP/BIS/BBM/BV-01-C
- GAP/BIS/BSE/BV-01-C
- GAP/SEC/SEM/BI-13-C

### DIFF
--- a/autopts/wid/gap.py
+++ b/autopts/wid/gap.py
@@ -1702,9 +1702,14 @@ def hdl_wid_300(params: WIDParams):
     return True
 
 
-def hdl_wid_301(_: WIDParams):
-    # Please click OK if IUT did not receive periodic advertising report.
+def hdl_wid_301(params: WIDParams):
+    # Please click OK when IUT received periodic advertisement report.
     stack = get_stack()
+
+    # GAP/PADV/PASE/BV-02-C starts PA Sync in WID 302 which comes first
+    if params.test_case_name not in ['GAP/PADV/PASE/BV-02-C']:
+        btp.gap_padv_create_sync(0, 0, 10, 0)
+
     return stack.gap.wait_periodic_report(10)
 
 
@@ -2564,29 +2569,46 @@ def hdl_wid_274(_: WIDParams):
     return True
 
 
+def hdl_wid_350(_: WIDParams):
+    '''
+    Wait for Broadcast ISO request.
+    '''
+    return True
+
+
+def hdl_wid_351(_: WIDParams):
+    '''
+    Wait for Broadcast ISO request.
+    '''
+    return True
+
+
 def hdl_wid_352(params: WIDParams):
     '''
     Please click OK when IUT establishes BIG sync, and ready to receive ISO data.
     '''
-    addr = pts_addr_get()
-    addr_type = pts_addr_type_get()
-
-    btp.gap_start_discov(transport='le', discov_type='passive', mode='observe')
-    sleep(10)
-    btp.gap_stop_discov()
-    if not btp.check_discov_results(addr_type=addr_type, addr=addr):
-        log('Peer device not found.')
-        return False
-
     stack = get_stack()
 
-    log('Synchronizing to broadcast')
-    btp.gap_padv_create_sync(0, 0, 0x200, 0)
-    if not stack.gap.wait_periodic_established(10):
-        log('Failed to periodic sync established')
-        return False
+    # create PA sync, except for tests that create it as part of WID 301
+    if params.test_case_name not in ['GAP/BIS/BSE/BV-01-C', 'GAP/SEC/SEM/BI-13-C']:
+        addr = pts_addr_get()
+        addr_type = pts_addr_type_get()
+
+        btp.gap_start_discov(transport='le', discov_type='passive', mode='observe')
+        sleep(10)
+        btp.gap_stop_discov()
+        if not btp.check_discov_results(addr_type=addr_type, addr=addr):
+            log('Peer device not found.')
+            return False
+
+        log('Synchronizing to broadcast')
+        btp.gap_padv_create_sync(0, 0, 0x200, 0)
+        if not stack.gap.wait_periodic_established(10):
+            log('Failed to periodic sync established')
+            return False
 
     biginfo = stack.gap.read_periodic_biginfo()
+    log(f'Big Info: {biginfo}')
     if not biginfo:
         log('Failed to read periodic biginfo')
         return False
@@ -2609,26 +2631,33 @@ def hdl_wid_352(params: WIDParams):
     return True
 
 
-def hdl_wid_351(_: WIDParams):
+def hdl_wid_353(_: WIDParams):
     '''
-    Wait for Broadcast ISO request.
+    Please configure IUT security to Mode 3 Level 2 or Level 3.
     '''
     return True
 
 
-def hdl_wid_356(_: WIDParams):
+def hdl_wid_354(params: WIDParams):
     '''
-    Please broadcast valid ISO data packets (more than 3 packets).
+    Please synchronize to lower tester in Security Mode 3 Level 2 or 3.
+    If IUT failed to synchnorize to the lower tester, please click ok.
     '''
     stack = get_stack()
-    try:
-        for _ in range(1, 100):
-            for bis_id in stack.gap.big_bis_data_path_setup:
-                btp.gap_bis_broadcast(bis_id, '00')
-    except types.BTPError:
-        log("Ignore BIS broadcast failure")
 
-    return True
+    biginfo = stack.gap.read_periodic_biginfo()
+    if not biginfo:
+        log('Failed to read periodic biginfo')
+        return False
+
+    broadcast_code = stack.gap.big_broadcast_code if biginfo.encryption else None
+
+    btp.gap_big_create_sync(biginfo.sid, biginfo.num_bis, 1, 255, broadcast_code)
+    if not stack.gap.wait_big_established():
+        log('Sync failed on BIG with low security level')
+        return True
+    else:
+        return False
 
 
 def hdl_wid_355(params: WIDParams):
@@ -2658,3 +2687,34 @@ def hdl_wid_355(params: WIDParams):
 
     log('Incorrect BIS data received')
     return False
+
+
+def hdl_wid_356(params: WIDParams):
+    '''
+    Please broadcast valid ISO data packets (more than 3 packets).
+    '''
+    stack = get_stack()
+    try:
+        match = re.search(r'(\d+)\W+packets', params.description)
+        pkt_cnt = int(match[1]) * 2 if match else 100
+        for _ in range(1, pkt_cnt):
+            for bis_id in stack.gap.big_bis_data_path_setup:
+                btp.gap_bis_broadcast(bis_id, '00')
+    except types.BTPError:
+        log("Ignore BIS broadcast failure")
+
+    return True
+
+
+def hdl_wid_357(_: WIDParams):
+    '''
+    Please send LE BIGInfo Advertising Report.
+    '''
+    btp.gap_create_big(0, 1, 10000, 20)
+
+    stack = get_stack()
+    if not stack.gap.wait_bis_data_path_setup():
+        log('Failed to setup BIS data path')
+        return False
+
+    return True


autopts/wid/gap.py
```diff
@@ -1702,10 +1702,17 @@ def hdl_wid_300(params: WIDParams):
     return True
 
 
-def hdl_wid_301(_: WIDParams):
+def hdl_wid_301(params: WIDParams):
     # Please click OK if IUT did not receive periodic advertising report.
+    # or
+    # Please click OK when IUT received periodic advertisement report.
```
Thalley: Can you specify when one happens and when the other happens? (it's somewhat obvious from the code though)

autopts/wid/gap.py
```diff
@@ -2658,3 +2673,34 @@ def hdl_wid_355(params: WIDParams):
...
+
+
+def hdl_wid_356(params: WIDParams):
+    '''
+    Please broadcast valid ISO data packets (more than 3 packets).
+    '''
+    stack = get_stack()
+    try:
+        match = re.search(r'(\d+)\W+packets', params.description)
+        pkt_cnt = int(match[1]) * 2 if match else 100
```
Thalley, line 2699: Can't we also just send 100? 

I feel like sending down to 6 would make it less robust

ardo-nordic: I'd say if sending twice as many packets as asked for is not enough, than something is wrong and we might be hiding issues.

autopts/wid/gap.py
```diff
@@ -2609,26 +2631,19 @@ def hdl_wid_352(params: WIDParams):
     return True
 
 
-def hdl_wid_351(_: WIDParams):
+def hdl_wid_353(_: WIDParams):
     '''
-    Wait for Broadcast ISO request.
+    Please configure IUT security to Mode 3 Level 2 or Level 3.
     '''
     return True
```
sjanc, line 2638: Should we really fake this?   That test will pass not because IUT is in higher sec level and rejects lower security, but simply by doing nothing and confirming that it done nothing..

ardo-nordic: All the checks are actually done in the WID 354, which I modified to avoid false negatives. It checks that we start sync on received big info, but fail.

autopts/wid/gap.py
Thalley: I've already expressed my concerns about how this was defined in the first place (IMO the IUT is way better at handling TX than the upper tester, especially if considering keeping track of PSNs, etc.), but if it works, it works, and we can improve on it later if needed.

autopts/wid/gap.py
```diff
@@ -1702,10 +1702,17 @@ def hdl_wid_300(params: WIDParams):
...
 
-def hdl_wid_301(_: WIDParams):
+def hdl_wid_301(params: WIDParams):
     # Please click OK if IUT did not receive periodic advertising report.
+    # or
+    # Please click OK when IUT received periodic advertisement report.
     stack = get_stack()
-    return stack.gap.wait_periodic_report(10)
+
+    if params.test_case_name in ['GAP/BIS/BSE/BV-01-C', 'GAP/SEC/SEM/BI-13-C']:
```
Thalley: What are the difference between `'GAP/BIS/BSE/BV-01-C', 'GAP/SEC/SEM/BI-13-C'` and the others? 

If `'GAP/BIS/BSE/BV-01-C', 'GAP/SEC/SEM/BI-13-C'` belong in some specific set of tests, then I suggest to put them into a list variable with a relevant name

ardo-nordic: For some reason same WID has been reused for scenarios which are very different from each other. In my opinion it should've been handled by a new WID.

Thalley: I agree - But that's unfortunately a common design choice by the PTS team - But since you've figured out which tests use one type, and which use the other type, I hoped that you had some idea of what the difference between the tests are? Number of BIS? Encrypted/unencrypted?

ardo-nordic: Well they just require completely different actions, as seen from the code. I can put the test names into a list, no problem, just haven't seen it being a common practice anywhere, at least not in that file.

Thalley: Instead of creating this ad-hoc list of testcases 2 places, I do think it's better to keep them in a single list somewhere.

ardo-nordic: Done now.

autopts/wid/gap.py
```diff
@@ -29,6 +29,9 @@
 log = logging.debug
 
 
+GAP_ISO_TESTS = ['GAP/BIS/BSE/BV-01-C', 'GAP/SEC/SEM/BI-13-C']
```
Thalley: This is misleading as there are several GAP_ISO tests not in this list

sjanc: TBH I'd be fine with opencoded list in wid handler  (this is at least consistent with what is used so far)

`if params.test_case_name in ['TEST_FOO', 'TEST_BAR']:`

autopts/wid/gap.py
```diff
@@ -1702,10 +1705,17 @@ def hdl_wid_300(params: WIDParams):
...
+    # or
+    # Please click OK when IUT received periodic advertisement report.
     stack = get_stack()
-    return stack.gap.wait_periodic_report(10)
+
+    if params.test_case_name in GAP_ISO_SYNC_TESTS:
+        btp.gap_padv_create_sync(0, 0, 10, 0)
+        return stack.gap.wait_periodic_established(10)
+    else:
+        return stack.gap.wait_periodic_report(10)
```
Thalley: Something here still doesn't look right.

The description is 

```
# Please click OK if IUT did not receive periodic advertising report.
# or
# click OK when IUT received periodic advertisement report.
```

So in some cases, we expect a PA report (and should click OK), and in some cases we don't expect a PA report (and should click OK). 

The implementation before seems wrong, i.e. it returns `return stack.gap.wait_periodic_report(10)` for `# Please click OK if IUT did not receive periodic advertising report.` - Would this click OK regardless of whether we receive the PA report? 

The new implementation here returns `return stack.gap.wait_periodic_established(10)`, but that doesn't seem to be related to the WID check at all since it doesn't check for any PA reports

ardo-nordic: Hardly a big surprise that PTS dialog descriptions isn't super accurate. Can't comment anything regarding initial handler, since I don't know which tests use it. As for the tests in this PR, it is clearly not enough to just get periodic report, we need a periodic sync.

Thalley: > it is clearly not enough to just get periodic report, we need a periodic sync.

Are you familiar with the periodic sync feature?

You cannot get a periodic advertising report before you have a sync, which is why I don't think the current implementaion here is correct, as we specifically don't wait for a periodic advertising report, which is what the WID is about. 

So it may be that it should have been

```
    if params.test_case_name in GAP_ISO_SYNC_TESTS:
        btp.gap_padv_create_sync(0, 0, 10, 0)
        stack.gap.wait_periodic_established(10)
    
    return stack.gap.wait_periodic_report(10)
```

Or even better, instead of checking for `if params.test_case_name in GAP_ISO_SYNC_TESTS:`, then check if we have a PA sync, and if not, then establish one first, and then wait for the report (i.e. replace `if params.test_case_name in GAP_ISO_SYNC_TESTS:` with something like `if stack.gap.pa_sync is None:`

And lastly, the WID implement based on the 2 (quite opposite instructions), we need to wait for a report, and then either press OK when we receive it, or don't receive it (the latter being kind of hard to do...)

ardo-nordic: I see, might have been some confusion on my side, the sync feature was just prerequisite for ISO and I haven't dug deep into it. I'll change to just checking for report, but I think I'll leave sync creation command only for the tests in iso list since calling the create sync command in for other tests might result in a failure if it has been called before I guess. As for having opposite instructions, again, it's a question to PTS implemented scenarios, isn't it?

Thalley: > I see, might have been some confusion on my side, the sync feature was just prerequisite for ISO and I haven't dug deep into it. I'll change to just checking for report, but I think I'll leave sync creation command only for the tests in iso list since calling the create sync command in for other tests might result in a failure if it has been called before I guess. 

If the difference between the tests in the list, and the other tests, is that the tests in the list does not have a PA sync prior to this WID, then I think checking if we have a PA sync established is a more robust solution than keeping a list of test cases that need to have PA sync established here, although that may be different than what AutoPTS usually does. @sjanc what do you think? 

> As for having opposite instructions, again, it's a question to PTS implemented scenarios, isn't it?

Sure, but our WID handler should still reflect that, i.e. we cannot just return `return stack.gap.wait_periodic_report(10)` for both cases; I guess depending on the description we should either do 

```python
return stack.gap.wait_periodic_report(10)
```
or
```python
return not stack.gap.wait_periodic_report(10)
```

Right?

ardo-nordic: > > As for having opposite instructions, again, it's a question to PTS implemented scenarios, isn't it?
> 
> Sure, but our WID handler should still reflect that, i.e. we cannot just return `return stack.gap.wait_periodic_report(10)` for both cases; I guess depending on the description we should either do
> 
> ```python
> return stack.gap.wait_periodic_report(10)
> ```
> 
> or
> 
> ```python
> return not stack.gap.wait_periodic_report(10)
> ```
> 
> Right?

I think changing existing handler logic is somewhat outside the scope of this PR. That can be inaccurate or outright wrong description in the WID, or anything else, and to fix that requires finding tests which use it and figuring out what is actually going on there.

Thalley: > I think changing existing handler logic is somewhat outside the scope of this PR. That can be inaccurate or outright wrong description in the WID, or anything else, and to fix that requires finding tests which use it and figuring out what is actually going on there.

That's a fair point, but you are adding new functionality to the handler, that needs to handle the opposite case of what already exists there, and you currently do that by not changing the behavior :) So it's also very weird to apply the opposite behavior by not changing the current behavior. At the very least we should add a comment why we return the same value to verify opposite behaviors 

The current implementation of the WID handler was implemented 3 years ago, and the author probably don't recall the choice here  @KKopyscinski ?). 

@sjanc I tried to run and verify the changes here with https://github.com/auto-pts/auto-pts/pull/1547#issuecomment-3280737615 but it didn't trigger the bot. Did something change?

ardo-nordic: > That's a fair point, but you are adding new functionality to the handler, that needs to handle the opposite case of what already exists there, and you currently do that by not changing the behavior :) So it's also very weird to apply the opposite behavior by not changing the current behavior. At the very least we should add a comment why we return the same value to verify opposite behaviors

I feel like I don't quite understand your point. New behavior == new tests, which have the same WID, but it even has  different text in it, which is obviously a bad decision, but not something we can change without submitting a ticket to PTS and wait for a year to get it fixed. The "old behavior", which is servicing a different set of tests (not known to me), is left unchanged in this PR. Since it apparently worked, changing it will likely result in failure. That said, I see the discrepancy between what is described in WID text and the action taken.

KKopyscinski: I took a look here, there are currently two tests that use this WID: `GAP/PADV/PASE/BV-01-C` and `GAP/PADV/PASE/BV-05-C`. For both, the WID text is "Please click OK if IUT did not received periodic advertising report." which aligns with a step in TS: 
> The IUT receives the periodic advertising synchronization information sent from the Lower Tester and
reports it to the Upper Tester.
The IUT does not report periodic advertising events to the Upper Tester.

For this WID description, and these tests, it should be as is now:
```python
def hdl_wid_304(_: WIDParams):
    # Please click OK if IUT did not receive periodic advertising report.
    stack = get_stack()
    return not stack.gap.wait_periodic_report(10)
```
so wait 10 seconds for the report and confirm it wasn't there.

Sadly, there are many tests that re-use WIDs with different description. In this situation, we do check what test requests this MMI, and add condition for it, or make it more generic and take action based on description text. I can see that sometimes it gets as bad as https://github.com/auto-pts/auto-pts/blob/master/autopts/wid/gap.py#L478 :)

But all in all, I believe the line `return not stack.gap.wait_periodic_report(10)`, with `not`, is correct for those two tests.

Thalley: > Since it apparently worked, changing it will likely result in failure. 

Sure, but chances are that it has been reporting a false positive for the last 3 years :) Arguably that's outside the scope of this PR, which is why I wanted a comment stating this in the code, as the current version with 2 opposite expected behavior using the same behavior to validate, seems like a bug (and probably is, but in the existing code/tests, and not the ones you add)

sjanc: > If the difference between the tests in the list, and the other tests, is that the tests in the list does not have a PA sync prior to this WID, then I think checking if we have a PA sync established is a more robust solution than keeping a list of test cases that need to have PA sync established here, although that may be different than what AutoPTS usually does. @sjanc what do you think?

if this is a case we could maybe extend Current_Settings with 'PA synchronizing' state and use that (we were thinking on using settings for this for a long time but so war those were informative only)

I do recall that we use to have similar cases in past already, where PTS depending on role supported or test was expecting different actions in same WID...

ardo-nordic: > I took a look here, there are currently two tests that use this WID: `GAP/PADV/PASE/BV-01-C` and `GAP/PADV/PASE/BV-05-C`. For both, the WID text is "Please click OK if IUT did not received periodic advertising report." which aligns with a step in TS:

If that's correct, and these two are the only tests that used to be with WID 301, than apparently this WID has been reallocated for other purposes and superseded by WID 304 which actually does the right thing and has the same description.

@Thalley , I'll remove the reference to old implementation and dialog altogether, but guess we should wait for PTS results, if we can trigger it after all?

Thalley: Yeah, if that's the case, then removing the `# Please click OK if IUT did not receive periodic advertising report.` from WID 301 handling making sense

Do you still prefer the choice of `if params.test_case_name in GAP_ISO_SYNC_TESTS:` compared to checking if we have a PA sync at the moment, i.e. a testcase based check rather than a state check? 

We should definitely trigger the existing PA sync tests and the new ISO tests

ardo-nordic: > Do you still prefer the choice of `if params.test_case_name in GAP_ISO_SYNC_TESTS:` compared to checking if we have a PA sync at the moment, i.e. a testcase based check rather than a state check?

I checked if there is something which keeps the state of PA sync, but there is nothing like that actually. There are some transient variables used to capture related events, but they gets cleared and thus are not reliable. So guess I'll stick with test list.

Thalley: That's OK for now. In the future I'd like to see the other way around; we do that with some other cases like connections, BIG syncs, etc. IIRC.

sjanc:
#AutoPTS run zephyr GAP/BIS/BBM/BV-01-C GAP/BIS/BSE/BV-01-C GAP/SEC/SEM/BI-13-C GAP/BIS/BSE/BV-01-C GAP/PADV/PASE/BV-02-C GAP/PADV/PASE/BV-04-C GAP/PADV/PASE/BV-06-C

sjanc:
GAP/PADV/PASE/BV-02-C seems to regress with this

this test calls wid 302 and then wid 301

Thalley:
> also this requires CONFIG_BT_ISO_BROADCASTER enabled in tester, wonder how we should handle this for GAP tests
> 
> this is currently enabled in le-audio overlay but maybe we should have autopts overlays for GAP tests? or just enabled this in prj.conf as part of GAP settings?
> 
> @Thalley what is your call on this?

I don't see an issue of always enabling the ISO features in the host build. It won't affect anything else IIRC

sjanc:
alright   https://github.com/zephyrproject-rtos/zephyr/pull/96050

ardo-nordic:
> I think this needs to be conditional in wid 301... (or maybe remove this from 302 and just add comment for clarification why this is done in 301?)

I've added conditional handling for the failing PASE test into WID 301.

ardo-nordic:
> Please also have a look at the Codepilot comment (need to import `re`).

Already at line 19, not sure how could it have missed it.

sjanc (approved):

PR #1556, opened 2025-09-15
cli: Add '--wid_run' option to run tcs by service & WID

szymon-czapracki:
Introduce --wid_run SERVICE WID argument in CliParser
Add wid_run_tcs() helper to resolve testcases from wid_usage_report.csv
Default wid_usage_report.csv path is BASEDIR/wid_usage_report.csv
Extend args.test_cases with resolved testcases when --wid_run is used
Expose load_wid_report() and get_tc_from_wid() utilities in utils.py for parsing wid_usage_report.csv and mapping (service, wid) -> testcases

**_NOTE -_** This will be fully functional after a complete wid_report.csv file is added.

### DIFF
--- a/autopts/utils.py
+++ b/autopts/utils.py
@@ -586,6 +586,57 @@ def extract_wid_testcases_to_csv(log_dir: Path = None):
     print(f"WID usage report saved to: {OUTPUT_CSV_PATH.resolve()}")
 
 
+Sep = re.compile(r'[,\s;|]+')
+
+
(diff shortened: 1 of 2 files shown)


mkasenberg (commented):
I suggest making the path to the .csv configurable, something like this https://github.com/mkasenberg/auto-pts/commit/3d2b93e634505b3cad7ee4fa709a62b9322fd87d

PR #1558, opened 2025-09-15
utils: Make WID report destination configurable

szymon-czapracki:
Now WID report destination can be set as the Path lives in config file.

### DIFF
--- a/autopts/utils.py
+++ b/autopts/utils.py
@@ -31,12 +31,13 @@ from time import sleep
 import hid
 import psutil
 
+from autopts.config import FILE_PATHS
+
 PTS_WORKSPACE_FILE_EXT = ".pqw6"
 
 # Global paths for wid report
 BASE_DIR = Path(__file__).parent.parent.resolve()
 LOG_DIR = BASE_DIR / "logs"
-OUTPUT_CSV_PATH = BASE_DIR / "tmp" / "wid_usage_report.csv"
 
 # Regex patterns for log field parsing in wid report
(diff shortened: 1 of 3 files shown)


autopts/utils.py
```diff
@@ -31,12 +31,14 @@
...
 
+from autopts.config import FILE_PATHS
+
 PTS_WORKSPACE_FILE_EXT = ".pqw6"
 
 # Global paths for wid report
 BASE_DIR = Path(__file__).parent.parent.resolve()
 LOG_DIR = BASE_DIR / "logs"
-OUTPUT_CSV_PATH = BASE_DIR / "tmp" / "wid_usage_report.csv"
+OUTPUT_CSV_PATH = Path(FILE_PATHS['WID_USE_CSV_FILE'])
```
mkasenberg: This import is done before the bot config.py is parsed, so overriding the path will not work.

mkasenberg (changes requested):

autopts/utils.py
mkasenberg: I added this
```python
OUTPUT_CSV_PATH = Path(FILE_PATHS['WID_USE_CSV_FILE'])
```
inside the `extract_wid_testcases_to_csv` and now it works.

mkasenberg (approved):

autopts/utils.py
szymon-czapracki: Thanks :)

PR #1568, opened 2025-09-25
wid: Use larger sync timeout in GAP WID 301

sjanc:
PTS seems to be using quite large periodic advertising intervals in some tests. To improve reliability use larger sync timeout.

This fix GAP/BIS/BSE/BV-01-C.

### DIFF
--- a/autopts/wid/gap.py
+++ b/autopts/wid/gap.py
@@ -1708,7 +1708,7 @@ def hdl_wid_301(params: WIDParams):
 
     # GAP/PADV/PASE/BV-02-C starts PA Sync in WID 302 which comes first
     if params.test_case_name not in ['GAP/PADV/PASE/BV-02-C']:
-        btp.gap_padv_create_sync(0, 0, 10, 0)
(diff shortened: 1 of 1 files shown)


autopts/wid/gap.py
```diff
@@ -1708,7 +1708,7 @@ def hdl_wid_301(params: WIDParams):
 
     # GAP/PADV/PASE/BV-02-C starts PA Sync in WID 302 which comes first
     if params.test_case_name not in ['GAP/PADV/PASE/BV-02-C']:
-        btp.gap_padv_create_sync(0, 0, 10, 0)
+        btp.gap_padv_create_sync(0, 0, 500, 0)
```
Thalley, line 1711: As I also mentioned on Discord: The appropriate way of handling this would be to apply a function to the PA interval we received. While unlikely, this may still fail if PTS changes the PA interval to something larger (is the interval PTS uses an IXIT value?)

sjanc: yes, I've already added this to my TODO

just want to keep changes minimal until we stabilize tests results

PR #1583, opened 2025-10-03
zephyr: add test cases to exceptioned list for creating sync

m-alperen-sener:
GAP/PADV/PASE/BV-04-C
GAP/PADV/PASE/BV-06-C

Sync is already established in previous steps following previous WIDs, adding test cases in the exception list to not start sync again.

### DIFF
--- a/autopts/wid/gap.py
+++ b/autopts/wid/gap.py
@@ -1707,7 +1707,9 @@ def hdl_wid_301(params: WIDParams):
     stack = get_stack()
 
     # GAP/PADV/PASE/BV-02-C starts PA Sync in WID 302 which comes first
-    if params.test_case_name not in ['GAP/PADV/PASE/BV-02-C']:
+    if params.test_case_name not in ['GAP/PADV/PASE/BV-02-C',
+                                     'GAP/PADV/PASE/BV-04-C',
+                                     'GAP/PADV/PASE/BV-06-C']:
         btp.gap_padv_create_sync(0, 0, 500, 0)
 
     return stack.gap.wait_periodic_report(10)


sjanc:
#AutiPTS run zephyr GAP/PADV

sjanc:
#AutoPTS run zephyr GAP/PADV

autopts/wid/gap.py
```diff
@@ -1707,7 +1707,9 @@ def hdl_wid_301(params: WIDParams):
     stack = get_stack()
 
     # GAP/PADV/PASE/BV-02-C starts PA Sync in WID 302 which comes first
-    if params.test_case_name not in ['GAP/PADV/PASE/BV-02-C']:
+    if params.test_case_name not in ['GAP/PADV/PASE/BV-02-C',
+                                     'GAP/PADV/PASE/BV-04-C',
+                                     'GAP/PADV/PASE/BV-06-C']:
```
Thalley, line 1712: Just adding the same comment here as I did for the PR adding this list in the first place:
IMO it makes more sense to keep track of the PA sync state in the `stack` and then check if we have a PA sync at this point or not, and if not, then we establish the sync. 

I assume that the newly added tests don't have a WID call earlier to establish the PA sync?

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (10)</summary>GAP     GAP/PADV/PAM/BV-01-C            PASS<br>GAP     GAP/PADV/PASE/BV-01-C           PASS<br>GAP     GAP/PADV/PASE/BV-02-C           PASS<br>GAP     GAP/PADV/PASE/BV-03-C           PASS<br>GAP     GAP/PADV/PASE/BV-04-C           PASS<br>GAP     GAP/PADV/PASE/BV-05-C           PASS<br>GAP     GAP/PADV/PASE/BV-06-C           PASS<br>GAP     GAP/PADV/PASM/BV-01-C           PASS<br>GAP     GAP/PADV/PAST/BV-01-C           PASS<br>GAP     GAP/PADV/PAST/BV-02-C           PASS</details>

autopts/wid/gap.py
sjanc: I'd extend settings with PA_SYNCING and eg CONN_INITIATING etc and start using those,  however this requires for UT to properly update settings (yet, we could enforce those when issuing command that should result in settings change)

this way state doesn't have to be in stack (which is IUT specific) but in generic code
this is a slightly bigger refactor though

m-alperen-sener: > Just adding the same comment here as I did for the PR adding this list in the first place: IMO it makes more sense to keep track of the PA sync state in the `stack` and then check if we have a PA sync at this point or not, and if not, then we establish the sync.

Hard coding is a bit safer in that sense I think, because according to test spec sync should be established in step 4, and after that, by any means, re-establishing sync is not OK IMO. So test should fail if sync is broken in between test steps after 4. If we add this info to stack and every time when sync is broken, IUT will be ordered to sync again and verify the steps.

Test spec section for BV-4:

4. The Lower Tester executes the Periodic Advertising Synchronization Transfer procedure over the
LE connection with the Periodic Advertising Type specified in Table 4.41.
5. The Upper Tester receives periodic advertising synchronization information from the IUT.
6. The Upper Tester receives periodic advertising reports with periodic advertising data from the
IUT.
7. Terminate the connection between the IUT and the Lower Tester

______________________________________________________________________
> I assume that the newly added tests don't have a WID call earlier to establish the PA sync?

IUT is ordered to establish sync with WID 308 earlier.

WIDs sent by PTS in order:

wid: 308: **Step 4**
    description: Click OK when IUT is ready to perform Periodic Advertising Synchronization Establishment Procedure with listening for periodic advertising events.

wid: 309 **Step 5**
    description: Click OK when IUT receives periodic advertising synchronization information.
    
wid: 301 **Step 6**
    test_case_name: GAP/PADV/PASE/BV-04-C
    description: Please click OK when IUT received periodic advertisement report.
    
 __________________________________________________________________________

**The situation is the same for BV-6**

sjanc (approved):

m-alperen-sener:
Additional comment for such cases:

In my opinion the cause of this problem is PTS it self, and their WID abuse. The test steps 4,5 and 6 should have been realized only one WID; establish sync with this parameters and verify that IUT reports that and this event, this amount of time.

It is OK to used those extra WID (like 309 or 301) only if there are not used in other test cases to both establish sync and verify events.

Implementation in that sense can take a wrong turn to address those extra WIDs and hide issues in IUT functionality.

We should start giving feedback to PTS to reduce such WID abuse IMO

PR #1588, opened 2025-10-13
bot: add no_retry_on_regression param

piotrnarajowski:
This param will be used on weekly runs. The purpose is to skip retrying test cases that have failed in previous runs. This way we can save some time


autopts/bot/common.py
```diff
@@ -121,6 +121,7 @@ def __init__(self, args, **kwargs):
         self.bd_addr = args.get('bd_addr', '')
         self.enable_max_logs = args.get('enable_max_logs', False)
         self.retry = args.get('retry', 0)
+        self.skip_retry = args.get('skip_retry')
```
Thalley: What is the difference between `skip_retry` and setting `retry=0`?

piotrnarajowski: When `skip_retry` is used, failed test cases are handled as follows: if test failure is not a regression, test case will not be retried (i.e. `retry` is ignored). If the failure is regression, test case will be retried for `retry` number of times. If you set `retry` to zero, no failed test cases will be retried.

Thalley: I think this needs to be significantly better documented, or/and the parameter should be renamed. The current naming, IMO, implies that it has the same behavior as `retry=0`, but that is not the case.

sjanc: yeap, I'd suggest self describing name, eg no_retry_on_regression  or similar

sjanc: and as discussed offline,  lets add readme in bot folder describing options and showing how to set this up

cliparser.py
```diff
@@ -85,6 +85,12 @@ def __init__(self, iut_modes=None, board_names=None, add_help=True, *args, **kwa
                           help="Repeat test if failed. Parameter specifies "
                                "maximum repeat count per test")
 
+        self.add_argument("--no_retry_on_regression", type=bool,
+                          help="When skip_retry is used, failed test cases are handled as follows: if test failure is "
```
sjanc: when no_retry_on_regression option is used

sjanc (commented):
also pleas amend commit message with updated param name :)

cliparser.py
piotrnarajowski: Done

sjanc (approved):

PR #1594, opened 2025-10-17
PTS:ZEPHYR: ENCRYPTED ADV TEST

mahori-ravinder:
Add Support for GAP/ADV/BV-20-C
Adds handling of WID 500, 501 and 503


autopts/ptsprojects/zephyr/gap.py
```diff
@@ -52,7 +57,15 @@ class CHAR:
                          Perm.read_authn | Perm.write_authn,
                          UUID.VND16_5),
                 TestFunc(btp.gatts_set_val, 0, '04'),
-                TestFunc(btp.gatts_start_server)]
+
+                #add encrypted data key value
+                TestFunc(btp.gatts_add_svc, 0, '0018'),
+                TestFunc(btp.gatts_add_char, 0,
+                         Prop.read, Perm.read, UUID.EDKM),
+                TestFunc(btp.gatts_set_val, 0, SESSION_KEY+IV),
```
sjanc, line 65: if PTS is looking for "Encrypted Data Key Material" characteristic in GAP service than I think this should be handled by IUT, and not by autoPTS    (I'm not sure on how zephyr will handle this but this looks like you'd create second instance of GAP service?)

Alternative would be to request PTS support ticket to use IXIT for key material or maybe IXIT for service UUID that would include "Encrypted Data Key Material" characteristic (16 or 128 bits),  this could default to GAP service

PR #1597, opened 2025-10-23
bot: add option to skip copying workspace file

piotrnarajowski:
Rename copy argument to copy_workspace for clarity. Set copy_workspace in your config_prj.py to false or use command line arg --nc if you don't want auto-pts to copy your .pqw6 file and keep working with original file. This will now keep log files in autopts/workspaces/your_workspace directory which makes report generation process a lot easier. You can go back and forth between auto-pts testing and manual seamlessly.

### DIFF
--- a/autopts/bot/common_features/report.py
+++ b/autopts/bot/common_features/report.py
@@ -100,7 +100,7 @@ def make_report_xlsx(report_xlsx_path: str,
                      regressions_list: list,
                      progresses_list: list,
                      descriptions: dict,
-                     xmls: str,
+                     xmls: str | None,
                      errata: dict):
     """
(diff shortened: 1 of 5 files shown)


autopts/bot/common_features/report.py
```diff
@@ -491,7 +500,8 @@ def _pull_logs(_pts):
         for addr in servers:
             with ServerProxy(f"http://{addr}:{servers[addr][0]}/",
                              allow_none=True) as proxy:
-                _pull_logs(proxy)
+                if args.copy_workspace:
```
mkasenberg: Let's move this check up to avoid entering the pull_server_logs function. I will git push here, because it will require other small changes.

PR #1601, opened 2025-10-30
BTP: CSIS: Add more commands

alexsven:
- Adds command to set size + rank
- Adds command to set SIRK
- Adds wid 20 and 21 for CSIS
- Allows the following tests to pass:
 - CSIS/SR/CN/BV-01-C
 - CSIS/SR/CN/BV-02-C
 - CSIS/SR/CN/BV-03-C
 - CSIS/SR/CN/BV-04-C


autopts/wid/csis.py
```diff
@@ -53,6 +53,24 @@ def hdl_wid_7(_: WIDParams):
...
+    return True
+
+
+def hdl_wid_21(_: WIDParams):
+    """
+        Please update the characteristic Set Identity Resolving Key value with different data and enter connectable mode.
+    """
+    btp.csis_set_sirk("012345689ABCDEF0")
+    btp.csis_set_set_size(3, 1)
+    return True
```
Thalley: I don't like the statefullness of these commands. 

If a test would ever call both, or either of these multiple times, they would attempt to set the same value, which wouldn't trigger a notification. 

For the set SIRK we could probably randomly generate a value that is unique and thus always set a new value, even if these are called multiple times. 

For the set size, it's a bit more tricky, but I guess we could keep track of current size, and then just increment it?

autopts/wid/csis.py
```diff
@@ -53,6 +53,24 @@ def hdl_wid_7(_: WIDParams):
...
+
+
+def hdl_wid_21(_: WIDParams):
+    """
+        Please update the characteristic Set Identity Resolving Key value with different data and enter connectable mode.
+    """
```
Thalley, line 90: Are both of these command actually identical, or is this a copy-paste error?

autopts/wid/csis.py
```diff
@@ -53,6 +53,24 @@ def hdl_wid_7(_: WIDParams):
     return True
 
 
+def hdl_wid_20(params: WIDParams):
+    """
+        Please update the characteristic Set Identity Resolving Key value with different data and send a notification.
+    """
+    btp.csis_set_sirk("012345689ABCDEF0")
+    btp.csis_set_set_size(3, 1)
```
Thalley: The text only mentions the set SIRK, but the command also updates the size. Is that intentional?

autopts/pybtp/defs.py
```diff
@@ -758,6 +758,8 @@ def BIT(bit):
 BTP_CSIS_CMD_SET_MEMBER_LOCK = 0x02
 BTP_CSIS_CMD_GET_MEMBER_RSI = 0x03
 BTP_CSIS_CMD_SET_SIRK_TYPE = 0x04
+BTP_CSIS_CMD_SET_SIRK = 0x05
+BTP_CSIS_CMD_SET_SET_SIZE = 0x06
```
Thalley, line 762: The new commands need to be added to the `btp_csis.txt`

Thalley (changes requested):

autopts/wid/csis.py
alexsven: These are not the same, one ends in "... and send a notification.", the other ends in "... and enter connectable mode."

Thalley: Sorry, yeah, the descriptions are different, but the implementations are no, i.e. `hdl_wid_21` does not make the IUT enter the connectable mode for example. Consider adding a comment if that is intentional

autopts/wid/csis.py
```diff
@@ -53,6 +54,48 @@ def hdl_wid_7(_: WIDParams):
...
+        Please update the characteristic Coordinated Set Size value with different data and enter connectable mode.
+    """
+    # The same wid is used for two different test cases, with different descriptions, so we check the description.
+    if 'Set Identity Resolving Key' in params.description:
+        # Generate 16 random charachters for SIRK
+        sirk = secrets.token_hex(16)
+        btp.csis_set_sirk(sirk)
+
+    elif 'Coordinated Set Size' in params.description:
+        btp.csis_set_set_size(2, 1)
```
Thalley: As pointed out by https://github.com/auto-pts/auto-pts/pull/1601/files#r2489437326 this only works if the current set size isn't 2. We should consider setting the set size right after registering CSIS (so that we know what the set size and store that in the `stack`), and then we can probably just increment it here. That way we know exactly what the set size is at any point, as well ensuring that this function works even if the IUT default set size changes

autopts/wid/csis.py
```diff
@@ -53,6 +54,48 @@ def hdl_wid_7(_: WIDParams):
...
+
+    else:
+        log(f"hdl_wid_21: Unknown description: {params.description}")
+        return False
+
+    return True
```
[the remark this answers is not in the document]
Thalley: Do we need to handle the "connectable state" part of the WID?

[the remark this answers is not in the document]
alexsven: No, they are already in a connectable state, should I add a comment stating that then?

autopts/wid/csis.py
alexsven: That's not entirely true. I tested with setting the size to either 2 or 3 and both works. As far as I can see the set size was 3 to begin with. So it seems the notification will be sent regardless if the value is changed or not, as long as the set-function is called

Thalley: Hmm, that could be a bug in the IUT, as it does

```
	if (svc_inst->set_size == size && svc_inst->rank == rank) {
		LOG_DBG("Set size %u and rank %u is already set", size, rank);
		return -EALREADY;
	}
```

But maybe the set size was the same, but the rank changed? Not really sure how to handle that case; might need an erratum for that actually. 

However even if that was the case, it's still an unreliable implementation to hardcore the size and rank in these WID command handlers and/or setting the new size and rank without knowing the existing one IMO

autopts/wid/csis.py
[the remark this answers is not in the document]
Thalley: Please do :) I'll leave it up to you whether to decide if e.g. wid_21 should just call wid_20, or if you want to keep them separate like this

autopts/wid/csis.py
alexsven: Changed to setting set_size to 2 as default when initing CSIS, and incrementing by 1 if correct wid is called

autopts/wid/csis.py
```diff
@@ -53,6 +54,40 @@ def hdl_wid_7(_: WIDParams):
...
+    # The same wid is used for two different test cases, with different descriptions, so we check the description.
+    if 'Set Identity Resolving Key' in params.description:
+        # Generate 16 random characters for SIRK
+        sirk = secrets.token_hex(16)
+        btp.csis_set_sirk(sirk)
+
+    elif 'Coordinated Set Size' in params.description:
+        # Increase the Coordinated Set Size by 1
+        stack = get_stack()
+        stack.csis.set_size = stack.csis.set_size + 1
```
Thalley:
```suggestion
        stack.csis.set_size += 1
```

Should also work

autopts/wid/csis.py
```diff
@@ -53,6 +54,40 @@ def hdl_wid_7(_: WIDParams):
...
+        btp.csis_set_set_size(stack.csis.set_size, 1)
+
+    else:
+        log(f"hdl_wid_20: Unknown description: {params.description}")
```
Thalley:
```suggestion
        logging.error(f"hdl_wid_20: Unknown description: {params.description}")
```

Suggest to log this as an error

autopts/wid/csis.py
```diff
@@ -53,6 +54,40 @@ def hdl_wid_7(_: WIDParams):
...
 
 
+def hdl_wid_20(params: WIDParams):
+    """
+        Please update the characteristic Set Identity Resolving Key value with different data and send a notification.
+        Please update the characteristic Coordinated Set Size value with different data and send a notification.
+    """
+    # The same wid is used for two different test cases, with different descriptions, so we check the description.
+    if 'Set Identity Resolving Key' in params.description:
+        # Generate 16 random characters for SIRK
```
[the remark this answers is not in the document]
alexsven: Will change to `sirk = secrets.token_hex(8)`

autopts/ptsprojects/stack/layers/__init__.py
```diff
@@ -21,6 +21,7 @@
 from .ccp import *  # noqa: F403 # used in many files : TODO import directly in files not with *
 from .core import *  # noqa: F403 # used in many files : TODO import directly in files not with *
 from .csip import *  # noqa: F403 # used in many files : TODO import directly in files not with *
+from .csis import *  # noqa: F403 # used in many files : TODO import directly in files not with *
```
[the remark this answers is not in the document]
alexsven: Not sure I understood that one, this is the same as all the others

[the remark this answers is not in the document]
Thalley: Yeah, I think this entire file is just a bad approach :P

Thalley (approved):

sjanc (approved):

PR #1602, opened 2025-10-31
New SVC data set function and added tests for TMAP

koffes:
TMAP/BMR/DDI/BV-01-C
TMAP/UMR/DDI/BV-01-C
TMAP/BMS/DDI/BV-01-C

OCT-3454

### DIFF
--- a/autopts/ptsprojects/zephyr/hap.py
+++ b/autopts/ptsprojects/zephyr/hap.py
@@ -15,7 +15,6 @@
 
 """HAP test cases"""
 import struct
-from enum import IntEnum
 
 from autopts.client import get_unique_name
 from autopts.ptsprojects.stack import get_stack
@@ -23,8 +22,9 @@ from autopts.ptsprojects.testcase import TestFunc
 from autopts.ptsprojects.zephyr.hap_wid import hap_wid_hdl
 from autopts.ptsprojects.zephyr.ztestcase import ZTestCase
 from autopts.pybtp import btp
+from autopts.pybtp.btp.gap import gap_set_uuid16_svc_data
 from autopts.pybtp.defs import HAS_TSPX_available_presets_indices, HAS_TSPX_unavailable_presets_indices
-from autopts.pybtp.types import Addr, AdType, Context
+from autopts.pybtp.types import UUID, Addr, AdType, BAPAnnouncement, CAPAnnouncement, Context
 
 # Options aligned with the overlay-le-audio.conf options
 BTP_HAP_HA_OPTS_DEFAULT = (btp.defs.HAP_HA_OPT_PRESETS_DYNAMIC |
@@ -32,14 +32,6 @@ BTP_HAP_HA_OPTS_DEFAULT = (btp.defs.HAP_HA_OPT_PRESETS_DYNAMIC |
 BTP_HAP_HA_OPTS_BINAURAL = (btp.defs.HAP_HA_OPT_PRESETS_INDEPENDENT |
                             btp.defs.HAP_HA_OPT_PRESETS_DYNAMIC |
                             btp.defs.HAP_HA_OPT_PRESETS_WRITABLE)
-
-
-class Uuid(IntEnum):
-    ASCS = 0x184E
-    BASS = 0x184F
-    PACS = 0x1850
-    BAAS = 0x1852
-    CAS = 0x1853
 
 
 def set_pixits(ptses):
@@ -84,16 +76,7 @@ def set_pixits(ptses):
     pts.set_pixit("HAP", "TSPX_num_presets", str(num_presets))
 
 
-def announcements(advData, targeted):
-    """
-        CAP General/Targeted Announcement
-    """
-    advData[AdType.uuid16_svc_data] = [struct.pack('<HB', Uuid.CAS, 1 if targeted else 0)]
-    """
-        BAP General/Targeted Announcement
-    """
-    advData[AdType.uuid16_svc_data] += [struct.pack('<HBHHB', Uuid.ASCS, 1 if targeted else 0,
-        Context.LIVE | Context.MEDIA, Context.LIVE, 0)]
+def set_member_rsi(advData, targeted):
     """
         RSI
     """
@@ -152,7 +135,10 @@ def test_cases(ptses):
     ]
 
     adv_conditions = [
-        TestFunc(announcements, advData, True),
+        TestFunc(gap_set_uuid16_svc_data, advData, UUID.CAS, struct.pack('<B', CAPAnnouncement.TARGETED)),
+        TestFunc(gap_set_uuid16_svc_data, advData, UUID.ASCS, struct.pack('<BHHB', BAPAnnouncement.TARGETED,
+                Context.LIVE | Context.MEDIA, Context.LIVE, 0)),
+        TestFunc(set_member_rsi, advData, True),
         TestFunc(btp.gap_set_extended_advertising_on),
         TestFunc(btp.gap_adv_ind_on, ad=advData),
     ]
--- a/autopts/ptsprojects/zephyr/tmap.py
+++ b/autopts/ptsprojects/zephyr/tmap.py
@@ -21,7 +21,8 @@ from autopts.ptsprojects.testcase import TestFunc
 from autopts.ptsprojects.zephyr.tmap_wid import tmap_wid_hdl
 from autopts.ptsprojects.zephyr.ztestcase import ZTestCase
 from autopts.pybtp import btp
-from autopts.pybtp.types import UUID, Addr, AdType, Context
+from autopts.pybtp.btp.gap import gap_set_uuid16_svc_data
+from autopts.pybtp.types import UUID, Addr, AdType, BAPAnnouncement, CAPAnnouncement, Context, TMAPRole
 
 
 def set_pixits(ptses):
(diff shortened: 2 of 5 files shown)


autopts/ptsprojects/zephyr/tmap.py
```diff
@@ -52,18 +52,37 @@ def set_pixits(ptses):
...
+    advData[AdType.uuid16_svc_data] = []
+
+
+def announcements_cap(advData, targeted):
+
     """
         CAS General/Targeted Announcement
     """
-    advData[AdType.uuid16_svc_data] = [struct.pack('<HB', int(UUID.CAS, 16), 1 if targeted else 0)]
+    advData[AdType.uuid16_svc_data] += [struct.pack('<HB', int(UUID.CAS, 16), 1 if targeted else 0)]
```
Thalley: We shouldn't simply append data like this. 

If `announcements_cap` is called twice like
```
announcements_cap(advData, false)
announcements_cap(advData, true)
```

Then PTS would always parse the advertising data with targeted=false first, which could be pretty hard to debug later on. Ideally the code would parse `advData[AdType.uuid16_svc_data]` to check if the LTV already exists, and then update it, instead of always just appending new data. 

This also applies for the other 2 functions here.

autopts/pybtp/types.py
```diff
@@ -125,6 +126,7 @@ class UUID:
     CPF = '2904'
     CAF = '2905'
     CSF = '2B29'
+    CAS = '1853'
```
Thalley: CAS already exists on line 119

autopts/pybtp/types.py
```diff
@@ -502,3 +504,12 @@ class BIGEncryption:
...
     BAD_CODE = 0x03
+
+
+class TMAPRole(IntEnum):
+    CALL_GATEWAY = 0x01
+    CALL_TERMINAL = 0x02
+    UNICAST_MEDIA_SENDER = 0x04
+    UNICAST_MEDIA_RECEIVER = 0x08
+    BROADCAST_MEDIA_SENDER = 0x10
+    BROADCAST_MEDIA_RECEIVER = 0x20
```
Thalley: Given that this is a bitfield, and that we define and use `BIT` function in `defs.py`, we should consider doing the same here, to make it clearer that it is a bitfield. 

Additionally, defining this as an `IntEnum` would actually prevent it from being a bitfield, so I think that is the wrong approach here

koffes: I re-did this part.

autopts/ptsprojects/zephyr/tmap.py
koffes: I have a new implementation ready with tests. If OK, I will update the existing files with the new API.

autopts/pybtp/btp/gap.py
```diff
@@ -1700,3 +1700,29 @@ def gap_set_broadcast_code(broadcast_code):
...
+    for entry in advData[AdType.uuid16_svc_data]:
+        existing_uuid = struct.unpack_from('<H', entry[:2])[0]
+        if existing_uuid == UUID:
+            advData[AdType.uuid16_svc_data].remove(entry)
+            break
+    if service_data is None:
+        advData[AdType.uuid16_svc_data] += [struct.pack('<H', UUID)]
+    else:
+        new_entry = struct.pack('<H', UUID) + service_data
+        advData[AdType.uuid16_svc_data] += [new_entry]
```
Thalley: Could this be


```suggestion
    advData[AdType.uuid16_svc_data] += [struct.pack('<H', UUID)]
    if service_data is not None:
        advData[AdType.uuid16_svc_data] += service_data
```

autopts/ptsprojects/zephyr/tmap.py
koffes: New method implemented

autopts/pybtp/btp/gap.py
koffes: That will AFAIK not work, as that will add the service data as a separate/new element in the service data.

koffes:
Changes done. TBD: Split into separate commits.

autopts/ptsprojects/zephyr/hap.py
```diff
@@ -84,16 +76,7 @@ def set_pixits(ptses):
...
-    """
-        CAP General/Targeted Announcement
-    """
-    advData[AdType.uuid16_svc_data] = [struct.pack('<HB', Uuid.CAS, 1 if targeted else 0)]
-    """
-        BAP General/Targeted Announcement
-    """
-    advData[AdType.uuid16_svc_data] += [struct.pack('<HBHHB', Uuid.ASCS, 1 if targeted else 0,
-        Context.LIVE | Context.MEDIA, Context.LIVE, 0)]
+def member_rsi_set(advData, targeted):
```
Thalley:
```suggestion
def set_member_rsi(advData, targeted):
```

Suggest to follow the naming scheme of the function above (`set_pixits`)

Thalley: Although I do see that it's currently follow the naming scheme of `gap_uuid16_svc_data_set`. Hmm; not sure which approach is the "pythonic" way

autopts/pybtp/btp/gap.py
```diff
@@ -1700,3 +1700,37 @@ def gap_set_broadcast_code(broadcast_code):
     logging.debug("%s %r", gap_set_broadcast_code.__name__, broadcast_code)
     stack = get_stack()
     stack.gap.big_broadcast_code = broadcast_code
+
+
+def gap_uuid16_svc_data_set(advData, UUID, service_data=None):
```
```suggestion
def gap_uuid16_svc_data_set(advData, uuid, service_data=None):
```

AutoPTS seems to use `uuid` over `UUID`. I think that makes sense since values in ALL_CAPS in Python is the Python (sigh...) way of denoting global constants.

autopts/pybtp/btp/gap.py
```diff
@@ -1700,3 +1700,37 @@ def gap_set_broadcast_code(broadcast_code):
...
+    """Update service data. Will overwrite existing UUID16 service data if present"""
+    # Convert UUID to integer if it's a string
+    if isinstance(UUID, str):
+        try:
+            uuid_int = int(UUID, 16)  # Convert hex string to int
+        except ValueError as err:
+            raise ValueError("Invalid UUID string format") from err
+    else:
+        uuid_int = int(UUID)  # Ensure it's an integer
```
Thalley: Not sure I understand this. 

If `UUID` is a string, we attempt to cast it to an int, with error handling. But if it's anything else (e.g. a boolean, an object, `None`, or whatever) then we attempt to cast it without any error handling. Should the `if isinstance(UUID, str):` be changed to `if not isinstance(UUID, int):` ?

autopts/pybtp/types.py
```diff
@@ -502,3 +503,22 @@ class BIGEncryption:
...
+    BROADCAST_MEDIA_SENDER = defs.BIT(4)
+    BROADCAST_MEDIA_RECEIVER = defs.BIT(5)
+
+
+class BAP_Announcement:
+    GENERAL = 0x0
+    TARGETED = 0x1
+
+
+class CAP_Announcement:
```
Thalley: Not sure why CI isn't catching this, but PEP8 (https://peps.python.org/pep-0008/#class-names) requires the class names are CapWords, so I guess these should be


```suggestion
class BAPAnnouncement:
    GENERAL = 0x0
    TARGETED = 0x1


class CAPAnnouncement:
```

Or perhaps


```suggestion
class BapAnnouncement:
    GENERAL = 0x0
    TARGETED = 0x1


class CapAnnouncement:
```

Thalley: @JakubMarciniak93 any idea why this wasn't caught?

JakubMarciniak93: @Thalley 
I think the problem is in the pyproject.toml "select" section for ruff.
https://github.com/auto-pts/auto-pts/blob/master/pyproject.toml
Currently, the following rules are checked:
select = ["F", "E", "W", "I", "UP", "ISC", "TID", "PT", "A", "PERF", "B"]
I think that additional validation would need to be added:
"N", but I am not sure know...quick glance

JakubMarciniak93: exactly, I think N801
Example:
pybtp\types.py:391:7: N801 Class name `LC3_LTV_Config_Type` should use CapWords convention
    |
390 | # Assigned Numbers (2023), 6.12.5 Codec_Specific_Configuration LTV structures
391 | class LC3_LTV_Config_Type:
    |       ^^^^^^^^^^^^^^^^^^^ N801
392 |     SAMPLING_FREQ = 0x01
393 |     FRAME_DURATION = 0x02

Thalley: Thanks. I assume that it isn't enabled because we haven't renamed the offending classes yet?

JakubMarciniak93: @Thalley 
That's right, the final plan was to rule:
["T", "C", "D", "U", "N", "F", "E", "W", "I", "UP", "ISC", "TID", "PT", "A", "PERF", "B"]
The PR with all the changes would be too extensive to revise in detail.
the plan was to divide everything into several PRs. Priorities :)

Edit:
and this what now is ignore
ignore = ["F841"]
it is big issue

test/unittests.py
```diff
@@ -250,6 +255,36 @@ def test_generate_stats(self):
...
+        with pytest.raises(ValueError, match="Invalid UUID16 value"):
+            gap_set_uuid16_svc_data(advData, -1, None)
+        with pytest.raises(TypeError):
+            gap_set_uuid16_svc_data(advData, 0xAAAA, 4)
+
+        # Test valid input
+        try:
+            gap_set_uuid16_svc_data(advData, 0x0000, None)
+            gap_set_uuid16_svc_data(advData, 0xFFFF, None)
+            gap_set_uuid16_svc_data(advData, 0xAABB, struct.pack('<B', 1))
```
Thalley, line 272: Since you added the support to cast to int, these could possibly be expanded. 

Additionally, we may want to consider just removing the attempt to cast, as you can cast most things to ints, including Boolean, floats, bytearrays, etc. 

Image

Not sure if we want to support providing `122.5` as a valid UUID :D

sjanc (approved):

PR #1603, opened 2025-11-06
Fix VCS tests with Zephyr

Frodevan:
Zephyr does not allow for changing the volume after registering VCS without also setting the volume persisted flag. Due to this, VCS needs to be registered with the host after the initial volume is set by the appropriate wid.

This PR is associated with https://github.com/zephyrproject-rtos/zephyr/pull/99025


autopts/pybtp/btp/vcs.py
```diff
@@ -32,7 +32,9 @@
     "mute": (defs.BTP_SERVICE_ID_VCS, defs.BTP_VCS_CMD_MUTE,
                CONTROLLER_INDEX, ""),
     "unmute": (defs.BTP_SERVICE_ID_VCS, defs.BTP_VCS_CMD_UNMUTE,
-               CONTROLLER_INDEX, "")
+               CONTROLLER_INDEX, ""),
+    "register": (defs.BTP_SERVICE_ID_VCS, defs.BTP_VCS_CMD_REGISTER,
+               CONTROLLER_INDEX)
```
Thalley:
```suggestion
               CONTROLLER_INDEX),
```

Suggest to end with a `,` so that adding another entry won't affect existing entries

autopts/pybtp/btp/vcs.py
```diff
@@ -67,6 +69,24 @@ def vcs_set_vol(vol):
...
+    logging.debug("%s", vcs_register.__name__)
+
+    iutctl = get_iut()
+
+    if isinstance(step, str):
+        step = int(step)
+    if not isinstance(mute, int):
+        mute = int(mute)
+    if isinstance(vol, str):
+        vol = int(vol)
```
Thalley, line 83: What is the reasoning for support e.g. step as both `int` and `str`?

autopts/wid/vcs.py
```diff
@@ -29,30 +29,54 @@ def vcs_wid_hdl(wid, description, test_case_name):
...
     btp.gap_adv_ind_on(ad=stack.gap.ad)
+
+    delayed_vcs_reg_tc = [
+        'VCS/SR/CP/BV-01-C',
+        'VCS/SR/CP/BV-02-C',
+        'VCS/SR/CP/BV-03-C',
+        'VCS/SR/CP/BV-04-C',
+    ]
+
+    if params.test_case_name not in delayed_vcs_reg_tc:
```
Thalley, line 44: I wonder if there's a way to avoid the `delayed_vcs_reg_tc` list, and then just always register VCS immediately? Registering VCS immediately now won't set the flag that volume has changed, so we should be able to do that always, right?

autopts/pybtp/defs.py
```diff
@@ -583,6 +583,7 @@ def BIT(bit):
 BTP_VCS_CMD_VOL_DOWN = 0x04
 BTP_VCS_CMD_MUTE = 0x05
 BTP_VCS_CMD_UNMUTE = 0x06
+BTP_VCS_CMD_REGISTER = 0x07
```
Thalley, line 586: The command needs to be described in `btp_vcs.txt`

autopts/wid/vcs.py
Frodevan: The problem is that at this point we don't know what initial settings will be requested. The requested volume and mute settings is first specified in the other WIDs.

autopts/pybtp/btp/vcs.py
Frodevan: That's mainly copy-pasted from vcs_set_vol, just in case someone else decides to use a str when calling this at some point in the future.

autopts/wid/vcs.py
Thalley: > The requested volume and mute settings comes with the other WIDs.

In the tests where that's true, do we still need to have the volume flag unset? If the test requires a different volume setting than what any IXIT values requests, then it sort of sounds like a test spec / PTS issue if we are requested to set a specific volume value without setting the volume flag

autopts/pybtp/defs.py
Frodevan: done

autopts/pybtp/btp/vcs.py
Frodevan: Done

autopts/wid/vcs.py
Frodevan: The following are exact quotes from the WIDs:
wid 1: `Please configure IUT's volume setting to (2 x Step Size), and persisted setting to false (0).`
wid 2: `Please configure IUT's volume setting to (2 x Step Size), persisted setting to false (0) and Mute State to Muted.`
wid 3: `Please configure IUT's volume setting to 255 - (2 x Step Size), and persisted setting to false (0).`
wid 4: `Please configure IUT's volume setting to 255 - (2 x Step Size), Mute State to Muted, and persisted setting to false (0).`

So technically only two of the four specifically mention the mute state, but all demand the persisted flag to be unset at the start of the test. If this has merit in the specs, or if it could be waivered, I haven't looked too much into.

autopts/pybtp/btp/vcs.py
Thalley: Sure - I don't know what is the norm for autopts, but this doesn't really hurt. 

Personally I'd prefer adding type hinting and only use a single type for the parameters though

autopts/wid/vcs.py
Thalley: Alright, I guess we have to do it this way then :) The alternative solution would be to raise a PTS support ticket, as the WIDs are kind of weird, but can be handled (as this PR shows)

doc/btp_vcs.txt
```diff
@@ -64,3 +64,15 @@ Commands and responses:
                 This command is used for setting mute value to "unmute" in VCS.
 
                 In case of an error, the error response will be returned.
+
+        Opcode 0x07 - Register Volume Control Service
+                Controller Index:       <controller id>
+                Command parameters:     Step (1 octet)
+                                        Mute (1 octet)
```
Thalley: Suggest to define possible values for `mute` (0 for unmuted or 1 for muted)

sjanc: I'd suggest to make mute flag and have flags field instead

Thalley: Good idea; easier to expand

Frodevan: Done

autopts/pybtp/btp/vcs.py
```diff
@@ -67,6 +69,24 @@ def vcs_set_vol(vol):
...
 
 
+def vcs_register(step, mute, vol):
+    logging.debug("%s", vcs_register.__name__)
+
+    iutctl = get_iut()
+
+    if isinstance(step, str):
+        step = int(step)
+    if not isinstance(mute, int):
```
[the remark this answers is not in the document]
Frodevan: No longer relevant

doc/btp_vcs.txt
```diff
@@ -64,3 +64,15 @@ Commands and responses:
...
+                                        Volume (1 octet)
+                Response parameters:    <none>
+
```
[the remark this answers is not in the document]
Frodevan: Fixed

autopts/pybtp/defs.py
```diff
@@ -583,6 +583,9 @@ def BIT(bit):
...
+
+VCS_REGISTER_FLAG_MUTED = 0x01
```
Thalley: Given that we probably want the flags to be a bitfield, I'd suggest to do

```suggestion
VCS_REGISTER_FLAG_MUTED = BIT(0)
```

Frodevan: Changed

PR #1615, opened 2025-11-17
wid: Fix for BASS/SR/SPE/BI-06-C

omkar3141:
BIS sync in WID 100 handler for BASS/SR/SPE/BI-06-C test case.

The BIS sync is not needed for this specific test. The WID 100 handler should only handle PA sync. This commit adds conditional logic to skip BIS sync waiting when running BI-06-C, preventing timeout issues in the automation framework until PTS provides a different WID for this test.

### DIFF
--- a/autopts/wid/bass.py
+++ b/autopts/wid/bass.py
@@ -29,7 +29,7 @@ def bass_wid_hdl(wid, description, test_case_name):
     return generic_wid_hdl(wid, description, test_case_name, [__name__])
 
 
-def hdl_wid_100(_: WIDParams):
+def hdl_wid_100(param: WIDParams):
     """
     Please synchronize with Broadcast ISO request.
     """
@@ -51,6 +51,15 @@ def hdl_wid_100(_: WIDParams):
     if ev is None:
         log(f'BIS not found, broadcast ID {broadcast_id}')
         return False
+
+    # Workaround for BASS/SR/SPE/BI-06-C: This test case validates invalid length handling
+    # and doesn't require PA or BIS sync. However, PTS sends Add Source requesting only PA sync,
+    # then sends WID 100 (which expects BIS sync per its description "synchronize with Broadcast
+    # ISO"). This mismatch causes the BIS sync wait to timeout. Skip BIS sync wait for this test
+    # case until PTS is fixed (e.g., by using a different WID, requesting BIS sync in Add Source,
+    # or not sending WID 100 for this test).
+    if param.test_case_name in ['BASS/SR/SPE/BI-06-C']:
+        return True
 
     bis_id = ev['bis_id']
     broadcast_id = ev['broadcast_id']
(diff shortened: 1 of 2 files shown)


autopts/wid/bass.py
```diff
@@ -52,6 +52,12 @@ def hdl_wid_100(_: WIDParams):
         log(f'BIS not found, broadcast ID {broadcast_id}')
         return False
 
+    # In BASS/SR/SPE/BI-06-C, the BIS sync is not needed. The WID handler 100 should only be
+    # concerned about the PA sync (and not the BIS sync, there should be seperate handler for that
+    # or this BASS/SR/SPE/BI-06-C test case should use different WID handler)
```
Thalley: Not sure I agree that this description is fully correct, especially the 

> The WID handler 100 should only be concerned about the PA sync (and not the BIS sync

part, which is actually the opposite; WID_100 should care about the BIS sync as per the description. 

What happens is that the test case actually doesn't care about the PA nor the BIS sync. Basically it just requires that

1. The Lower Tester (PTS) shall discover the control point handle
2. The Lower Tester shall perform an Add Source with valid values and await a receive state notification
3. The Lower Tester then performs several invalid writes and expect error responses. 

The issue is that in step 2, PTS sends an Add Source request where it requests a sync to the PA (this triggers 2 notifications; one for the initial state of the receive state, and one for the PA sync update when the IUT syncs, but this is seemingly handled well by PTS), but without a request to sync to the BIG. However PTS then sends the WID 100, which states 

> Please synchronize with Broadcast ISO request.

However that does not match the Add Source request from PTS, so there is arguably a mismatch between the Add Source from PTS and the WID handler. 

The solution (workaround) is just one of several possible solutions:

1. PTS should use a different WID if it just wants us to sync to the PA without the BIG sync
2. PTS should add the request to sync to the BIG in the Add Source operation
3. AutoPTS should instruct the IUT to sync to the BIG based on receiving WID 100, regardless of whether it has been requested by the broadcast assistant (currently it seems that AutoPTS assumes that when we receive WID 100, we have been requested to sync to the BIG by something else)
4. The workaround here simply skipping the call to `wait_bis_synced_ev` for this specific (and possibly other) test cases. 
5. PTS should not request any PA or BIG sync, nor send WID 100 at all (not needed for the test). 

A mix of the above could also be implemented; AutoPTS should, IMO, in any case never just blindly expect `wait_bis_synced_ev` without knowing if the IUT has been instructed to sync to the BIG

alexsven:
> > > Nevermind - Since the workaround here in this PR is for a specific test case (workaround based on the test case name), it will of course not pass any other tests. @omkar3141 Could you try locally to see if this workaround also works for the other failing BI tests?
> > 
> > 
> > I will try.
> 
> Please coordinate with @alexsven as I think you may both be modifying this code :)

I am not, at the moment

PR #1616, opened 2025-11-18
zephyr: mesh: fix model rx payload length

alxelax:
Commit fixes rx model event payload length.
Since model might receive full access payload
then 1 byte length is not sufficient. It is 2 bytes now.


sjanc:
in general this sort of changes should be done with v2 command/event (as this will break any user that is not updated)

PR #1617, opened 2025-11-18
btp: Add new ASCS events for CIS connection state

Thalley:
Added 2 new events to indicate the state of the CIS associated with an ASE, as that cannot be determined directly from the ASE state.

The "connected" event is the only one used so far, and is used to wait for the even when we expect the CIS to be connected.


autopts/pybtp/btp/ascs.py
```diff
@@ -269,8 +269,46 @@ def ascs_ev_ase_state_changed_(ascs, data, data_len):
                         (addr_type, addr, ase_id, state))
 
 
+def ascs_ev_cis_connected_(ascs, data, data_len):
+    logging.debug("%s %r", ascs_ev_cis_connected_.__name__, data)
```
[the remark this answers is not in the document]
Thalley: Do we have any guidelines or rules of this? The `ruff` extension in VSCode automatically applies `"` over `'`

autopts/pybtp/btp/ascs.py
```diff
@@ -269,8 +269,46 @@ def ascs_ev_ase_state_changed_(ascs, data, data_len):
...
+    fmt = "<B6sBB"
+    header_size = struct.calcsize(fmt)
+    if len(data) < header_size:
+        raise BTPError("Invalid data length")
+
+    addr_type, addr, ase_id, cis_id = struct.unpack_from(fmt, data)
+
+    addr = binascii.hexlify(addr[::-1]).lower().decode("utf-8")
+
+    logging.debug(f"CIS connected: ase_id {ase_id}, cis_id {cis_id}")
```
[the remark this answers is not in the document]
Thalley: Could be done, but not done for any other event

autopts/pybtp/btp/ascs.py
```diff
@@ -269,8 +269,46 @@ def ascs_ev_ase_state_changed_(ascs, data, data_len):
...
+
+    ascs.event_received(defs.BTP_ASCS_EV_CIS_CONNECTED, (addr_type, addr, ase_id, cis_id))
+
+
+def ascs_ev_cis_disconnected_(ascs, data, data_len):
+    logging.debug("%s %r", ascs_ev_cis_disconnected_.__name__, data)
```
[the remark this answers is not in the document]
Thalley: Ditto

autopts/pybtp/btp/ascs.py
```diff
@@ -269,8 +269,46 @@ def ascs_ev_ase_state_changed_(ascs, data, data_len):
...
+    fmt = "<B6sBBB"
+    header_size = struct.calcsize(fmt)
+    if len(data) < header_size:
+        raise BTPError("Invalid data length")
+
+    addr_type, addr, ase_id, cis_id, reason = struct.unpack_from(fmt, data)
+
+    addr = binascii.hexlify(addr[::-1]).lower().decode("utf-8")
+
+    logging.debug(f"CIS disconnected: ase_id {ase_id}, cis_id {cis_id}, reason {reason}")
```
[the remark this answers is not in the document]
Thalley: Ditto

autopts/wid/bap.py
```diff
@@ -1447,6 +1447,15 @@ def hdl_wid_311(params: WIDParams):
...
         btp.ascs_receiver_start_ready(config.ase_id, config.addr_type, config.addr)
 
+        ev = stack.ascs.wait_ascs_cis_connected_ev(
+            config.addr_type,
+            config.addr,
+            config.ase_id,
+            config.cis_id,
+        )
+        if ev is None:
+            return False
```
Thalley: @sjanc should this part be added in a separate PR, since it technically break nearly all BAP tests until https://github.com/zephyrproject-rtos/zephyr/pull/99597 has been merged (or until other platform implement the events)?

autopts/ptsprojects/stack/layers/ascs.py
```diff
@@ -46,3 +48,31 @@ def wait_ascs_ase_state_changed_ev(self, addr_type, addr, ase_id, state, timeout
...
+            lambda _addr_type, _addr, _ase_id, _cis_id, *_: (addr_type, addr, ase_id, cis_id)
+            == (_addr_type, _addr, _ase_id, _cis_id),
+            timeout,
+            remove,
+        )
+
+    def wait_ascs_cis_disconnected_ev(
+        self, addr_type, addr, ase_id, cis_id, reason=None, timeout=10, remove=True
+    ):
+        def _matches(ev_addr_type, ev_addr, ev_ase_id, ev_cis_id, ev_reason, *_):
```
Thalley, line 64: Applied a function here, as doing the optional check on `reason` was pretty hard to easily do in a `lambda`

autopts/wid/bap.py
sjanc: maybe this could be conditional for time being (declared by stack)?

also, why is this needed here?  isn't ases state change validation later on not enough?

Thalley: I think I can move this from this commit, to the commit that really needs it from github.com/auto-pts/auto-pts/pull/1618, so I will just do that. That would simply mean that the events from the IUT will go unhandled, but that should be fine

> also, why is this needed here? isn't ases state change validation later on not enough?

The CIS state changes does not trigger a change in the ASEs. A CIS can be connected while in the QoS Configured state or the Enabling state, and it won't affect the ASE state when it happens.

autopts/pybtp/btp/ascs.py
```diff
@@ -269,8 +269,46 @@ def ascs_ev_ase_state_changed_(ascs, data, data_len):
                         (addr_type, addr, ase_id, state))
 
 
+def ascs_ev_cis_connected_(ascs, data, data_len):
+    logging.debug("%s %r", ascs_ev_cis_connected_.__name__, data)
+
+    fmt = "<B6sBB"
+    header_size = struct.calcsize(fmt)
+    if len(data) < header_size:
```
koffes: Should the check be

```suggestion
    if len(data) != header_size:
``` ?

Comment (koffes) [reply autopts/pybtp/btp/ascs.py]: Same further down

**Review by koffes — APPROVED** [review]

File: autopts/pybtp/btp/ascs.py (hunk 7 above)
Comment (Thalley) [reply autopts/pybtp/btp/ascs.py]: Hmm, I don't actually know. This was just copied form the `ascs_ev_ase_state_changed_` above. 

I'm not sure if `header_size` refers to the size value in the header, or the size of the header. 

@sjanc ?

Comment (sjanc) [reply autopts/pybtp/btp/ascs.py]: I'm fine either way,  however ```!=``` might allow to catch any discrepancies in BTP implementation sooner

File: doc/btp_ascs.txt [hunk 8]
```diff
@@ -241,3 +241,26 @@ Events:
...
 
                 This event indicates that ASE state has been changed.
+
+        Opcode 0x83 - CIS connected
+
+                Controller Index:       <controller id>
+                Event parameters:       Address_Type (1 octet)
+                                        Address (6 octets)
+                                        ASE_ID (1 octet)
+                                        CIS_ID (1 octet)
```
sjanc, line 251: I wonder if maybe we should start adding status fields to connected events to allow IUT to fail immediately instead of timing out on missing event in case of error?

(not going to force this since afair none of our events has this, just something to think about)

autopts/pybtp/btp/ascs.py
Thalley: Okay, so `header_size` is the size field of the header, and not the size of the header? :D

doc/btp_ascs.txt
Thalley: That's an interesting idea. For example, this event is triggered by the CIS established event which does have a `Status` field in HCI, so we can add that for something like this, but other events like advertising reports there are no status field. However I think it's probably possible for the majority of events (at least the ones that give the IUT a status from e.g. HCI).

autopts/pybtp/btp/ascs.py
Thalley: Updated

doc/btp_ascs.txt
sjanc: in long term we can think ab out this, for now I guess it is ok to rely on timeout like we do for other

PR #1618, opened 2025-11-18
wid: BAP: Support connecting CIS in qos or enabled state 

Thalley:
Some test cases (roughly 30 tests) require that we connect
the CIS of an ASE (or 2) in either the QoS configured state,
or the enabled state.

Depends on https://github.com/zephyrproject-rtos/zephyr/pull/99597

### DIFF
--- a/autopts/wid/bap.py
+++ b/autopts/wid/bap.py
@@ -24,7 +24,13 @@ from typing import Final
 from autopts.ptsprojects.stack import WildCard, get_stack
 from autopts.ptsprojects.testcase import MMI
 from autopts.pybtp import btp, defs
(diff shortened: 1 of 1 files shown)


koffes (approved):

Thalley:
This PR should fix all cases from https://github.com/auto-pts/auto-pts/blob/master/errata/zephyr.yaml#L5-L37 with the exception of 

- BAP/UCL/STR/BV-545-C
- BAP/UCL/STR/BV-548-C
- BAP/UCL/STR/BV-551-C

That unfortunately still fail due to an HCI timeout on PTSs side

codecoup-tester:
Job failed at setup

PR #1623, opened 2025-11-20
wid: add support for wid 145 and 167

PavelVPV:
These 2 wids are introduced in GATT/SR/GAW/BI-41-C test.

wid 145 requests handle of a UUID of a long characteristic. wid 167 requests to remove the characteristic by handle requested in wid 145.

Corresponding Zephyr PR: https://github.com/zephyrproject-rtos/zephyr/pull/99731


m-alperen-sener (changes requested):
Seems OK to me, but we need to btp_gatt_server  doc.

Also can be merged before updating PTS to 8.11.0 since these are new WIDs coming in and not changing the behavior of existing ones.

m-alperen-sener (approved):

PR #1625, opened 2025-11-20
Fixes for pts 8 11

m-alperen-sener:
- Removing the IXITs (TSPX) that are removed from PTS
-   wid: gap: add wid 405
    - Adding wid 405 for test case GAP/PRIV/CONN/BV-10-C.

### DIFF
--- a/autopts/wid/gap.py
+++ b/autopts/wid/gap.py
@@ -1812,6 +1812,19 @@ def hdl_wid_403(_: WIDParams):
     return True
 
 
+def hdl_wid_405(_: WIDParams):
+    """
+    Please change private addresse and enter Undirected Connectable Mode.
+    """
+
+    # We don't need to do anything, RPA will change automatically after RPA
+    # timeout. We should not change the RPA by any other means because this
+    # WID is used by test cases that checks the RPA change according to the
+    # RPA timeout. WID 405 exists only for PTS itself.
+
+    return True
+
+
 def hdl_wid_406(_: WIDParams):
     # Please enter General Connectable Mode using private addresses.
     # Note: IUT is expected to do directed advertising with target address set


autopts/wid/gap.py
```diff
@@ -1994,10 +2005,18 @@ def hdl_wid_2004(params: WIDParams):
...
     """
         Please prepare IUT into a connectable mode.
     """
+
+    # GAP/PRIV/CONN/BV-10-C is already in connectable mode due to
+    # previous WIDs, in particular WID 90 orders IUT to enter connectable mode.
+    # We can remove this exception once we keep device connectable mode in
+    # stack gap.
+    if params.test_case_name in ['GAP/PRIV/CONN/BV-10-C']:
+        return True
```
Thalley: We do actually have the option of doing something more generic like 
```
    if stack.gap.current_settings_get(
            gap_settings_btp2txt[defs.GAP_SETTINGS_ADVERTISING]):
        return True
```

To check the current advertising state

autopts/wid/gap.py
```diff
@@ -1812,6 +1812,17 @@ def hdl_wid_403(_: WIDParams):
     return True
 
 
+def hdl_wid_405(_: WIDParams):
+    # Please change private addresse and enter Undirected Connectable Mode.
```
Thalley:
```suggestion
    ```
        Please change private addresse and enter Undirected Connectable Mode.
    ```
```

Just for the sake of convention

autopts/ptsprojects/zephyr/gatt.py
```diff
@@ -109,10 +109,7 @@ def set_pixits(ptses):
     pts.set_pixit("GATT", "TSPX_security_enabled", "FALSE")
     pts.set_pixit("GATT", "TSPX_delete_link_key", "TRUE")
     pts.set_pixit("GATT", "TSPX_time_guard", "180000")
-    pts.set_pixit("GATT", "TSPX_selected_handle", "0012")
```
Thalley, line 112: I'm curious; how did you determine which IXIT values we can remove? 

I know there are XML files that contain the IXIT values for each test suite, but they don't always seem to be exhaustive

autopts/wid/gap.py
m-alperen-sener: I'll try that.

autopts/ptsprojects/zephyr/gatt.py
m-alperen-sener: PTS tells this actually. New PTS doesn't have those IXITs anymore and fails if you try to set it, at least for GATT.

sjanc: yeap, this is common for all tests since setting/updating nonexisting IXIT results in error from PTS

since those are typically set in preconfig what I usually do is to run single test from each testset to validate this

Thalley: :upside_down_face:

autopts/wid/gap.py
m-alperen-sener: > 

I have check this and it is apperent it is not working properly, I haven't dig deep. both latter [gap_set_conn](https://github.com/auto-pts/auto-pts/blob/master/autopts/pybtp/btp/gap.py#L927) and [gap_adv_ind_on](https://github.com/auto-pts/auto-pts/blob/master/autopts/pybtp/btp/gap.py#L602) has internal check. It seems for my GAP/PRIV/CONN/BV-10-C case it was a PTS racing condition, connection happens before 20001.

Thalley: Hmm, yeah, I have seen a race condition with this approach as well :|

m-alperen-sener:
> btw you validated mynewt changes in PR? if not lets just drop those form this PR

Removal of IXIT values is based on PTS itself, so it does not metter zephyr, mynewt or any other user. Any PTS 8.11.0 user will have the same failure. 

Anyway I ll remove mynewt part, even though it does not need verification at all.

Thalley:
#AutoPTS run zephyr GATT/CL/GAC/BV-01-C

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (1)</summary>GAP     GAP/ADV/BV-01-C                 PASS</details>

sjanc (approved):

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (1)</summary>GATT    GATT/CL/GAC/BV-01-C             PASS</details>

PR #1627, opened 2025-11-21
wid: Exposes Broadcast Receive State

omkar3141:
Exposes Broadcast Receive State by handling WID 116 correctly.


autopts/pybtp/btp/bap.py
```diff
@@ -348,6 +350,33 @@ def bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync,
     bap_command_rsp_succ()
 
 
+def bap_add_broadcast_receive_state(broadcaster_addr_type, broadcaster_addr,
```
Thalley: Needs a definition in the `btp_bap.txt` file

Thalley: I think `add_source` or `add_src` is better to describe what this is doing, i.e. referring to the action defined by the BASS specification

Thalley: Let's avoid previous mistakes done in BTP BAP and make sure to denote which role this is for; i.e. it should be clear whether this is an action done as the BAP Scan Delegator or the BAP Broadcast Assistant. In this case it's the former, but we don't want to confuse this with `BTP_BAP_CMD_ADD_BROADCAST_SRC` which is for the BAP Broadcast Assistant

autopts/pybtp/btp/bap.py
```diff
@@ -348,6 +350,33 @@ def bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync,
...
+                                    advertiser_sid, broadcast_id, big_encryption,
+                                    num_subgroups, subgroups):
+    logging.debug(f"{bap_add_broadcast_receive_state.__name__}")
+
+    if len(subgroups) == 0:
+        raise BTPError('At least one subgroup entry must be provided')
```
Thalley: That is incorrect; you can add a source without any subgroups

autopts/pybtp/btp/bap.py
```diff
@@ -348,6 +350,33 @@ def bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync,
...
+
+    iutctl = get_iut()
+
+    data = bytearray()
```
Thalley: Looking at `bap_add_broadcast_src` this line doesn't seem needed?

autopts/pybtp/btp/bap.py
```diff
@@ -348,6 +350,33 @@ def bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync,
...
+    data += struct.pack('B', num_subgroups)
+    data += subgroups
+
+    iutctl.btp_socket.send(*BAP['add_receive_state'], data=data)
+
+    tuple_data = bap_command_rsp_succ()[0]
+    if len(tuple_data) < 1:
+        raise BTPError('Invalid response length for add_receive_state')
+
+    return tuple_data[0]
```
Thalley, line 381: Should `bap_command_rsp_succ()` be used here instead?

autopts/pybtp/btp/bap.py
```diff
@@ -497,8 +526,8 @@ def bap_ev_codec_cap_found_(bap, data, data_len):
     if len(data) < struct.calcsize(fmt):
         raise BTPError('Invalid data length')
 
-    addr_type, addr, pac_dir, coding_format, frequencies, frame_durations,\
-        octets_per_frame, channel_counts = struct.unpack_from(fmt, data)
+    (addr_type, addr, pac_dir, coding_format, frequencies, frame_durations,
+     octets_per_frame, channel_counts) = struct.unpack_from(fmt, data)
```
Thalley: Unrelated change?

autopts/wid/bass.py
```diff
@@ -214,6 +216,28 @@ def hdl_wid_116(_: WIDParams):
...
+
+    stack = get_stack()
+    broadcaster_addr_type = 0x00  # Public address
+    broadcaster_addr = '001122334455'
+    advertiser_sid = 0x01
+    broadcast_id = 0x123456
+
+    bis_sync = 0
+    metadata_len = 0
+    subgroups = struct.pack('<IB', bis_sync, metadata_len)
```
Thalley, line 229: Suggest to add a comment stating why these values are set. 

For the specific test case you are working on, any values work as we just need to expose the BIG Sync state to be 0x01 (which oddly enough is the only state you aren't setting in a variable here). 

It looks like 116 is only called for `BASS/SR/CP/BV-21-C`, so I don't expect this code to break any other tests

autopts/wid/bass.py
```diff
@@ -214,6 +216,28 @@ def hdl_wid_116(_: WIDParams):
...
+    btp.bap_add_broadcast_receive_state(
+        broadcaster_addr_type, broadcaster_addr,
+        advertiser_sid, broadcast_id,
+        0x01, 1, subgroups)
+
+    ev = wait_event_with_condition(stack.bap.event_queues[defs.BTP_BAP_EV_BROADCAST_RECEIVE_STATE],
+                                   lambda ev: ev.get('big_encryption') == 0x01, 30, True)
+    if ev is None:
+        log('Broadcast Receive State notification with BIG_Encryption=0x01 not received.')
+        return False
```
Thalley: Suggest to follow the convention to define helper functions when waiting for events like this. We should also check that all the other values match

Thalley (changes requested):

autopts/pybtp/btp/bap.py
omkar3141: Do you have any suggestion for this?

autopts/pybtp/btp/bap.py
omkar3141: Yes, I think linter is complaining.

autopts/pybtp/btp/bap.py
Thalley: As discussed offline, I think `btp_bap_scan_delegator_add_src` is a fine name for this

autopts/wid/bass.py
omkar3141: Added a comment.

autopts/pybtp/btp/bap.py
Thalley: In CI, or locally on your PC?

autopts/pybtp/btp/bap.py
Thalley: Ah, I see that it gets a response with some data that we want to return here. Please add a comment stating that we are here returning the src_id

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -140,6 +140,19 @@ def wait_broadcast_receive_state_ev(self, broadcast_id, peer_addr_type, peer_add
                         ev['pa_sync_state']),
             timeout, remove)
 
+    def wait_broadcast_receive_state_with_fields_ev(self, broadcast_id,
+                                                    broadcaster_addr_type, broadcaster_addr,
+                                                    advertiser_sid, big_encryption,
+                                                    subgroups, timeout, remove=False):
```
Thalley:
```suggestion
                                                    subgroups, timeout=10, remove=False):
```

Suggest to have some default timeout value

autopts/wid/bass.py
```diff
@@ -214,6 +215,33 @@ def hdl_wid_116(_: WIDParams):
...
+
+    # Add a Broadcast Receive State with some arbitrary valid values
+    broadcaster_addr_type = 0x00  # Public address
+    broadcaster_addr = '001122334455'
+    advertiser_sid = 0x01
+    broadcast_id = 0x123456
+    bis_sync = 0
+    metadata_len = 0
+    subgroups = struct.pack('<IB', bis_sync, metadata_len)
+    big_encryption = 0x01
```
Thalley: `# Add a Broadcast Receive State with some arbitrary valid values`

`big_encryption` is not some arbitrary valid value; suggest to make that clear. Additional, I would consider defining the possible BIG Encryption values in an enum and use that to set the value here

autopts/wid/bass.py
```diff
@@ -214,6 +215,33 @@ def hdl_wid_116(_: WIDParams):
...
+
+    btp.bap_scan_delegator_add_src(
+        broadcaster_addr_type, broadcaster_addr,
+        advertiser_sid, broadcast_id,
+        big_encryption, 1, subgroups)
```
Thalley: For simplicity, I think we can actually omit the subgroups here, can't we?

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -140,6 +140,19 @@ def wait_broadcast_receive_state_ev(self, broadcast_id, peer_addr_type, peer_add
                         ev['pa_sync_state']),
             timeout, remove)
 
+    def wait_broadcast_receive_state_with_fields_ev(self, broadcast_id,
```
Thalley: Do we need this AND `wait_broadcast_receive_state_ev`?

autopts/pybtp/btp/bap.py
omkar3141: Locally. I will just revert it.

autopts/wid/bass.py
omkar3141: I will just leave it. It checks that code path anyway.

autopts/ptsprojects/stack/layers/bap.py
omkar3141: I will keep this new function. Modifying the old function to start taking optional parameters will - make it more confusing, and might require changes at many places. I suggest to keep this as refactoring work later.

Thalley: Hmm, it looks like `BTP_BAP_EV_BROADCAST_RECEIVE_STATE` is defined to be the event that we receive as a notification as the BAP Broadcast Assistant. 

We should either apply the necessary changes to allow this BTP event to work for both the client and the server, or we should make a new event that is specific to the server (as the current one is specific for the client). The current solution of half-way using the same event for the reversed role is significantly confusing IMO.

autopts/wid/bass.py
```diff
@@ -214,6 +215,34 @@ def hdl_wid_116(_: WIDParams):
     Please expose a Broadcast Receive State with BIG_Encryption
     field indicating Broadcast_Code required.
     """
+
+    stack = get_stack()
+
+    # Add a Broadcast Receive State with BIG_Encryption field set to indicate
+    # Broadcast_Code required (as per test requirement for BASS/SR/CP/BV-21-C)
```
Thalley:
```suggestion
    # Broadcast_Code required
```

The WID handler perhaps shouldn't be considered unique for this test

autopts/wid/bass.py
Thalley: @omkar3141 looks like you removed the comment again. Please re-add the comment about the values just need to be valid values, and that they do not need to correspond to any actual broadcast source. The only field, as per the WID description, that needs a specific value, is the BIG_Encryption field

autopts/wid/bass.py
```diff
@@ -214,6 +215,34 @@ def hdl_wid_116(_: WIDParams):
...
+    big_encryption = BIGEncryption.BROADCAST_CODE_REQUIRED
+
+    btp.bap_scan_delegator_add_src(
+        broadcaster_addr_type, broadcaster_addr,
+        advertiser_sid, broadcast_id,
+        big_encryption, 1, subgroups)
+
+    ev = stack.bap.wait_broadcast_receive_state_with_fields_ev(
+        broadcast_id, broadcaster_addr_type, broadcaster_addr,
+        advertiser_sid, big_encryption, subgroups, 30, True)
```
Thalley:
```suggestion
    ev = stack.bap.wait_broadcast_receive_state_with_fields_ev(
        broadcast_id, broadcaster_addr_type, broadcaster_addr,
        advertiser_sid, big_encryption, subgroups)
```

omkar3141: Why? I'm using subgroups here, so last one should be True?

Thalley: > Why? I'm using subgroups here, so last one should be True?

Not sure I follow. The last argument is whether to remove the event from the queue - I don't think that has anything to do with subgroups. 

If you do mean to remove it (which arguably should probably be the default for the event (as it is for most)), then I suggest to do

```suggestion
    ev = stack.bap.wait_broadcast_receive_state_with_fields_ev(
        broadcast_id, broadcaster_addr_type, broadcaster_addr,
        advertiser_sid, big_encryption, subgroups, remove=True)
```

autopts/pybtp/btp/bap.py
```diff
@@ -348,6 +350,35 @@ def bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync,
...
 
 
+def bap_scan_delegator_add_src(broadcaster_addr_type, broadcaster_addr,
+                               advertiser_sid, broadcast_id, big_encryption,
+                               num_subgroups, subgroups,
+                               pa_state=BAPPaState.NOT_SYNCED):
+    logging.debug(f"{bap_scan_delegator_add_src.__name__}")
+
+    if not 0x00 <= pa_state <= 0x04:
+        raise BTPError(f'Invalid pa_state {pa_state}')
```
Thalley: Suggest to also add value checks for the other values to ensure correctness. I don't think returning a BTPError is the right approach; I would have gone for a ValueError exception instead

doc/btp_bap.txt
```diff
@@ -362,6 +362,23 @@ Commands and responses:
...
 
+    Opcode 0x1a - Scan Delegator Add Source
+
+                Controller Index:       <controller id>
+                Command parameters:     Broadcaster Address Type (1 octet)
+                                        Broadcaster Address (6 octets)
+                                        Advertiser SID (1 octet)
+                                        Broadcast ID (3 octets)
+                                        BIG Encryption (1 octet)
+                                        PA State (1 octet)
```
Thalley: Seem my comment from https://github.com/zephyrproject-rtos/zephyr/pull/99788#discussion_r2559990522 w.r.t. the order of arguments

autopts/pybtp/btp/bap.py
omkar3141: What do you mean by other values here? Also, Why these are ok ([1](https://github.com/omkar3141/auto-pts/blob/0cdc08461b60fffbff50dd8aa8d8ccc242903312/autopts/pybtp/btp/bap.py#L748-L749), [2](https://github.com/omkar3141/auto-pts/blob/0cdc08461b60fffbff50dd8aa8d8ccc242903312/autopts/pybtp/btp/bap.py#L511-L512)) but here BTPError is not ok? This is framework, so I'm not sure how far we go here.

Or maybe I should just remove value check for "pa_state" as no other params are checked for values anyway.

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -129,15 +129,29 @@ def wait_scan_delegator_found_ev(self, addr_type, addr, timeout, remove=False):
 
     def wait_broadcast_receive_state_ev(self, broadcast_id, peer_addr_type, peer_addr,
                                         broadcaster_addr_type, broadcaster_addr,
-                                        pa_sync_state, timeout, remove=False):
+                                        pa_sync_state, timeout, remove=False,
+                                        *, advertiser_sid=None,
+                                        big_encryption=None, subgroups=None):
```
Thalley: Please keep `timeout` and `remove` as the last parameters to keep up with convention. 

Why do we need the `*` parameter? 

Unless we _need_ to check the `subgroups`, suggest to remove for now.

autopts/pybtp/btp/bap.py
```diff
@@ -348,6 +350,37 @@ def bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync,
     bap_command_rsp_succ()
 
 
+def bap_scan_delegator_add_src(broadcaster_addr_type, broadcaster_addr,
+                               advertiser_sid, broadcast_id, pa_sync_state,
+                               big_encryption, num_subgroups, subgroups):
+    logging.debug(f"{bap_scan_delegator_add_src.__name__}")
+
+    if not 0x00 <= pa_sync_state <= 0x04:
+        raise BTPError(f'Invalid pa_sync_state {pa_sync_state}')
```
Thalley: Since you just added `BAPPaState`, suggest to use that.

autopts/pybtp/types.py
```diff
@@ -505,6 +505,18 @@ class BIGEncryption:
     BAD_CODE = 0x03
 
 
+class BAPPaState:
```
Thalley:
```suggestion
class BASSPASyncState:
```

Althought it may be easier on the eyes to to `class BassPaSyncState`. Not sure what convention we want to follow

autopts/pybtp/btp/bap.py
```diff
@@ -348,6 +350,37 @@ def bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync,
...
+
+    if not 0x00 <= big_encryption <= 0x03:
+        raise BTPError(f'Invalid big_encryption {big_encryption}')
```
Thalley: w.r.t. checking the other values, you are now checking valid values for the `pa_sync_state` and `big_encryption` arguments, but what about the valid values for `broadcaster_addr_type`, `advertiser_sid`, `broadcast_id` and `num_subgroups`?

autopts/wid/bass.py
```diff
@@ -214,6 +215,38 @@ def hdl_wid_116(_: WIDParams):
...
+    big_encryption = BIGEncryption.BROADCAST_CODE_REQUIRED
+
+    btp.bap_scan_delegator_add_src(
+        broadcaster_addr_type, broadcaster_addr,
+        advertiser_sid, broadcast_id,
+        BAPPaState.NOT_SYNCED, big_encryption, 1, subgroups)
+
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, WildCard(), WildCard(),
+        broadcaster_addr_type, broadcaster_addr, WildCard(),
```
Thalley, line 239: Hmm, I see that the other cases for `wait_broadcast_receive_state_ev` uses the `WildCard()` approach. Was not aware that we had such a class. I guess that works well, but in that case consider to either use `=None` in `wait_broadcast_receive_state_ev` or wildcard arguments. Doing both seems sort of weird as they are 2 different solutions to the same problem

autopts/pybtp/btp/bap.py
Thalley: > Why these are ok ([1](https://github.com/omkar3141/auto-pts/blob/0cdc08461b60fffbff50dd8aa8d8ccc242903312/autopts/pybtp/btp/bap.py#L748-L749), [2](https://github.com/omkar3141/auto-pts/blob/0cdc08461b60fffbff50dd8aa8d8ccc242903312/autopts/pybtp/btp/bap.py#L511-L512)) but here BTPError is not ok?

Because they currently don't do anything :upside_down_face: You call `bap_scan_delegator_add_src` as 

```python
    btp.bap_scan_delegator_add_src(
        broadcaster_addr_type, broadcaster_addr,
        advertiser_sid, broadcast_id,
        BAPPaState.NOT_SYNCED, big_encryption, 1, subgroups)
```

But if you ignore the return value, then it doesn't matter that you are checking the values. The Pythonic way of checking input values to a function would be to throw an `ValueError`

autopts/ptsprojects/stack/layers/bap.py
omkar3141: This is going to trigger cascading changes, because we repurposed this event to add pa_sync_state. That is why '*' was also added to prevent breaking other call sites.

autopts/pybtp/btp/bap.py
omkar3141: Decided to skip additional checks for now.

Thalley: That is OK as most other function don't validate anything anyways :upside_down_face:

Thalley:
#AutoPTS run zephyr BASS

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -129,15 +129,29 @@ def wait_scan_delegator_found_ev(self, addr_type, addr, timeout, remove=False):
...
-                                        pa_sync_state, timeout, remove=False):
+                                        pa_sync_state, advertiser_sid=None,
+                                        big_encryption=None, subgroups=None,
+                                        timeout=10, remove=False):
+        def _matches(expected, value):
+            if expected is None:
+                return True
+            if isinstance(expected, WildCard):
+                return True
+            return expected == value
```
Thalley, line 140:
```suggestion
            return expected == value
```

Looking at the implementation of `WildCard`, I think we can skip the extra check here, as comparing with a wildcard object with `==` always returns `True`

codecoup-tester:
Scheduled PR https://github.com/auto-pts/auto-pts/pull/1627#issuecomment-3580368683, board: nrf54, estimated start time: 10:20:03, test case count: 33, estimated duration: 0:26:08<details><summary>Test cases to be run</summary>BASS/SR/SGGIT/SER/BV-01-C<br>BASS/SR/SGGIT/CHA/BV-01-C<br>BASS/SR/SGGIT/CHA/BV-02-C<br>BASS/SR/CP/BV-01-C<br>BASS/SR/CP/BV-02-C<br>BASS/SR/CP/BV-03-C<br>BASS/SR/CP/BV-04-C<br>BASS/SR/CP/BV-05-C<br>BASS/SR/CP/BV-06-C<br>BASS/SR/CP/BV-07-C<br>BASS/SR/CP/BV-08-C<br>BASS/SR/CP/BV-09-C<br>BASS/SR/CP/BV-10-C<br>BASS/SR/CP/BV-11-C<br>BASS/SR/CP/BV-12-C<br>BASS/SR/CP/BV-13-C<br>BASS/SR/CP/BV-14-C<br>BASS/SR/CP/BV-15-C<br>BASS/SR/CP/BV-16-C<br>BASS/SR/CP/BV-17-C<br>BASS/SR/CP/BV-18-C<br>BASS/SR/CP/BV-21-C<br>BASS/SR/CP/BV-22-C<br>BASS/SR/CP/BV-23-C<br>BASS/SR/SPE/BI-01-C<br>BASS/SR/SPE/BI-02-C<br>BASS/SR/SPE/BI-03-C<br>BASS/SR/SPE/BI-04-C<br>BASS/SR/SPE/BI-05-C<br>BASS/SR/SPE/BI-06-C<br>BASS/SR/SPE/BI-07-C<br>BASS/SR/SPE/BI-08-C<br>BASS/SR/SPE/BI-09-C</details>

autopts/ptsprojects/stack/layers/bap.py
Thalley: OK to keep as is

Thalley:
#AutoPTS run zephyr BAP/BA

codecoup-tester:
AutoPTS Bot failed:
Traceback (most recent call last):
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/bot/common.py", line 513, in start
    stats = self.run_tests()
            ^^^^^^^^^^^^^^^^
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/bot/common.py", line 593, in run_tests
    return super().start()
           ^^^^^^^^^^^^^^^
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/client.py", line 1419, in start
    return self.main(args)
           ^^^^^^^^^^^^^^^
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/client.py", line 1465, in main
    self.setup_project_pixits(self.ptses)
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/client.py", line 1484, in setup_project_pixits
    setup_project_pixits(ptses)
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/client.py", line 1663, in setup_project_pixits
    mod.set_pixits(ptses)
  File "/home/codecoup/workspace/autopts/auto-pts/autopts/ptsprojects/zephyr/gatt.py", line 112, in set_pixits
    pts.set_pixit("GATT", "TSPX_selected_handle", "0012")
  File "/usr/lib/python3.12/xmlrpc/client.py", line 1122, in __call__
    return self.__send(self.__name, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/xmlrpc/client.py", line 1461, in __request
    response = self.__transport.request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/xmlrpc/client.py", line 1166, in request
    return self.single_request(host, handler, request_body, verbose)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/xmlrpc/client.py", line 1182, in single_request
    return self.parse_response(resp)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/xmlrpc/client.py", line 1351, in parse_response
    return u.close()
           ^^^^^^^^^
  File "/usr/lib/python3.12/xmlrpc/client.py", line 668, in close
    raise Fault(**self._stack[0])
xmlrpc.client.Fault: <Fault 1: "<class 'autopts.ptscontrol.BadPixitException'>:GATT TSPX_selected_handle 0012:\nPTSCONTROL_E_PIXIT_PARAM_NOT_FOUND">

PR #1628, opened 2025-11-21
zephyr: sm: add wid 175, update wid 108

m-alperen-sener:
WID 175 is added for test case SM/CEN/JW/BI-06-C
Please confirm that IUT reports the Confirm Value Failed to the Upper Tester

SM/CEN/JW/BV-01-C needs bondable flag to be disabled.


autopts/wid/sm.py
```diff
@@ -245,6 +253,25 @@ def hdl_wid_174(_: WIDParams):
...
+
+    stack = get_stack()
+    data = stack.gap.gap_wait_for_pairing_fail(10)
+
+    if data is None:
+        return False
+    else:
+        logging.debug("Pairing failed with reason: %r", data[-1])
+        # BT_SMP_ERR_CONFIRM_FAILED
+        if data[-1] == 0x01:
```
m-alperen-sener: https://github.com/nrfconnect/sdk-zephyr/blob/main/subsys/bluetooth/host/smp.c#L481

BT_SMP_ERR_CONFIRM_FAILED is converted to BT_SECURITY_ERR_AUTH_FAIL.

HaavardRei: Maybe we should leave a comment here mentioning that we expect BT_SMP_ERR_CONFIRM_FAILED (0x04), but before triggering the pairing_failed callback this is translated to BT_SECURITY_ERR_AUTH_FAIL (0x01)?

autopts/wid/sm.py
```diff
@@ -245,6 +253,22 @@ def hdl_wid_174(_: WIDParams):
...
+    Please confirm that IUT reports the Confirm Value Failed to the Upper Tester
+    """
+
+    stack = get_stack()
+    _, _, reason = stack.gap.gap_wait_for_pairing_fail(10)
+
+    # BT_SMP_ERR_CONFIRM_FAILED(0x04) is converted to BT_SECURITY_ERR_AUTH_FAIL(0x01)
+    # in zephyr host.
+    if reason == 0x01:
+        return True
```
szymon-czapracki: If this WID has specific behavior for Zephyr host, maybe it would be better to implement it in zephyr/sm_wid.py ?

sjanc: I'd suggest to define this on BTP level  (I'm fine if you want to map this one-to-one to zephyr enum)

m-alperen-sener: > If this WID has specific behavior for Zephyr host, maybe it would be better to implement it in zephyr/sm_wid.py ?

Fair point, seems like it is Zephyr specific, so better to have it under zephyr.

> I'd suggest to define this on BTP level (I'm fine if you want to map this one-to-one to zephyr enum)

I think moving it to zephyr/sm_wid.py is better option for now but we should keep it also here with correct error Code checking.

IMO The reasons should be passed to AutoPTS to match/compare with the WID description. [Table 3.7: Pairing Failed reason codes](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-62/out/en/host/security-manager-specification.html#UUID-4fde8eb1-530c-dd12-bbd2-78a326f12f31_table-idm13359014795812).

This way it will guide future work in the correct direction, because from what I see any host should be able to deliver the Pairing Failed reason codes to User, in our case Upper Tester.

sjanc: There is no need to IUT specific WID handler here.

We should clarify that 'reason' in ` Opcode 0x8c - Pairing Failed event` is expected to be SMP reason.
And then it is up to IUT (ie tester application) to map any internal behavior into that code.

autopts/ptsprojects/zephyr/sm_wid.py
```diff
@@ -36,3 +37,19 @@ def hdl_wid_143(desc):
...
+    """
+    Please confirm that IUT reports the Confirm Value Failed to the Upper Tester
+    """
+
+    stack = get_stack()
+    _, _, reason = stack.gap.gap_wait_for_pairing_fail(10)
+
+    # BT_SMP_ERR_CONFIRM_FAILED(0x04) is converted to BT_SECURITY_ERR_AUTH_FAIL(0x01)
+    # in zephyr host.
+    if reason == 0x01:
```
szymon-czapracki, line 53: Might be a good idea to define this value instead of "magic" 0x01 - after that the comment might be unnecessary, as it becomes self-explanatory; then again, having a well commented code is always good - so leaving it is good too :).

m-alperen-sener: My intention is not to invest on that much I intend to update zephyr and remove this completely, soon.

autopts/pybtp/defs.py
```diff
@@ -988,3 +988,23 @@ def BIT(bit):
...
+BTP_SMP_PAIR_FAIL_REASON_COMMAND_NOT_SUPPORTED = 0x07
+BTP_SMP_PAIR_FAIL_REASON_UNSPECIFIED = 0x08
+BTP_SMP_PAIR_FAIL_REASON_REPEATED_ATTEMPTS = 0x09
+BTP_SMP_PAIR_FAIL_REASON_INVALID_PARAMS = 0x0A
+BTP_SMP_PAIR_FAIL_REASON_DHKEY_CHECK_FAILED = 0x0B
+BTP_SMP_PAIR_FAIL_REASON_NUMERIC_COMPARISON_FAILED = 0x0C
+BTP_SMP_PAIR_FAIL_REASON_BR_EDR_PAIRING_IN_PROGRESS = 0x0D
+BTP_SMP_PAIR_FAIL_REASON_CROSS_TRANSPORT_KEY_NOT_SUPPORTED = 0x0E
+BTP_SMP_PAIR_FAIL_REASON_KEY_REJECTED = 0x0F
+BTP_SMP_PAIR_FAIL_REASON_BUSY = 0x10
```
Thalley, line 1010: Are these specific to BTP, or are they the SMP reason from the core spec? If the former, this is fine, if the latter, I think we should define them somewhere else, or at least without the BTP prefix, preferably as an `enum`

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (1)</summary>SM      SM/CEN/JW/BV-01-C               PASS</details>

autopts/pybtp/defs.py
m-alperen-sener: > Are these specific to BTP, or are they the SMP reason from the core spec? If the former, this is fine, if the latter, 

from spec ```# Vol 3, Part H, Section 3.5.5``` I added as comment.

>I think we should define them somewhere else, or at least without the BTP prefix, preferably as an `enum`

I have no strong idea about it, since but maybe it is better in `pybtp/btp/sm.py`. If this is OK I can move the list there without BTP prefix and as enum.

m-alperen-sener: And here the link to spec https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-62/out/en/host/security-manager-specification.html#UUID-4fde8eb1-530c-dd12-bbd2-78a326f12f31_table-idm13359014795812

autopts/ptsprojects/zephyr/sm_wid.py
Thalley: If this is a Zephyr workaround, then shouldn't this go in into https://github.com/auto-pts/auto-pts/blob/master/autopts/ptsprojects/zephyr/sm_wid.py instead of the generic file? 

Please also create a Zephyr GH issue or PR that we can refer to here if you intent on this being a temporary fix that depends on a change in Zephyr.

m-alperen-sener: > If this is a Zephyr workaround, then shouldn't this go in into https://github.com/auto-pts/auto-pts/blob/master/autopts/ptsprojects/zephyr/sm_wid.py instead of the generic file?
> 

Correct me if I am wrong but I though I already did that, "autopts/ptsprojects/zephyr/sm_wid.py" is used for zephyr workaround.

> Please also create a Zephyr GH issue or PR that we can refer to here if you intent on this being a temporary fix that depends on a change in Zephyr.

I created an issue for it: https://github.com/zephyrproject-rtos/zephyr/issues/100290

Thalley: Somehow I misread the path - My bad :D Thanks for creating the issue

PR #1639, opened 2025-11-27
Change csv report profile view

szymon-czapracki:
Grepping through profile was sometimes tiresome.
Now before every row there is information about profile.


szymon-czapracki:
@sjanc FYI

sjanc (approved):

PR #1641, opened 2025-11-28
Reset Broadcast ID IXIT for other test cases

Frodevan:
Since BAP/BSRC/SCC/BV-38-C is the only test-case to use the Broadcast ID IXIT values, the other test-cases should reset this back to null. Otherwise, certain other test-cases can fail by chance if run after BAP/BSRC/SCC/BV-38-C in a run.

### DIFF
--- a/autopts/ptsprojects/zephyr/bap.py
+++ b/autopts/ptsprojects/zephyr/bap.py
@@ -55,6 +55,8 @@ def set_pixits(ptses):
     pts.set_pixit("BAP", "TSPX_VS_Codec_ID", "ffff")
     pts.set_pixit("BAP", "TSPX_VS_Company_ID", "ffff")
     pts.set_pixit("BAP", "TSPX_broadcast_code", BROADCAST_CODE)
+    pts.set_pixit("BAP", "TSPX_Broadcast_ID", "0")
+    pts.set_pixit("BAP", "TSPX_Broadcast_ID_2", "0")
 
     if len(ptses) < 2:
         return


Thalley:
> Since BAP/BSRC/SCC/BV-38-C is the only test-case to use the Broadcast ID IXIT values, the other test-cases should reset this back to null. Otherwise, certain other test-cases can fail by chance if run after BAP/BSRC/SCC/BV-38-C in a run.

Can you elaborate on this? 

If `BAP/BSRC/SCC/BV-38-C` is the only test-case that uses this value, it should not affect any other tests, so your description seems to counter itself. 

If other tests are using it, then it sounds like we need to investigate _why_ and ensure that we set the value correctly.

How did you determine that it was "by chance"?

Thalley:
> > Can you elaborate on this?
> 
> We see in the logs of our test runs that for BAP tests that are run after BAP/BSRC/SCC/BV-38-C, the Broadcast ID IXIT is set to the value that was previously set by BAP/BSRC/SCC/BV-38-C. This causes the other test cases which should not use the v2 command to use it, and this is a problem for test-cases such as BAP/BSRC/SCC/BV-06-C and BAP/BSRC/STR/BV-16-C.

But that goes against your statement 

> BAP/BSRC/SCC/BV-38-C is the only test-case to use the Broadcast ID IXIT values

If this is the only test, then others aren't affected. If others are affected, when this may actually just hide the real problem with those tests, rather than fixing them. Broadcast_ID = 0x000000 is a valid value, so something is suspicious here. The default value seems to be 0, but PTS also has

> <Description>Broadcast ID. When this value is used, PTS will search for the matching device. If it is non-zero value, it will search by Bluetooth address</Description>

PTS should not use a valid broadcast_id as a special value (indicating search by address). 

@sjanc Are we not resetting the IXIT values between each test run? 

I'll approve this PR, as the changes are sound (just setting some default value), but to me it seems like there are 2 "bugs":
1) IXIT values are not reset between each test run (this can harmful if tests are relying on PTS defaults that we aren't actively setting, if it reuses the value(s) from previous tests)
2) PTS is using a valid value, 0, as a special value

PR #1649, opened 2025-12-03
wid: Disable MITM in SM WID 108 handler

sjanc:
SM/CEN/JW/BV-01-C test requires IUT to send Pairing Request with any IO capa but MITM and Bonding flags set to 0.

### DIFF
--- a/autopts/wid/sm.py
+++ b/autopts/wid/sm.py
@@ -105,9 +105,10 @@ def hdl_wid_108(params: WIDParams):
     Please start pairing process.
     """
 
-    # SM/CEN/JW/BV-01-C AuthReq bonding flag needs to be set to 0x00
+    # SM/CEN/JW/BV-01-C AuthReq Bonding Flags set to '00' and the MITM flag set to '0'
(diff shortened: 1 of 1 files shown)


autopts/wid/sm.py
```diff
@@ -105,9 +105,10 @@ def hdl_wid_108(params: WIDParams):
     Please start pairing process.
     """
 
-    # SM/CEN/JW/BV-01-C AuthReq bonding flag needs to be set to 0x00
+    # SM/CEN/JW/BV-01-C  AuthReq Bonding Flags set to ‘00’ and the MITM flag set to ‘0’ a
```
Thalley:
```suggestion
    # SM/CEN/JW/BV-01-C  AuthReq Bonding Flags set to '00' and the MITM flag set to '0' a
```

Suggest to use the ASCII quotes instead :)

Thalley (approved):

autopts/wid/sm.py
sjanc: ah yes :)  c'n'p from spec :)

PR #1653, opened 2025-12-09
client: Add BTP fatal error status

mkasenberg:
Add a new test case status representing errors that occur at critical points (such as enabling Bluetooth on the IUT). When this status is reported, the IUT requires recovery.


autopts/pybtp/btp/btp.py
```diff
@@ -438,7 +438,10 @@ def set_lt3_addr(addr, addr_type):
 
 
 def core_reg_svc_gap():
-    core_reg_svc_univ("gap_reg", "GAP")
+    try:
+        core_reg_svc_univ("gap_reg", "GAP")
+    except BTPError as e:
+        raise BTPFatalError("Failed to enable Bluetooth in the IUT!") from e
```
sjanc: Lets be explicit "Failed to register BTP GAP service in the IUT"

szymon-czapracki (approved):

PR #1660, opened 2025-12-19
Conn dconn test fix

m-alperen-sener:
Addressing PTS 8.11.1b6 updates for test cases:
GAP/CONN/DCON/BV-04-C
GAP/CONN/DCON/BV-05-C

PTS fixes the GATT CAR attributes request and response issue. Includes disconnect WID into the test run.

Required Tester Changes:
https://github.com/zephyrproject-rtos/zephyr/pull/101349

zephyr: Updating WID 118 and WID 406
    
    WID 118: For test cases GAP/CONN/DCON/BV-[05-C, 04-C] we need to
    wait for CAR received event before allowing disconnect.
    
    WID 406: Returning True when privacy is not supported
    might cause false positive test result. If privacy is not supported
    test should fail becuase those WIDs are specifically dictates address
    resolution, which means privacy should be supported.

zephyr: Adding BTP_GAP_EV_PEER_CAR_RECEIVED event to GAP stack
    
    Test cases GAP/CONN/DCON/BV-05-C and GAP/CONN/DCON/BV-04-C needs
    to read Central Address Resolution characteristics. In order to
    prevent disconnect we need to know if CAR characteristic is read.

### DIFF
--- a/autopts/wid/gap.py
+++ b/autopts/wid/gap.py
@@ -1826,18 +1826,13 @@ def hdl_wid_405(_: WIDParams):
 
 
 def hdl_wid_406(_: WIDParams):
-    # Please enter General Connectable Mode using private addresses.
-    # Note: IUT is expected to do directed advertising with target address set
-    # to peer RPA...when WID 403 is received
+    # Please enter Direct Connectable Mode targeting the PTS by its identity
+    # address.
+    bd_addr = btp.pts_addr_get()
+    bd_addr_type = btp.pts_addr_type_get()
 
-    stack = get_stack()
-
-    if stack.gap.iut_has_privacy():
-        addr_type = OwnAddrType.le_resolvable_private_address
-    else:
-        addr_type = OwnAddrType.le_identity_address
-
-    btp.gap_adv_ind_on(ad=stack.gap.ad, own_addr_type=addr_type)
+    btp.gap_set_conn()
+    btp.gap_direct_adv_on(bd_addr, bd_addr_type, high_duty=0, peer_rpa=1)
 
     return True
 
--- a/autopts/ptsprojects/zephyr/gap_wid.py
+++ b/autopts/ptsprojects/zephyr/gap_wid.py
@@ -15,6 +15,7 @@
 
 import logging
 
+from autopts.pybtp import btp
 from autopts.pybtp.types import WIDParams
 from autopts.wid import generic_wid_hdl
 
@@ -31,6 +32,21 @@ def hdl_wid_104(_: WIDParams):
 
 
 def hdl_wid_114(_: WIDParams):
+    return True
+
+
+def hdl_wid_118(params: WIDParams):
+    """
+    Please press ok to disconnect the link.
+    """
+
+    # Directed connection test cases with privacy needs to
+    # make sure that Central Address Resolution attribute is
+    # received.
+    if params.test_case_name in ['GAP/CONN/DCON/BV-04-C',
+                                 'GAP/CONN/DCON/BV-05-C']:
+        return btp.gap_wait_for_car_receive()
+
     return True
 
 
--- a/autopts/pybtp/btp/gap.py
+++ b/autopts/pybtp/btp/gap.py
@@ -688,7 +688,8 @@ def gap_adv_off():
 
 
 def gap_direct_adv_on(addr, addr_type, high_duty=0, peer_rpa=0):
-    logging.debug("%s %r %r", gap_direct_adv_on.__name__, addr, high_duty)
+    logging.debug("%s %r %r %r %r", gap_direct_adv_on.__name__, addr,
+                  addr_type, high_duty, peer_rpa)
 
     stack = get_stack()
 


autopts/wid/gap.py
```diff
@@ -1802,13 +1806,9 @@ def hdl_wid_402(_: WIDParams):
 def hdl_wid_403(_: WIDParams):
     # Please perform the Directed Connection Establishment Procedure using RPA
     # then resolve the PTS address and connect with PTS.
-    stack = get_stack()
-
-    if not stack.gap.iut_has_privacy():
```
sjanc, line 1807: why is this check removed?

autopts/wid/gap.py
```diff
@@ -1802,13 +1806,9 @@ def hdl_wid_402(_: WIDParams):
...
-        return False
-
-    btp.gap_conn()
+    bd_addr = btp.pts_addr_get()
 
+    btp.gap_conn(bd_addr=bd_addr, bd_addr_type=OwnAddrType.le_identity_address)
```
sjanc: shouldn't this be le_resolvable_private_address ?

Initial conditions for this tests are
- "The IUT is in Privacy mode."
- "The IUT is using a resolvable private address and has distributed its IRK."

autopts/wid/gap.py
```diff
@@ -1826,18 +1826,12 @@ def hdl_wid_405(_: WIDParams):
...
-        addr_type = OwnAddrType.le_resolvable_private_address
-    else:
-        addr_type = OwnAddrType.le_identity_address
-
-    btp.gap_adv_ind_on(ad=stack.gap.ad, own_addr_type=addr_type)
+    # Please enter Direct Connectable Mode targeting the PTS by its identity
+    # address.
+    bd_addr = btp.pts_addr_get()
+    btp.gap_set_conn()
+    btp.gap_direct_adv_on(bd_addr, OwnAddrType.le_identity_address,
```
sjanc: same here

autopts/wid/gap.py
m-alperen-sener: Again PTS (ab)use of WID. This WID used to be used for test case GAP/CONN/DCON/BV-05-C, along with WID 406 which was the one should have been actually used. Now they have removed it from GAP/CONN/DCON/BV-05-C test case process and now it can do what  it supposed to do.

Anywat WID was a ghost before; returning true or false depending on privacy enabled or not, but now it is set to start advs with RPAs.

autopts/wid/gap.py
m-alperen-sener: I need to check this, I must have done something wrong because I remember seeing Random Private Address on sinnefer. Also PTS was positive about it, they might also have bug :(

I will get back on this ASAP

autopts/wid/gap.py
m-alperen-sener: yeap I ll check that.

m-alperen-sener:
> BTW I was thinning on this and my suggestion would be to issue PTS support ticket on how CAR read should be handled in that test (my opinion is that if IUT claims support for CAR read than PTS should wait until read is done and only then terminate link), until this is solved on PTS side we can have zephyr specific WID handler (eg with simple sleep for time being). And additional BTP event should be added only if PTS would require CAR confirmation from IUT

I think the PTS implementation is fine; consider this; PTS should support both IUT implementation, IUT does read CAR and IUT does not read CAR. From PTS point of view, well, it is fair to expect IUT to end connection, otherwise how would they new, introduce some timeout etc. unnecessarily complex. Again it fair enough, then we should do our part.

We can go the other way and stop reading CAR status and embed the behavior into AutoPTS, because we know which test case is what and return fake trues, well this functionality is not in zephyr host anyway and testing the tester has no benefit, like we used to do. But I am not OK with this.

Also please check:  https://github.com/zephyrproject-rtos/zephyr/issues/101955

autopts/wid/gap.py
m-alperen-sener: Changed this to use  btp.pts_addr_type_get() to make sure we are using what is reported by IUT. Also this is not used by IUT when deciding on directed adv should be with or without RPA, it is decided by peer_rpa parameter in gap_direct_adv_on API if the advs should be directed with RPA.

autopts/wid/gap.py
```diff
@@ -698,10 +698,18 @@ def hdl_wid_114(params: WIDParams):
...
     """
         Please press ok to disconnect the link.
     """
+
+    # Directed connection test cases with privacy needs to
+    # make sure that Central Address Resolution attribute is
+    # received.
+    if params.test_case_name in ['GAP/CONN/DCON/BV-04-C',
+                                 'GAP/CONN/DCON/BV-05-C']:
+        return btp.gap_wait_for_car_receive()
```
sjanc: this should be optional eg.

```
if stack.gap.car_read:
   return btp.gap_wait_for_car_receive()
```

m-alperen-sener: OK I see your point now, again my mistake is not considering other stacks, sorry.

is it OK if I move this into zephyr? because this is needed for zephyr.

sjanc: yeap, this is fine too

sjanc (approved):

autopts/wid/gap.py
m-alperen-sener: done

PR #1661, opened 2026-01-08
Wid ead implementation

kedMertens:
I've noticed that [101333](https://github.com/zephyrproject-rtos/zephyr/pull/101333) didn't affected latest test runs. Turns out implementation was not pushed to upstream.
Here is the implementation for encrypted advertising (EAD) test cases, GAP/SCN/BV-01-C and GAP/ADV/BV-20-C.


autopts/pybtp/btp/gap.py
```diff
@@ -1479,6 +1482,57 @@ def check_scan_rep_and_rsp(report, response):
     return False
 
 
+def _setup_ead_key_material():
```
Thalley, line 1490: Why is this specific function prefixed with `_`?

autopts/pybtp/btp/gap.py
```diff
@@ -1479,6 +1482,57 @@ def check_scan_rep_and_rsp(report, response):
...
+    """
+    Set up EAD key material (key and IV) for encryption/decryption.
+    Following must match the values in IXIT.
+    PTS considers IV to be a number big-endian format hence the [::-1]
```
Thalley: > PTS considers IV to be a number big-endian format hence the [::-1]

Consider moving this line from the function description to the line that does this

autopts/pybtp/btp/gap.py
```diff
@@ -1479,6 +1482,57 @@ def check_scan_rep_and_rsp(report, response):
     return False
 
 
+def _setup_ead_key_material():
+    """
+    Set up EAD key material (key and IV) for encryption/decryption.
+    Following must match the values in IXIT.
```
Thalley: > Following must match the values in IXIT.

There isn't anything following this line :)

autopts/pybtp/btp/gap.py
```diff
@@ -1479,6 +1482,57 @@ def check_scan_rep_and_rsp(report, response):
...
+    logging.debug("check_decrypted_ead")
+    if not plain_text:
+        logging.debug('plain_text is None, no hex sequence')
+        return False
+    stack = get_stack()
+    devices = stack.gap.found_devices.data
+
+    for device in devices:
+        # reject if EAD data is not present
+        if len(device.eir) < 10 or device.eir[1] != 0x31:
```
Thalley: What are the magic numbers `10` and `0x31` here? 

If they are defined values, please use the existing definitions (if any) or add new ones

autopts/pybtp/btp/gap.py
```diff
@@ -1479,6 +1482,57 @@ def check_scan_rep_and_rsp(report, response):
...
+    """
+    iutctl = get_iut()
+    key = bytes.fromhex("00112233445566778899AABBCCDDEEFF")
+    iv = bytes.fromhex("0001020304050607")[::-1]
+    material_ba = bytearray(key) + iv
+    iutctl.btp_socket.send(*GAP['set_ead_key_material'], data=material_ba)
+    gap_command_rsp_succ(defs.BTP_GAP_CMD_SET_EAD_KEY_MATERIAL)
+
+
+def check_decrypted_ead(plain_text):
```
Thalley: Please add `docstring` for this function describing the parameters and return values

autopts/pybtp/btp/gap.py
```diff
@@ -1479,6 +1482,57 @@ def check_scan_rep_and_rsp(report, response):
...
+        try:
+            tuple_data = gap_command_rsp_succ(defs.BTP_GAP_CMD_DECRYPT_EAD_DATA)
+        except BTPError as e:
+            logging.error(f"Failed to decrypt EAD data: {e}")
+            return False
+        return tuple_data[0][1:] == b''.fromhex(plain_text)
+    return False
+
+
+def create_ead_adv(payload):
```
Thalley, line 1555: Please add docstring for this function describing the parameters and return values

autopts/wid/gap.py
```diff
@@ -1894,6 +1894,38 @@ def hdl_wid_451(_: WIDParams):
...
 
+def hdl_wid_500(_: WIDParams):
+    return hdl_wid_501(_)
+
+
+def hdl_wid_501(_: WIDParams):
+    # Have IUT send Encrypted Advertising Data (EAD) packets
+    # PTS asks payload to be 'payload 1' for WID 500 or 'payload 2' for WID 501
+
+    WID_TO_PAYLOAD = {500: 1, 501: 2}
```
Thalley: I would actually probably prefer that `hdl_wid_500` does not call `hdl_wid_501` in this case when they differ in actions/values. 

If you want to share the core, suggest to create another function that both `hdl_wid_500` and `hdl_wid_501` calls with the specific payload

autopts/wid/gap.py
```diff
@@ -1894,6 +1894,38 @@ def hdl_wid_451(_: WIDParams):
...
+    adv_data = {}
+    encrypted = btp.create_ead_adv(payload)
+    btp.gap_adv_off()
+    adv_data[AdType.encrypted_data] = struct.pack(f'<{len(encrypted)}B', *encrypted)
+    btp.gap_adv_ind_on(ad=adv_data)
+    return True
+
+
+def hdl_wid_502(params: WIDParams):
+    # Find all hex sequences of reasonable length and take the longest
```
Thalley: Please add the WID description as we get it from PTS so it's easier to know what this handler does

autopts/wid/gap.py
```diff
@@ -1894,6 +1894,38 @@ def hdl_wid_451(_: WIDParams):
...
+
+    hex_sequences = re.findall(r"\b([0-9a-fA-F]{8,})\b", params.description)
+    plain_text = max(hex_sequences, key=len) if hex_sequences else None
```
Thalley: If `hex_sequences` is None, doesn't that indicate an issue/bug?

autopts/wid/gap.py
```diff
@@ -1894,6 +1894,38 @@ def hdl_wid_451(_: WIDParams):
...
+    log(f'plain text : {plain_text}')
+    # scan for a while to get encrypted adv packet
+    btp.gap_start_discov(discov_type='passive', mode='observe')
+    sleep(5)
+    btp.gap_stop_discov()
```
Thalley, line 1941: Ideally we'd run discover until we get the expected event instead of using a timer, but I also know that this is common practice in gap.py, so it's OK

autopts/pybtp/btp/gap.py
kedMertens: To show that it is only used within this file "internally", so that it is obvious that it is just a helper function

sjanc:
#AutoPTS run zephyr GAP/SCN/BV-01-C GAP/ADV/BV-20-C

codecoup-tester:
Scheduled PR https://github.com/auto-pts/auto-pts/pull/1661#issuecomment-3737710506, board: nrf54, estimated start time: 10:55:22, test case count: 1, estimated duration: 0:10:26<details><summary>Test cases to be run</summary>GAP/SCN/BV-01-C</details>

autopts/pybtp/btp/gap.py
```diff
@@ -40,6 +40,14 @@
 from autopts.pybtp.btp.btp import get_iut_method as get_iut
 from autopts.pybtp.types import Addr, AdDuration, AdType, BTPError, OwnAddrType, addr2btp_ba, gap_settings_btp2txt
 
+# Import EAD constants from PIXIT configuration
+try:
+    from autopts.ptsprojects.zephyr.gap import EAD_ENCRYPTION_KEY, EAD_INITIALIZATION_VECTOR
+except ImportError:
+    # Fallback for other projects that may not have these constants
+    EAD_ENCRYPTION_KEY = "00112233445566778899AABBCCDDEEFF"
+    EAD_INITIALIZATION_VECTOR = "0001020304050607"
```
Thalley: Why don't we define them here, instead of the Zephyr-specific file?

autopts/pybtp/btp/gap.py
```diff
@@ -1479,6 +1490,78 @@ def check_scan_rep_and_rsp(report, response):
...
 
+def _setup_ead_key_material():
+    """
+    Set up EAD key material (key and IV) for encryption/decryption.
+    The values must match the IXIT configuration used by PTS.
+    """
+    iutctl = get_iut()
+    key = bytes.fromhex(EAD_ENCRYPTION_KEY)
+    # PTS considers IV to be a number big-endian format hence the [::-1]
+    iv = bytes.fromhex(EAD_INITIALIZATION_VECTOR)[::-1]
```
Thalley, line 1498: Outside the scope of this PR, but perhaps we should consider a helper function to convert between big and little endian 

Image

autopts/wid/gap.py
```diff
@@ -1894,6 +1894,48 @@ def hdl_wid_451(_: WIDParams):
     return True
 
 
+def _send_ead_adv(payload_number):
+    """Helper function to send Encrypted Advertising Data (EAD) packets."""
+    payload = f'payload {payload_number}'.encode()
+    adv_data = {}
+    encrypted = btp.create_ead_adv(payload)
+    btp.gap_adv_off()
```
Thalley: I think you can use `stack.gap.current_settings_get` to get the current state and determine if we need to stop advertising first, rather than always attempting to do so before restarting it

Additionally, unless sending a large amount of data, it shouldn't be necessary to restart advertising to set new data

autopts/wid/gap.py
```diff
@@ -1894,6 +1894,48 @@ def hdl_wid_451(_: WIDParams):
     return True
 
 
+def _send_ead_adv(payload_number):
+    """Helper function to send Encrypted Advertising Data (EAD) packets."""
```
Thalley: Please add documentation for the parameter

Thalley (commented):
Generally looks good. A few comments

autopts/pybtp/btp/gap.py
sjanc: yeap, I think in general we should have convention of  btp2foo and foo2btp API for such things (addr, uuid, IV etc)

autopts/pybtp/btp/gap.py
kedMertens: I was trying to get rid of duplication and to use same values as for PIXIT in zephyr file, to make it error prone.

Thalley: Not sure I understand. Wouldn't it make more sense to define them here, and have the Zephyr gap.py file use the general definitions, instead of the other way around?

kedMertens: ah, got it. Yeah it makes more sense, thanks

Thalley:
> however currently zephyr host is marked as Core 6.0
> 
> @Thalley should we bump Core version to 6.2 ?

Marked as 6.0 in the ICS, or somewhere else in Zephyr? Either way, yeah let's bump it to 6.2

autopts/pybtp/btp/gap.py
```diff
@@ -1479,6 +1487,78 @@ def check_scan_rep_and_rsp(report, response):
...
+    material_ba = bytearray(key) + iv
+    iutctl.btp_socket.send(*GAP['set_ead_key_material'], data=material_ba)
+    gap_command_rsp_succ(defs.BTP_GAP_CMD_SET_EAD_KEY_MATERIAL)
+
+
+def check_decrypted_ead(plain_text):
+    """Check if decrypted EAD data matches the expected payload.
+
```
Thalley, line 1506: I think there's a mismatch between the function name + description, and the implementation. 

The function seems to do a lot more than "check", i.e. it sends a `BTP_GAP_CMD_SET_EAD_KEY_MATERIAL` and also one or more `BTP_GAP_CMD_DECRYPT_EAD_DATA` commands. 

A function called `check` should, IMO, be stateless, which isn't the case here. 

Consider renaming the function and updating the description to better describe what it does, or implement the functionality of this function into several functions and keep the `check` part as a stateless function

autopts/wid/gap.py
```diff
@@ -1894,6 +1894,55 @@ def hdl_wid_451(_: WIDParams):
     return True
 
 
+def _send_ead_adv(payload_number):
+    """Helper function to send Encrypted Advertising Data (EAD) packets.
+
+    Arguments:
+    payload_number -- integer indicating which payload to send (1 or 2).
```
Thalley: And payload here is just a value like 0x01 and 0x02 that is being encrypted and sent? Or do they refer to some payload indexed or otherwise identified with the values `1` or `2`? 

If the former, then the function should just take any input to be encrypted. 
If the latter, it needs a reference to those payloads

PR #1676, opened 2026-01-20
wid: Fix BAP WID 349 handler

sjanc:
IUT is expected to send sync transfer to lower tester. This was affecting BAP/BA/BASS/BV-08-C.

### DIFF
--- a/autopts/wid/bap.py
+++ b/autopts/wid/bap.py
@@ -2170,7 +2170,7 @@ def hdl_wid_349(params: WIDParams):
     broadcast_id = ev['broadcast_id']
     padv_interval = ev['padv_interval']
     num_subgroups = 1
-    padv_sync = 0x01
+    padv_sync = PaSyncState.SYNC_INFO_REQ
     bis_sync = 0
     metadata_len = 0
     subgroups = struct.pack('<IB', bis_sync, metadata_len)
@@ -2181,10 +2181,20 @@ def hdl_wid_349(params: WIDParams):
 
     ev = stack.bap.wait_broadcast_receive_state_ev(
         broadcast_id, addr_type, addr, broadcaster_addr_type,
-        broadcaster_addr, padv_sync, timeout=10, remove=False)
+        broadcaster_addr, padv_sync, timeout=10, remove=True)
 
     if ev is None:
         return False
+
+    if ev['pa_sync_state'] == PaSyncState.SYNC_INFO_REQ:
+        btp.bap_send_past(ev['src_id'])
+
+        ev = stack.bap.wait_broadcast_receive_state_ev(
+            broadcast_id, addr_type, addr, broadcaster_addr_type,
+            broadcaster_addr, PaSyncState.SYNCED, timeout=10, remove=True)
+
+        if ev is None:
+            return False
 
     return True
 


sjanc:
#AutoPTS run zephyr BAP/BA/BASS/BV-08-C.

codecoup-tester:
Scheduled PR https://github.com/auto-pts/auto-pts/pull/1676#issuecomment-3773683667, board: native+nrf54, estimated start time: 17:03:18, test case count: estimation not available

autopts/wid/bap.py
```diff
@@ -2186,6 +2186,16 @@ def hdl_wid_349(params: WIDParams):
     if ev is None:
         return False
 
+    if ev['pa_sync_state'] == PaSyncState.SYNC_INFO_REQ:
+        btp.bap_send_past(ev['src_id'])
+
+        ev = stack.bap.wait_broadcast_receive_state_ev(
+            broadcast_id, addr_type, addr, broadcaster_addr_type,
+            broadcaster_addr, WildCard(), timeout=10, remove=True)
```
Thalley: In the lines `R2182-R2184` we also wait for this event, but don't remove it there. 

1) Should we remove the previous event (i.e. set `remove=True`
2) I think we should also expect a specific `pa_sync_state` state here instead of `WildCard()`; we want to wait for a receive state with a synced PA, right?

sjanc: this is good point!  TBH I don't remember why there is even 'remove' parameter in those event handlers in first place...

however in this test we don't need to wait for this at all since test completes on PAST, but I'd keep it here to make it similar to other wids

sjanc: ~hmm_ I did some testing and doing only PAST makes test conclude \~10s sooner (\~55s vs \~45s) :)~

Thalley: I'm not saying that it won't pass, but it looks like it may have a false positive if we use `WildCard()` instead of the actual PA state that we want (especially if we don't remove previous events, as this will new `wait_broadcast_receive_state_ev` would just return immediately, right?

autopts/wid/bap.py
```diff
@@ -2181,11 +2181,21 @@ def hdl_wid_349(params: WIDParams):
...
 
     if ev is None:
         return False
 
+    if ev['pa_sync_state'] == PaSyncState.SYNC_INFO_REQ:
+        btp.bap_send_past(ev['src_id'])
+
+        ev = stack.bap.wait_broadcast_receive_state_ev(
+            broadcast_id, addr_type, addr, broadcaster_addr_type,
+            broadcaster_addr, WildCard(), timeout=10, remove=True)
```
sjanc: actually, it looks like returning from MMI right after issues bap_send_past() triggers some race in PTS and 2/10 I got inconclusive while current implementation has 100% pass rate in my setup

autopts/wid/bap.py
sjanc: yeah, I guess so,  however in this case test spec doesn't define what should happen after PAST so I'm not sure which state should I expect,  currently on logs I see another INFO_REQ and then NOT_SYNCED states

Thalley: Given that it is a non-BI test, I would assume that the expected PA sync state would be "Synced" :D

sjanc: :-D   it is "Synced"!   I checked wrong field in BTP logs :P

will update PR

autopts/wid/bap.py
```diff
@@ -2181,7 +2181,16 @@ def hdl_wid_349(params: WIDParams):
 
     ev = stack.bap.wait_broadcast_receive_state_ev(
         broadcast_id, addr_type, addr, broadcaster_addr_type,
-        broadcaster_addr, padv_sync, timeout=10, remove=False)
+        broadcaster_addr, padv_sync, timeout=10, remove=True)
+
+    if ev is None:
+        return False
+
+    btp.bap_send_past(ev['src_id'])
```
Thalley: Is this only called when PAST is enabled and support on both devices, or also on devices that doesn't support PAST?

autopts/wid/bap.py
```diff
@@ -2181,7 +2181,16 @@ def hdl_wid_349(params: WIDParams):
...
+
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, PaSyncState.SYNCED, timeout=10, remove=True)
```
Thalley: If the PA sync state from the previous event has `PaSyncState.SYNCED`, shouldn't we skip the PAST request and this event? 

Although I guess since `padv_sync` gets assigned `0x01` (`SYNC_INFO_REQ`), then PAST must have been requested, and this is all fine :D

Thalley: Might be worth setting `padv_sync = PaSyncState.SYNC_INFO_REQ` instead of `padv_sync = 0x01` to make it more clear :)

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (1)</summary>BAP     BAP/BA/BASS/BV-08-C             PASS</details>

autopts/wid/bap.py
sjanc: this WID is called only for BAP/BA/BASS/BV-08-C [Transfers SyncInfo Data to Scan Delegator]

autopts/wid/bap.py
sjanc: > If the PA sync state from the previous event has `PaSyncState.SYNCED`, shouldn't we skip the PAST request and this event?
> 
> Although I guess since `padv_sync` gets assigned `0x01` (`SYNC_INFO_REQ`), then PAST must have been requested, and this is all fine :D

ok, I guess it can be more flexible this way

PR #1677, opened 2026-01-21
common: Add initial support for multiple IUTs

mkasenberg:
CliParser parameters such as --tty-file or --hci can now accept multiple values, allowing AutoPTS to work with multiple IUTs. For now, all IUTs must operate in the same mode, but this limitation can be easily lifted in the future.

To enable test cases that require multiple IUTs, define them as custom test cases of a project and specify the number of IUTs to be used, for example:

ZTestCase(
    "CSIP", "CSIP/SR/SP/BV-06-C",
    cmds=pre_conditions + pre_conditions_iut2,
    generic_wid_hdl=csip_wid_hdl,
    iut_count=2
)

Use set_current_iutctl_id(iutctl_id=1) to switch the active IUT context. See the example at autopts/ptsprojects/zephyr/csip.py.

Design notes and challenges:
The get_iut() and get_stack() functions are widely used across WID handlers and BTP event handlers. Ideally, these functions should accept an explicit IUT ID, but introducing this would require a major refactor to propagate the ID through the entire call chain.

Since PTS currently defines only a small number of multi-IUT test cases, and to avoid a large refactor, an interim experimental solution based on ContextVar is used.

With this approach, get_iut() and get_stack() return different references depending on the value previously set in the current execution context via set_current_iutctl_id(iutctl_id=iut_id). As a result, threads such as MainThread, LTThread-*, and BTPWorkers must manage their active IUT context so that get_iut() and get_stack() resolve to the correct instance within their call stacks.

Another difficulty is that, at least for now, some WID messages do not specify the IUT ID for which the handler should be executed. For test cases marked as multi-IUT, we therefore assume that subsequent occurrences of the same WID are to be handled for subsequent IUTs.

### DIFF
--- a/autopts/bot/common.py
+++ b/autopts/bot/common.py
@@ -113,7 +113,7 @@ class BotConfigArgs(Namespace):
         self.hid_vid = args.get('hid_vid', None)
         self.hid_pid = args.get('hid_pid', None)
         self.hid_serial = args.get('hid_serial', None)
-        self.kernel_cpu = args.get('kernel_cpu', 'qemu_cortex_m3')
+        self.kernel_cpu = args.get('kernel_cpu', None)
         self.setcap_cmd = args.get('setcap_cmd', None)
         self.hci = args.get('hci', None)
         self.test_cases = args.get('test_cases', [])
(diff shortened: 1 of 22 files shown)


mkasenberg:
> Another difficulty is that, at least for now, some WID messages do not specify the IUT ID for which the handler should be executed.

I asked PTS support about this and they will add the "For IUT2." info to the end of a WID/MMI description, which solves our problem.

I wasn't able to provide them with a complete list of WIDs that need fixing, so please note that while implementing those multi-IUT test cases.

PR #1680, opened 2026-01-26
auto-pts: add bearer registration and improve WID handling

ciascai:
This update implements missing TBS/GTBS bearer registration with full metadata support, aligns GTBS behavior with PTS expectations, corrects BTP definitions to match Zephyr/PTS requirements.


autopts/pybtp/defs.py
```diff
@@ -992,6 +992,8 @@ def BIT(bit):
 BTP_SDP_CMD_SEARCH_ATTR_REQ = 0x04
 BTP_SDP_EV_SERVICE_RECORD_HANDLE = 0x80
 
+# Registration Bearer for TBS/GTBS
+BTP_TBS_CMD_REGISTER_BEARER = 0x0C
```
Thalley: Should be placed right after `BTP_TBS_CMD_TERMINATE_CALL` in this file :)

Thalley: Please also add the new command to https://github.com/auto-pts/auto-pts/blob/master/doc/btp_tbs.txt

autopts/pybtp/btp/tbs.py
```diff
@@ -56,8 +56,65 @@
...
                               CONTROLLER_INDEX),
+    'register_bearer':       (defs.BTP_SERVICE_ID_TBS,
+                              defs.BTP_TBS_CMD_REGISTER_BEARER,
+                              CONTROLLER_INDEX),
+}
+
+# GTBS registration command dictionary (mirroring TBS, but with GTBS service ID)
+GTBS = {
+    'register_bearer': (0x1c, defs.BTP_TBS_CMD_REGISTER_BEARER, CONTROLLER_INDEX),
 }
```
Thalley: BTP does not have a `GTBS` service, so we can just keep the TBS version, and have a field in the command to indicate if it is TBS or GTBS

autopts/pybtp/btp/tbs.py
```diff
@@ -56,8 +56,65 @@
...
+    Register a TBS bearer with all fields packed in one message.
+    Layout:
+        uint8_t  index
+        uint8_t  provider_name_len
+        uint8_t  provider_name[]
+        uint8_t  uci_len
+        uint8_t  uci[]
+        uint8_t  uri_scheme_list_len
+        uint8_t  uri_scheme_list[]
+        uint32_t supported_features
```
Thalley: Consider doing something like https://github.com/auto-pts/auto-pts/blob/master/doc/btp_cap.txt#L38-L57

i.e. when we have multiple variable-length fields like the name, UCI and URIs, we can place them in a single dynamically-length field at the end

i.e. this should probably be

```suggestion
    Layout:
        uint8_t  index
        uint32_t supported_features
        uint8_t  provider_name_len
        uint8_t  uci_len
        uint8_t  uri_scheme_list_len
        uint8_t  strings[]
```

Or something like that (not sure if `strings` is a good name). 

Then when the IUT decodes the message, it can/should assume that the first `provider_name_len` bytes of `strings` is the name, the next `uci_len` bytes are the UCI and the last `uri_scheme_list_len` bytes are the URI schemes

autopts/pybtp/btp/tbs.py
```diff
@@ -56,8 +56,65 @@
...
+    tbs_command_rsp_succ()
+
+def gtbs_register_bearer(index, bearer_type=0):
+    """Register a GTBS bearer (opcode 0x0C, service ID 0x1c)."""
+    logging.debug(f"{gtbs_register_bearer.__name__}")
+    data = bytearray(struct.pack("bb", index, bearer_type))
+    iutctl = get_iut()
+    iutctl.btp_socket.send(*GTBS['register_bearer'], data=data)
+    tbs_command_rsp_succ()
+def tbs_register_bearer_full(index, provider_name, uci, uri_scheme_list, supported_features):
```
Thalley: I think we should only have the `tbs_register_bearer_full`

It should additionally be expanded to also contain whether this is registering TBS or GTBS (see my other comment), and the `technology`.

The `authorization_required` field can be considered a Zephyr-specific value, and probably should not be part of this.. (In case this hasn't been mentioned: AutoPTS is a tool used by more projects than just Zephyr, and thus shouldn't be considered Zephyr-only)

autopts/wid/gtbs.py
```diff
@@ -31,3 +31,11 @@ def hdl_wid_1(params: WIDParams):
     # Example WID
 
     return True
+
+
+def hdl_wid_register_bearer(params: WIDParams):
```
Thalley: `wid`s are "commands" from PTS, and shouldn't be defined as "normal" functions.

The call to `tbs_register_bearer` could be done in e.g. `hdl_wid_20001` (similar to the VCP/VCS PR I shared with you earlier) before starting the advertising, or even as part of the `pre_conditions` in https://github.com/auto-pts/auto-pts/blob/master/autopts/ptsprojects/zephyr/tbs.py#L71-L99

The former follows existing approaches in AutoPTS, but is a breaking change for other platforms, where the latter approach would only affect Zephyr (given that https://github.com/auto-pts/auto-pts/blob/master/autopts/ptsprojects/zephyr/tbs.py is a Zephyr specific file)

autopts/pybtp/btp/tbs.py
```diff
@@ -56,9 +56,41 @@
     'terminate_call':        (defs.BTP_SERVICE_ID_TBS,
                               defs.BTP_TBS_CMD_TERMINATE_CALL,
                               CONTROLLER_INDEX),
+    'register_bearer':       (defs.BTP_SERVICE_ID_TBS,
+                              defs.BTP_TBS_CMD_REGISTER_BEARER,
+                              CONTROLLER_INDEX),
 }
 
 
+def tbs_register_bearer_full(index, provider_name, uci, uri_scheme_list, supported_features, gtbs, technology):
```
Thalley:
```suggestion
def tbs_register_bearer_full(provider_name, uci, uri_scheme_list, supported_features, gtbs, technology):
```

Consider reordering the parameters here to match the order of the command

autopts/wid/tbs.py
```diff
@@ -228,3 +228,19 @@ def hdl_wid_20142(params: WIDParams):
         btp.tbs_remote_incoming(0, ovs_receiver, ovs_caller, ovs_friendly)
 
     return True
+
+
+def hdl_wid_20143(params: WIDParams):
+    """Register TBS or GTBS bearer at the start of the test."""
```
Thalley: Is this a real WID request from PTS? 

I unresolved https://github.com/auto-pts/auto-pts/pull/1680#discussion_r2728020812 as that hasn't been properly resolved yet I think

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -86,7 +86,8 @@ def set_addr(addr):
         TestFunc(btp.core_reg_svc_tbs),
         TestFunc(lambda: pts.update_pixit_param(
             "TBS", "TSPX_iut_device_name_in_adv_packet_for_random_address", iut_device_name)),
-        TestFunc(stack.tbs_init)
+        TestFunc(stack.tbs_init),
+        TestFunc(lambda: __import__('autopts.pybtp.btp.tbs').pybtp.btp.tbs.register_gtbs_then_tbs())
```
Thalley: Suggest to move the import to top of file and just do something like

```suggestion
        TestFunc(btp.tbs.register_gtbs_then_tbs())
```

autopts/pybtp/btp/tbs.py
```diff
@@ -56,9 +56,66 @@
...
+    Register a TBS or GTBS bearer with all fields packed in one message.
+    Layout (matches Zephyr struct):
+        uint8_t  gtbs
+        uint8_t  technology
+        uint16_t optional_opcodes
+        uint8_t  provider_name_len
+        uint8_t  uci_len
+        uint8_t  uri_scheme_list_len
+        uint8_t  strings[] (provider_name + uci + uri_scheme_list)
+    """
```
Thalley:
```suggestion
def tbs_register_bearer(gtbs, technology, optional_opcodes, provider_name, uci, uri_scheme_list):
    """
    Register a TBS or GTBS bearer
    
    """
```

And add Python documentation for the parameters. This is not related to Zephyr, but rather to BTP (AutoPTS is an project independent on Zephyr). 

It doesn't look like there's any standardized way to provide documentation in AutoPTS (cc @sjanc ), but perhaps something like 

```
    """
    Finds matching versions for the given project repo hash
    params: csv_file_path - path to compatibility.csv
    params: project_repo - path to local project repo
    params: commit_hash - hash of the project repo
    """
```

Arguably outside the scope of this, but ideally the function would also have type hinting (https://docs.python.org/3/library/typing.html)

autopts/pybtp/btp/tbs.py
```diff
@@ -56,9 +56,66 @@
...
+    provider_name_bytes = provider_name.encode('utf-8')
+    uci_bytes = uci.encode('utf-8')
+    uri_scheme_list_bytes = uri_scheme_list.encode('utf-8')
+    strings = provider_name_bytes + uci_bytes + uri_scheme_list_bytes
+    data = bytearray()
+    data.append(gtbs)
+    data.append(technology)
+    data.extend(struct.pack('<H', optional_opcodes & 0xFFFF))
```
Thalley, line 87: I don't think we have a standardize way of doing this either, but I do think that raising an error is better than silently truncating values, e.g.

```python
    if (optional_opcodes < 0) or (optional_opcodes > 0xFFFF):
        raise ValueError(f"Invalid optional_opcodes value" {optional_opcodes}")
```

Thalley: Arguably this should be done for all parameters, but it is not a blocker as we have so many existing places where parameters are not properly validated

Thalley: Actually what happens if you do

```suggestion
    data.extend(struct.pack("H", optional_opcodes))
```

and optional_opcodes is > 0xFFFF? Does it fail to pack it?

autopts/pybtp/btp/tbs.py
```diff
@@ -56,9 +56,66 @@
...
+        provider_name="TBS",
+        uci="un000",
+        uri_scheme_list="tel,skype",
+        optional_opcodes=0x0001,
+        gtbs=0,
+        technology=4,
+    )
+    logging.info("TBS registered.")
+
+
```
Thalley: I would actually remove this, and then have users call `tbs_register_bearer_full` directly with the parameters they want. I think it makes it more explicit what is going on, and the caller can control the parameters more (which we need to do in the case where we need to set `optional_opcodes` to either support or not support join for the join tests

python
Thalley, line 1: What is this file?

Thalley (changes requested):
Overall looks good, with some comments

autopts/pybtp/btp/tbs.py
```diff
@@ -56,9 +56,55 @@
     'terminate_call':        (defs.BTP_SERVICE_ID_TBS,
                               defs.BTP_TBS_CMD_TERMINATE_CALL,
                               CONTROLLER_INDEX),
+    'register_bearer':       (defs.BTP_SERVICE_ID_TBS,
+                              defs.BTP_TBS_CMD_REGISTER_BEARER,
+                              CONTROLLER_INDEX),
 }
 
 
+def tbs_register_bearer_full(provider_name, uci, uri_scheme_list, optional_opcodes, gtbs, technology):
```
Thalley:
```suggestion
def tbs_register_bearer_full(gtbs, technology, optional_opcodes, provider_name, uci, uri_scheme_list):
```

Suggest to order the parameters in the same order they appear in the BTP command

Thalley:
```suggestion
def tbs_register_bearer(provider_name, uci, uri_scheme_list, optional_opcodes, gtbs, technology):
```

I'd say that `full` is implied :)

autopts/pybtp/btp/tbs.py
```diff
@@ -56,9 +56,55 @@
...
+    The function assembles the BTP register_bearer message with the
+    following packed layout:
+
+        uint8  gtbs
+        uint8  technology
+        uint16 optional_opcodes
+        uint8  provider_name_len
+        uint8  uci_len
+        uint8  uri_scheme_list_len
+        bytes  provider_name + uci + uri_scheme_list
```
Thalley: It's not _wrong_, but BTP is platform agnostic, and may be communicating with something written in python where types like `uint8` are "unknown". I think the description in `btp_tbs.txt` is enough, and any limitations to the parameters should be added in the above (`Params:`)

```suggestion
```

autopts/pybtp/btp/tbs.py
```diff
@@ -56,9 +56,55 @@
...
+def tbs_register_bearer_full(provider_name, uci, uri_scheme_list, optional_opcodes, gtbs, technology):
+    """
+    Registers a TBS or GTBS bearer over BTP using a packed payload.
+
+    Params:
+        provider_name (str)      - Bearer provider name string.
+        uci (str)                - Unified Call Identifier string.
+        uri_scheme_list (str)    - Comma-separated list of supported URI schemes.
+        optional_opcodes (int)   - Bitmask representing supported optional operations.
+        gtbs (int)               - 1 for GTBS, 0 for TBS.
```
Thalley:
```suggestion
        gtbs (bool)               - True GTBS, False for TBS.
```

Suggest to make this a boolean value, and then convert to `1` or `0` when creating the BTP command

autopts/pybtp/btp/tbs.py
Thalley: Did you check this one?

autopts/pybtp/btp/tbs.py
Thalley: The order of the parameters in the docstring should also be updated :)

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -86,9 +95,36 @@ def set_addr(addr):
...
 
+    for tc_name in test_case_name_list:
+        opcodes = get_optional_opcodes(tc_name)
+        cmds = pre_conditions + [
+            TestFunc(lambda opcodes=opcodes: tbs_register_bearer_full(
+                provider_name="Generic TBS",
+                uci="un000",
+                uri_scheme_list="tel,skype",
+                optional_opcodes=opcodes,
+                gtbs=1,
```
Thalley:
```suggestion
                gtbs=True,
```

See comment below

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -86,9 +95,36 @@ def set_addr(addr):
...
+                technology=3
+            )),
+            TestFunc(lambda opcodes=opcodes: tbs_register_bearer_full(
+                provider_name="TBS",
+                uci="un000",
+                uri_scheme_list="tel,skype",
+                optional_opcodes=opcodes,
+                gtbs=0,
```
Thalley:
```suggestion
                gtbs=False,
```

See comment below

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -68,6 +69,14 @@ def test_cases(ptses):
     def set_addr(addr):
         iut_addr.set(addr)
 
+    def get_optional_opcodes(tc_name):
+        if tc_name in ["GTBS/SR/CP/BV-10-C", "TBS/SR/CP/BV-10-C"]:
+            return 0x0001
+        return 0x0003
```
Thalley: A personal pet peeve of mine are magic numbers :D 

I suggest to add a definition of all possible Supported Opcodes values in https://github.com/auto-pts/auto-pts/blob/master/autopts/pybtp/types.py similar to the ones for e.e.g TMAPRole: https://github.com/auto-pts/auto-pts/blob/master/autopts/pybtp/types.py#L533-L539

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -86,9 +95,36 @@ def set_addr(addr):
...
+                gtbs=1,
+                technology=3
+            )),
+            TestFunc(lambda opcodes=opcodes: tbs_register_bearer_full(
+                provider_name="TBS",
+                uci="un000",
+                uri_scheme_list="tel,skype",
+                optional_opcodes=opcodes,
+                gtbs=0,
+                technology=4
```
Thalley: Similar to my request for Supported Opcodes, the possible values for technology should be defined in types.py as a IntEnum, so that this becomes something like `BearerTech.WIFI` instead of `4` which add some nice readability and context to the line

autopts/pybtp/types.py
```diff
@@ -543,3 +543,21 @@ class BAPAnnouncement:
...
+class BearerTech(IntEnum):
+    Tech3G = 0x01
+    Tech4G = 0x02
+    TechLTE = 0x03
+    TechWIFI = 0x04
+    Tech5G = 0x05
+    TechGSM = 0x06
+    TechCDMA = 0x07
+    Tech2G = 0x08
+    TechWCDMA = 0x09
```
Thalley: I would omit `Tech` for each definition as `BearerTech.TechX` is superfluous, but that of course gives an issue with things like 3G, 4G, etc. 

I don't have a great solution for this. Maybe we can do CELL_3G, CELL_4G, etc.?

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -68,6 +69,14 @@ def test_cases(ptses):
     def set_addr(addr):
         iut_addr.set(addr)
 
+    def get_optional_opcodes(tc_name):
+        if tc_name in ["GTBS/SR/CP/BV-10-C", "TBS/SR/CP/BV-10-C"]:
+            return OptionalOpcode.LOCAL_HOLD
+        return OptionalOpcode.ALL
```
Thalley: It is OK as is, but I am wondering if this makes more sense:

```suggestion
    def get_optional_opcodes(tc_name):
        optional_opcodes = OptionalOpcode.ALL
        
        if tc_name in ["GTBS/SR/CP/BV-10-C", "TBS/SR/CP/BV-10-C"]:
            # Disable join for these tests 
            optional_opcodes = optional_opcodes & ~OptionalOpcode.JOIN
            
        return optional_opcodes
```

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -68,6 +69,14 @@ def test_cases(ptses):
...
+
+    test_case_name_list = pts.get_test_case_list('TBS')
```
Thalley: Does this work for both `TBS` and `GTBS`?

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -86,9 +95,36 @@ def set_addr(addr):
...
+                uci="un000",
+                uri_scheme_list="tel,skype",
+                optional_opcodes=opcodes,
+                gtbs=False,
+                technology=BearerTech.TechWIFI
+            )),
+        ]
+        instance = ZTestCase("TBS", tc_name,
+                             cmds=cmds,
+                             generic_wid_hdl=tbs_wid_hdl)
```
Thalley:
```suggestion
        instance = ZTestCase("TBS", tc_name, cmds=cmds, generic_wid_hdl=tbs_wid_hdl)
```

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -86,9 +95,36 @@ def set_addr(addr):
...
+        tc_list.append(instance)
+
+    return tc_list
+
```
Thalley: The return here makes the following 10 lines unreachable

autopts/ptsprojects/zephyr/tbs.py
ciascai: yes

autopts/pybtp/types.py
```diff
@@ -543,3 +543,21 @@ class BAPAnnouncement:
...
+
+class BearerTech(IntEnum):
+    Cell_3G = 0x01
+    Cell_4G = 0x02
+    LTE = 0x03
+    WIFI = 0x04
+    Cell_5G = 0x05
+    GSM = 0x06
+    CDMA = 0x07
+    Cell_2G = 0x08
```
Thalley:
```suggestion
    CELL_3G = 0x01
    CELL_4G = 0x02
    LTE = 0x03
    WIFI = 0x04
    CELL_5G = 0x05
    GSM = 0x06
    CDMA = 0x07
    CELL_2G = 0x08
```

Enums and constants in Python are usually all-caps :)

autopts/ptsprojects/zephyr/tbs.py
```diff
@@ -68,6 +69,18 @@ def test_cases(ptses):
...
+        optional_opcodes = OptionalOpcode.ALL
+
+        if tc_name in ["GTBS/SR/CP/BV-10-C", "TBS/SR/CP/BV-10-C"]:
+            # Disable join for these tests
+            optional_opcodes = optional_opcodes & ~OptionalOpcode.JOIN
+
+        return optional_opcodes
+
+    test_case_name_list = pts.get_test_case_list('TBS')
+    tc_list = []
```
Thalley: Minor: Consider moving these back to where they were before

Thalley (changes requested):
Minor comments, otherwise LGTM. 

@sjanc Ideally this change only directly affects Zephyr at the moment to avoid breaking other platforms. Do we have any other platforms that support LE Audio that we can run the bot on to verify this?

Thalley:
Zephyr PR has been merged; will merge this when/if the tests pass :)

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (2)</summary>GTBS    GTBS/SR/CP/BV-10-C              PASS<br>TBS     TBS/SR/CP/BV-10-C               PASS</details>

PR #1692, opened 2026-02-12
Zephyr: iut_config: Add br_l2cap_ext_feat.conf

lylezhu2012:
Add new configuration for L2CAP extended features testing with BR/EDR support. This configuration enables connectionless L2CAP channels and is specifically used for the L2CAP/EXF/BV-07-C test case.

The configuration:
- Uses prj_br.conf as the base overlay
- Enables CONFIG_BT_L2CAP_CONNLESS for connectionless L2CAP support

### DIFF
--- a/autopts/bot/iut_config/zephyr.py
+++ b/autopts/bot/iut_config/zephyr.py
@@ -437,6 +437,18 @@ iut_config = {
             'GAP/PADV/PASE/BV-12-C',
         ]
     },
+
+    "br_l2cap_ext_feat.conf": {
+        "pre_overlay": "prj_br.conf",
+        "overlay": {
+            'CONFIG_BT_L2CAP_CONNLESS': 'y',
+        },
+        "test_cases": [
+            'L2CAP/EXF/BV-07-C',
+            'L2CAP/EXF/BV-08-C',
+            'L2CAP/FIX/BV-03-C',
+        ]
+    },
 }
 
 retry_config = {


autopts/bot/iut_config/zephyr.py
```diff
@@ -400,6 +400,28 @@
             'GAP/SEC/SEM/BV-35-C',
         ]
     },
+
+    "br_l2cap_ext_feat.conf": {
+        "overlay": {
```
sjanc, line 443: hint: you can use pre_overlay feature to include bredr overlay from tester repo for common settings

autopts/bot/iut_config/zephyr.py
```diff
@@ -400,6 +400,28 @@
...
+            'CONFIG_BT_L2CAP_DYNAMIC_CHANNEL': 'y',
+            'CONFIG_BT_PAGE_TIMEOUT': '0xFFFF',
+            'CONFIG_BT_L2CAP_RET': 'y',
+            'CONFIG_BT_L2CAP_FC': 'y',
+            'CONFIG_BT_L2CAP_ENH_RET': 'y',
+            'CONFIG_BT_L2CAP_STREAM': 'y',
+            'CONFIG_BT_L2CAP_FCS': 'y',
+            'CONFIG_BT_L2CAP_EXT_WIN_SIZE': 'y',
+            'CONFIG_BT_L2CAP_MAX_WINDOW_SIZE': '5',
+            'CONFIG_BT_L2CAP_CONNLESS': 'y',
```
sjanc, line 444: why not just enable CONFIG_BT_L2CAP_CONNLESS in prj_br.conf instead?  does it break any other tests?

we did that for LE GAP where we enable all supported features (privacy being notable exception for now) and use overlays only if necessary

Thalley: > we did that for LE GAP where we enable all supported features (privacy being notable exception for now) and use overlays only if necessary

I do also recall us talking about making some of the current compile-time options run-time configurable so we can reduce overlays etc. For example adding a `bt_gap_set_privacy(bool on)` or something like that. Possibly hiding that behind `BT_TEST` or something. Overall it would make things a lot simpler if we could produce a single binary for all tests

sjanc: yeap, although since we switched to native build this is less pressing now :P  simply because how fast this starts comparing to re-flashing process of real HW

but yes, if privacy and GAP security levels could be enforced in runtime we would be able to get rid of most 'host' overlays

lylezhu2012: > why not just enable CONFIG_BT_L2CAP_CONNLESS in prj_br.conf instead? 

From Zephyr Classic L2CAP perspective, the kconfig `CONFIG_BT_L2CAP_CONNLESS ` can be enabled in prj_br.conf by default.

> does it break any other tests?

Yes, there are some issues. When CONFIG_BT_L2CAP_CONNLESS is enabled, if the feature ( Extended feature No. 9) also be set by dongle, the request `L2CAP_INFORMATION_REQ (info type : Connectionless MTU)` will be sent by Zephyr to require the Connectionless MTU of peer device. But in some cases (these cases are not related to the feature `Connectionless`), the dongle will not reply to the request `L2CAP_INFORMATION_REQ`.


> I do also recall us talking about making some of the current compile-time options run-time configurable so we can reduce overlays etc. For example adding a `bt_gap_set_privacy(bool on)` or something like that. Possibly hiding that behind `BT_TEST` or something. Overall it would make things a lot simpler if we could produce a single binary for all tests

Yes, you're referring to features that can be dynamically controlled. But for `Unicast Connectionless Data Reception`, it should not be able to be modified dynamically.

Thalley: > yeap, although since we switched to native build this is less pressing now :P simply because how fast this starts comparing to re-flashing process of real HW
> 
> but yes, if privacy and GAP security levels could be enforced in runtime we would be able to get rid of most 'host' overlays

The main issue I have with these overlays, is that it is easy to forget to add them when testing locally (unless using the bot as a CLI ;) ). Sometimes when you are debugging a test locally, you may not be aware that a specific overlay was added for that test, and then your local results may differ significantly from the weekly runs. Having everything in the same binary would make it more consistent

lylezhu2012: Should there be a system boot configuration for features that cannot be dynamically modified? This would allow the system to determine whether to enable or disable the feature during startup.

In this way, one binary is achieved.

sjanc (approved):
I'm fine with overlay for time being, as discussed in long term we will be working on reducing those anyway

however if enabling this doesn't affect other BR/EDR test I'd consider simply putting this in `prj_br.conf`

This is what we are now doing for LE part.


So either rebase if you want to go with overlay for now, or update `prj_br.conf` instead

lylezhu2012:
> So either rebase if you want to go with overlay for now, or update `prj_br.conf` instead

For now, due to the PTS behaviour (maybe it is an issue), the the configuration `CONFIG_BT_L2CAP_CONNLESS` cannot be configured in `prj_br.conf`. I will rerun all l2cap test cases with the setting `CONFIG_BT_L2CAP_CONNLESS=y`, and report all PTS-related issues. Having resolved all related issues, I will remove `br_l2cap_ext_feat.conf`.

PR #1693, opened 2026-02-12
L2CAP: Improve WID 265 handler for L2CAP/CLS/CID/BV-01-C test case

lylezhu2012:
Update hdl_wid_265 to properly handle L2CAP/CLS/CID/BV-01-C test case by disconnecting all L2CAP channels when confirming no data was received.

Changes:
- Accept WIDParams parameter instead of ignoring it
- Add test case specific handling for L2CAP/CLS/CID/BV-01-C
- Disconnect all active L2CAP channels using _safe_l2cap_disconnect

The channel disconnection is required by the test case. And with the fix, it can reduce the execution time consumption of the test case.

### DIFF
--- a/autopts/wid/l2cap.py
+++ b/autopts/wid/l2cap.py
@@ -1164,9 +1164,14 @@ def hdl_wid_33(_: WIDParams):
     return True
 
 
-def hdl_wid_265(_: WIDParams):
+def hdl_wid_265(params: WIDParams):
     '''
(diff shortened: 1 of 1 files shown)


sjanc (approved):
btw we are working on limiting this type of always-confirm-true MMIs for proper IUT behavior validation

no need to do it in this PR but please consider that for future

PR #1707, opened 2026-02-25
bluez: Implement specific BAP hdl_wid_302()

fdanis-oss:
By default and if audio endpoints are registered before the tests starts BlueZ will request codec and QoS configs before PTS has sent them.
To mitigate this, the endpoints are registered only on btp.ascs_config_qos() call so we are sure that codec and QoS configs has been stored.

This version of hdl_wid_302() calls btp.ascs_config_qos() with fake values to trigger the endpoint creation.

### DIFF
--- a/autopts/ptsprojects/bluez/bap_wid.py
+++ b/autopts/ptsprojects/bluez/bap_wid.py
@@ -14,8 +14,20 @@
 #
 
 import logging
+import re
 
+from autopts.ptsprojects.stack import get_stack
+from autopts.pybtp import btp
+from autopts.pybtp.types import (
+    FRAME_DURATION_STR_TO_CODE,
+    QOS_CONFIG_SETTINGS,
+    SAMPLING_FREQ_STR_TO_CODE,
+    AudioDir,
+    WIDParams,
+    create_lc3_ltvs_bytes,
+)
 from autopts.wid import generic_wid_hdl
+from autopts.wid.bap import create_default_config
 
 log = logging.debug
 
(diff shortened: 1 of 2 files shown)


autopts/ptsprojects/bluez/bap_wid.py
```diff
@@ -14,12 +14,91 @@
...
+    """
+
+    numbers = re.findall(r'\d+(?:\.\d+)?', params.description)
+    sampling_freq = numbers[0]
+    frame_duration = numbers[1]
+
+    if 'SINK' in params.description:
+        audio_dir = AudioDir.SINK
+    else:
+        audio_dir = AudioDir.SOURCE
```
Thalley, line 53: Is the docstring correct?

If PTS sends
```
    """
    Please configure ASE state to CODEC configured with SINK/SOURCE ASE,
    Freq: X KHz, Frame Duration: X ms
    """
```

Then we always set dir to SINK; or is either `SINK` or `SOURCE` instead of `SINK/SOURCE`?

autopts/ptsprojects/bluez/bap_wid.py
```diff
@@ -14,12 +14,91 @@
...
+    stack.ascs.wait_ascs_operation_complete_ev(addr_type, addr, ase_id, 30)
+
+    # Codec and QoS configuration are triggered automatically by BlueZ as soon as a
+    # local audio endpoint can be used, i.e. before PTS has requested the WID 302.
+    # To prevent this BlueZ's btpclient only registers the audio endpoint after receiving
+    # the QoS configuration.
+    # As this WID is meant to verify that the IUT can be configured with the requested Codec
+    # only this calls the QoS configuration with fake parameters to trigger the endpoint
+    # registration after the Codec configuration has been stored.
+    qos_config = QOS_CONFIG_SETTINGS['8_1_1']
```
Thalley, line 92: Does it matter that the QoS config here may not match the sampling frequency and frame duration that is in the request?

Thalley (changes requested):

PR #1709, opened 2026-02-27
Bluez: Add support for external audio server

fdanis-oss:
This allows to run the PTS test with BlueZ standalone or with PipeWire and WirePlumber.

It also adds specific WirePlumber congifurations as some config QoS tests requests to limit properties selected by
PipeWire/WirePlumber to "low-latency" or "high-reliability".  
This can be achieved by using the "bluez5.bap.force-target-latency" property.

### DIFF
--- a/autopts/ptsprojects/bluez/bap.py
+++ b/autopts/ptsprojects/bluez/bap.py
@@ -19,8 +19,9 @@
 from autopts.client import get_unique_name
 from autopts.ptsprojects.bluez.bap_wid import bap_wid_hdl
 from autopts.ptsprojects.bluez.btestcase import BTestCase
+from autopts.ptsprojects.bluez.iutctl import get_iut
 from autopts.ptsprojects.stack import get_stack
-from autopts.ptsprojects.testcase import TestFunc
+from autopts.ptsprojects.testcase import TestFunc, TestFuncCleanUp
 from autopts.pybtp import btp
 from autopts.pybtp.types import Addr, IOCap
 from autopts.utils import ResultWithFlag
@@ -91,10 +92,15 @@ def test_cases(ptses):
 
     stack.gap_init()
 
+    iut = get_iut()
     iut_addr = ResultWithFlag()
 
     def set_addr(addr):
         iut_addr.set(addr)
+
+    def _stop_audio():
+        iut.set_audio_profile(None)
+        iut.stop_audio()
 
     pre_conditions = [
         TestFunc(btp.core_reg_svc_gap),
@@ -115,6 +121,8 @@ def test_cases(ptses):
         TestFunc(lambda: stack.bap.set_broadcast_code(BROADCAST_CODE)),
         TestFunc(lambda: set_addr(
             stack.gap.iut_addr_get_str())),
+        TestFunc(iut.start_audio),
+        TestFuncCleanUp(_stop_audio),
         ]
 
     custom_test_cases = [
--- a/autopts/ptsprojects/bluez/bap.py
+++ b/autopts/ptsprojects/bluez/bap.py
@@ -126,6 +126,50 @@ def test_cases(ptses):
         ]
 
     custom_test_cases = [
+        BTestCase("BAP", "BAP/UCL/SCC/BV-038-C",
+                  cmds=[TestFunc(lambda: iut.set_audio_profile("low-latency"))] +
+                        pre_conditions,
+                  generic_wid_hdl=bap_wid_hdl),
+        BTestCase("BAP", "BAP/UCL/SCC/BV-042-C",
+                  cmds=[TestFunc(lambda: iut.set_audio_profile("low-latency"))] +
+                        pre_conditions,
+                  generic_wid_hdl=bap_wid_hdl),
+        BTestCase("BAP", "BAP/UCL/SCC/BV-046-C",
+                  cmds=[TestFunc(lambda: iut.set_audio_profile("low-latency"))] +
+                        pre_conditions,
+                  generic_wid_hdl=bap_wid_hdl),
+        BTestCase("BAP", "BAP/UCL/SCC/BV-048-C",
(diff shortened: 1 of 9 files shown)


autopts/ptsprojects/bluez/bap.py
```diff
@@ -110,14 +113,61 @@ def set_addr(addr):
...
+                        pre_conditions,
+                  generic_wid_hdl=bap_wid_hdl),
+        BTestCase("BAP", "BAP/UCL/SCC/BV-042-C",
+                  cmds=[TestFunc(lambda: iut.set_audio_profile("low-latency"))] +
+                        pre_conditions,
+                  generic_wid_hdl=bap_wid_hdl),
+        BTestCase("BAP", "BAP/UCL/SCC/BV-046-C",
+                  cmds=[TestFunc(lambda: iut.set_audio_profile("low-latency"))] +
+                        pre_conditions,
+                  generic_wid_hdl=bap_wid_hdl),
```
Thalley, line 140: Is "low-latency" and "high-reliability" really good enough to pass these tests? 

They are very specific on which codec configuration and QoS settings to use: 
```
BAP/UCL/SCC/BV-046-C [UCL SRC Config QoS, LC3 48_2_1]
```

autopts/ptsprojects/bluez/bap_wid.py
```diff
@@ -23,3 +36,80 @@
...
+    codec_ltvs_bytes = create_lc3_ltvs_bytes(sampling_freq, frame_duration,
+                                             audio_locations, octets_per_frame,
+                                             frames_per_sdu)
+    btp.ascs_config_codec(ase_id, coding_format, 0x0000, 0x0000, codec_ltvs_bytes)
+    stack.ascs.wait_ascs_operation_complete_ev(addr_type, addr, ase_id, 30)
+
+    # Send fake QoS config to trigger endpoint registration
+    qos_config = QOS_CONFIG_SETTINGS['8_1_1']
+    btp.ascs_config_qos(ase_id, 0, 0, *qos_config, 0)
+    stack.ascs.wait_ascs_operation_complete_ev(addr_type, addr, ase_id, 30)
```
Thalley: You may want to elaborate on this. What makes `btp.ascs_config_qos(ase_id, 0, 0, *qos_config, 0)` a fake QoS config?

autopts/ptsprojects/bluez/bap_wid.py
```diff
@@ -23,3 +36,80 @@
...
+
+
+def hdl_wid_303(params: WIDParams):
+    """
+    Please configure ASE state to QoS Configured with 16_2_1 in SINK direction.
+    """
+
+    if get_iut().external_audio is not None:
+        logging.debug("External audio supported, skipping WID 303 handling")
+        return False
```
Thalley: Wouldn't returning `False` here fail the test?

autopts/ptsprojects/bluez/bap.py
fdanis-oss: Yes, PipeWire will select the right configuration based on rate, duration frame min and max (e.g. 48_2), but doesn't have enough info during this test to select between low-latency (48_2_1) and high-availability (48_2_2).

autopts/ptsprojects/bluez/bap_wid.py
fdanis-oss: This is fake because the values doesn't mean anything and will not be used.

autopts/ptsprojects/bluez/bap_wid.py
fdanis-oss: No, as far from my tests.  
But I'm not sure to understand the effect of the return value.

Thalley: @sjanc can correct me if I'm wrong, but returning False (or any negative/false-like value) should result in the same as clicking "No" or "Cancel" in the PTS GUI which should fail the test, without knowing the specifics

autopts/ptsprojects/bluez/bap_wid.py
Thalley: For someone who isn't familiar with Bluez, that statement (and the current comment) doesn't really say much :)

autopts/ptsprojects/bluez/bap_wid.py
sjanc: this is what we have seen so far, however this may depend on case-by-case as it is up to PTS to decide what to do with returned MMI, also it may depends on MMI type

however in this case I'm surprised it doesn't fail test

fdanis-oss: In any case it will be better to change it to "True"

autopts/ptsprojects/bluez/bap_wid.py
```diff
@@ -23,3 +26,16 @@
...
+    """
+    Please configure ASE state to CODEC configured with SINK/SOURCE ASE,
+    Freq: X KHz, Frame Duration: X ms
+    """
+
+    if get_iut().external_audio is not None:
+        logging.debug("External audio supported, skipping WID 302 handling")
+        return True
+
+    return autopts.wid.bap.hdl_wid_302(params)
```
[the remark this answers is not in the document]
fdanis-oss: Other values can be used to only skip normal handling, e.g. during development phase when creating the endpoints using `bluetoothctl`.

fdanis-oss:
@sjanc Hi, is there anything preventing to merge this?

Thalley:
@fdanis-oss You need to rebase the PR :) 

To merge this, someone has to approve it first ^_^

Thalley (approved):
No objections from my side, but I also do not have a stake in this code :)

fdanis-oss:
@sjanc Is there any other changes requested before merging this?

sjanc (approved):

PR #1713, opened 2026-03-02
Extend periodic adv test coverage

m-alperen-sener:
Fixing (adding support) for the following PAwR test cases:

GAP/PADV/PASM/BV-02-C
GAP/PADV/PAM/BV-02-C
GAP/PADV/PAST/BV-03-C
GAP/PADV/PAST/BV-04-C
GAP/PADV/PASE/BV-07-C
GAP/PADV/PASE/BV-08-C
GAP/PADV/PASE/BV-10-C
GAP/PADV/PASE/BV-12-C

Related Zephyr Tester Changes: https://github.com/zephyrproject-rtos/zephyr/pull/104776


doc/btp_gap.txt
```diff
@@ -529,6 +529,12 @@ Commands and responses:
                 Command parameters:     Flags (1 octet)
                                         Interval Min (2 octets)
                                         Interval Max (2 octets)
+                                        Number of Subevents (1 octet)
```
sjanc: This will require V2 command

m-alperen-sener: Yeap, I created a new one; BTP_GAP_CMD_PAWR_CONFIGURE

autopts/bot/iut_config/zephyr.py
```diff
@@ -385,6 +385,25 @@
             'L2CAP/CLS/CID/BV-01-C',
         ]
     },
+
+    "ble_pawr.conf": {
+        "overlay": {
+            'CONFIG_BT_PER_ADV_RSP': 'y',
+            'CONFIG_BT_PER_ADV_SYNC_RSP': 'y',
+        },
```
Thalley, line 402: I would ideally want to avoid this (in fact, I would prefer to be able to remove this entire file...). 

Have you considered moving this to the BT Tester config files instead, but only apply it to the boards that support it (which in this case would only be the native_posix board I guess since the ZLL does not support this). 

Alternatively this may be better placed in the NCS fork of Zephyr so that it's only enabled when SDC is used

autopts/ptsprojects/stack/layers/gap.py
```diff
@@ -170,6 +170,8 @@ def __init__(self, name, manufacturer_data, appearance, svc_data, flags,
 
         # if IUT doesn't support it, it should be disabled in preconditions
         self.pair_user_interaction = True
+        self.periodic_report_data_rxed = False
+        self.periodic_report_data = None
```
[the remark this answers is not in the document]
Thalley: `periodic_data` is used though but is not initialized here; typo? 

What sets `periodic_data` and  `periodic_report_data` apart? Could you please add documentation for each new field added here?

autopts/ptsprojects/stack/layers/gap.py
```diff
@@ -230,6 +232,16 @@ def wait_periodic_report(self, timeout):
 
         return False
 
+    def wait_periodic_adv_data(self, timeout):
```
Thalley: Misleading name since it's not related to `periodic_data` IMO

autopts/pybtp/btp/gap.py
```diff
@@ -1590,16 +1601,27 @@ def create_ead_adv(payload):
     return tuple_data
 
 
-def gap_padv_configure(include_tx_power, intvl_min, intvl_max):
-    logging.debug("%s", gap_padv_configure.__name__)
+def gap_padv_configure(include_tx_power, intvl_min, intvl_max,
+                       with_responses=False):
```
```suggestion
def gap_padv_configure(include_tx_power, intvl_min, intvl_max, with_responses=False):
```

Should fit on 127 line width ;)

autopts/pybtp/btp/gap.py
```diff
@@ -1590,16 +1601,27 @@ def create_ead_adv(payload):
...
-    data_ba = bytearray(struct.pack("<BHH", include_tx_power,
-                                    intvl_min, intvl_max))
 
-    iutctl.btp_socket.send(*GAP['padv_configure'], data=data_ba)
+    data_ba = bytearray(struct.pack("<BHH", include_tx_power, intvl_min, intvl_max))
+
+    if with_responses:
+        # number of subevents = 10, subevent interval = 15, response slot delay = 1,
+        # response slot spacing = 4, number of response slots = 5
+        sub_event_cfg = bytearray(struct.pack("<BBBBB", 10, 15, 1, 4, 5))
```
Thalley: Suggest to put the value into variables with reasonable names instead of using literal values with documentation. The variable names should be documentation enough, wouldn't you say?

autopts/wid/gap.py
```diff
@@ -1711,8 +1711,17 @@ def hdl_wid_267(_: WIDParams):
...
-    if stack.gap.periodic_data:
+
+    test_cases_with_responses = [
+        'GAP/PADV/PASM/BV-02-C',
+        'GAP/PADV/PAM/BV-02-C'
+    ]
+
+    with_responses = params.test_case_name in test_cases_with_responses
+
+    btp.gap_padv_configure(0, 150, 200, with_responses=with_responses)
```
Thalley, line 1733: No pixit values to think about here?

autopts/wid/gap.py
```diff
@@ -1799,7 +1816,9 @@ def hdl_wid_308(_: WIDParams):
     # Click OK when IUT is ready to perform Periodic Advertising Synchronization
     # Establishment Procedure with listening for periodic advertising events.
 
-    btp.gap_padv_sync_transfer_recv(0, 10, 0)
+    # PTS uses 600x1.25 ms = 750 ms for Periodic Advertising Interval (0x0258)
+    # So it must be at least 750 ms sync timeout. (skip + 1) * interval.
+    btp.gap_padv_sync_transfer_recv(0, 100, 0)
```
Thalley: Is the PA interval set by a PIXIT value? If so, please do use that and calculate an appropriate timeout value. If not, even if PTS uses 750 PA interval, doesn't this set a timeout of 100 * 10ms => 1000? That seems quite low. Suggest to have the timeout to be at least a few intervals (e.g. 6) to make it more stable

autopts/wid/gap.py
```diff
@@ -1810,6 +1829,41 @@ def hdl_wid_309(_: WIDParams):
...
+    # advertising subevent data.
+    stack = get_stack()
+
+    # WID 303 will be called after this WID. WID 303 should start synchronization
+    # procedure as for GAP/PADV/PASE/BV-01-C "without listening for Periodic Advertising".
+    # Otherwise we need to add exception here for flags parameter.
+    if params.test_case_name in ['GAP/PADV/PASE/BV-07-C']:
+        return True
+
+    btp.gap_padv_create_sync(0, 0, 100, 0)
```
Thalley: Ditto w.r.t. timeout

autopts/pybtp/btp/gap.py
```diff
@@ -362,6 +363,16 @@ def gap_padv_sync_lost_ev_(gap, data, data_len):
 def gap_padv_report_ev_(gap, data, data_len):
     logging.debug("%s", gap_padv_report_ev_.__name__)
     stack = get_stack()
+
+    # Extract the data length from the header
+    _, _, _, _, _, length = struct.unpack_from('<HBBBBB', data[:7])
+
+    padv_data = data[7:]
```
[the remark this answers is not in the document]
m-alperen-sener: added the check against data size and header.

autopts/ptsprojects/stack/layers/gap.py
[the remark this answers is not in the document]
m-alperen-sener: corrected the usage and started using correct data variable and added explanation what event flag is somewhat duplicated.

autopts/wid/gap.py
```diff
@@ -1810,6 +1829,41 @@ def hdl_wid_309(_: WIDParams):
...
+def hdl_wid_313(params: WIDParams):
+    # description: Please confirm IUT received following data.
+    data = params.description.split('\n')[-1]
+    data = data.replace('-', '')
+    log(f'Expected data: {data}')
+
+    stack = get_stack()
+
+    if not stack.gap.wait_periodic_adv_data(10):
+        log('Failed to periodic advertising data')
```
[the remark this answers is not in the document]
m-alperen-sener: done

doc/btp_gap.txt
```diff
@@ -892,6 +893,28 @@ Commands and responses:
 
                 In case of an error, the error response will be returned.
 
+        Opcode 0x34 - Configure Periodic Advertising with Responces command/response
```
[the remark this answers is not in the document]
m-alperen-sener: done

doc/btp_gap.txt
```diff
@@ -892,6 +893,28 @@ Commands and responses:
...
+                                        Interval Max (2 octets)
+                                        Number of Subevents (1 octet)
+                                        Subevent Interval (1 octet)
+                                        Response Slot Delay (1 octet)
+                                        Response Slot Spacing (1 octet)
+                                        Number of Response Slots (1 octet)
+
+                This command is used to configure Periodic Advertising instance with
+                Pesponses (PAwR).
+                IUT is expected to have 1 Periodic Advertising instance.
```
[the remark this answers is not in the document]
m-alperen-sener: done

autopts/bot/iut_config/zephyr.py
m-alperen-sener: I agree, I don't like this file but currently it is the only way to tell that default zephyr tester app by default does not support those configs, I can not see any alternative. Otherwise this information will be lost:

Even though ZLL does not support PAwR, zephyr host does, so it should be visible somehow.

autopts/ptsprojects/stack/layers/gap.py
m-alperen-sener: it is autocomplete mistake I did; corrected by changing clearer names and description.

autopts/wid/gap.py
m-alperen-sener: Nope; I looked around, there is no IXIT for this in core spec.
Ref: Core.IXIT.p24.xlsx

autopts/wid/gap.py
m-alperen-sener: Initially. when I used 1 seconds it was quite stable, among lots of BLE traffic. But in case, I have increased the time to 5 seconds. 

Also this everywhere we set sync timeout; as I can see from the PTS logs it uses min:100ms and max 750ms, So I set the sync timeout to 5 second; which is the worst case sync as soon as sync happens tests continues, so it is not a constant delay added to the test routine.

autopts/wid/gap.py
m-alperen-sener: Ditto w.r.t. timeout

autopts/ptsprojects/stack/layers/gap.py
```diff
@@ -170,7 +170,19 @@ def __init__(self, name, manufacturer_data, appearance, svc_data, flags,
...
+
+        # During test multiple periodic advertising events can be received,
+        # and it would be difficult to trigger wait_event_ again and again
+        # until we get the expected event with periodic advertising data.
+        # also test case might be waiting both type of events in different
+        # steps. Therefore we use a different event flags.
+
+        # Periodic Advertising events with periodic advertising data.
+        self.periodic_adv_report_data_rxed = False
+        self.periodic_adv_report_data = None
```
Thalley: Why do we need both? Won't we be able to just check if `periodic_adv_report_data == None`?

autopts/ptsprojects/stack/layers/gap.py
```diff
@@ -220,12 +232,32 @@ def wait_for_peer_car_status(self, timeout):
...
+        '''
+        self.periodic_adv_report_rxed = True
+
+        if data:
+            self.periodic_adv_report_data = data
+            self.periodic_adv_report_data_rxed = True
+
     def wait_periodic_report(self, timeout):
-        if self.periodic_report_rxed:
+        if self.periodic_adv_report_rxed:
```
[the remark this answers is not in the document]
m-alperen-sener: this is the expected behavior; there might be previously received event even before we start waiting.

autopts/ptsprojects/stack/layers/gap.py
```diff
@@ -220,12 +232,32 @@ def wait_for_peer_car_status(self, timeout):
...
+            return True
+
+        if wait_for_event(timeout, lambda: self.periodic_adv_report_rxed):
+            self.periodic_adv_report_rxed = False
+            return True
+
+        return False
+
+    def wait_periodic_adv_report_with_data(self, timeout):
+        if self.periodic_adv_report_data_rxed:
```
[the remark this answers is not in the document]
m-alperen-sener: this is the expected behavior; there might be previously received event even before we start waiting.

autopts/ptsprojects/stack/layers/gap.py
m-alperen-sener: Against race condition happening:
when we receive the WID and start waiting the "event with data": we might have already received an event with data and another event without data after it, I saw that happening in the logs, luckily actually, even before implementing WID 313 . To handle this either I needed to create a a buffer to push whatever comes and wait with constant delay assuming expected data has arrived and somewhere in the buffer. Like using [wait_event_with_condition](https://github.com/auto-pts/auto-pts/blob/master/autopts/ptsprojects/stack/common.py#L73) with which a TODO and a sleep in it, so I chose not to use it but do the lamest and cleanest way I think: just act as if "PADV" and "PADV" with data are just two separate events.

It can bee seen in the PTS logs; data is given to controller to be sent before MMI is triggered

> +5812 ms
> Send Event               HCI!HCI_LE_SET_PERIODIC_ADV_SUBEVENT_DATA{
>                            adv_handle=0x00,
>                            num_subevents=0x01,
>                            {
>                              
> [subevent=0x00,
>                              response_slot_start=0x00,
>                              response_slot_count=0x01,
>                              subevent_data_length=0x1B,
>                              subevent_data=1A095054532D4741502D504144562D433030374538324242313443],
>                              
> 
>                            }
>                          }
> 
> 
> +5812 ms
> Receive Event            HCI?HCI_LE_SET_PERIODIC_ADV_SUBEVENT_DATA_COMPLETE_EVENT{
>                            status=HCI_OK,
>                            adv_handle=0x00
>                          }
> 
> 
> +5812 ms
> Preliminary Verdict:     PASS
> 
> 
> 
> +5812 ms
> User Action Required:    MMI_313 (MMI_Style_Ok_Cancel1=0x00011041): Please confirm IUT received following data.
> 
> 1A09-5054-532D-4741-502D-5041-4456-2D43-3030-3745-3832-4242-3134-43

Thalley: Hmm, the current function is

```python
    def periodic_adv_event_received(self, data=None):
        '''
        Periodic Advertising event received. With or without periodic advertising data.
        '''
        self.periodic_adv_report_rxed = True

        if data:
            self.periodic_adv_report_data = data
            self.periodic_adv_report_data_rxed = True
```

if we do
```python
    def periodic_adv_event_received(self, data=None):
        '''
        Periodic Advertising event received. With or without periodic advertising data.
        '''
        if data:
            self.periodic_adv_report_data = data

        self.periodic_adv_report_rxed = True
```

Doesn't the circumvent the race condition (not really since there are no thread safety at all here) without the additional boolean?

m-alperen-sener: I see your point, in that case, as you have first suggested, tacking only data variable is enough to wait in the event. Yeap you re right. In any case it seems there will be one event with data; it is OK to capture it and look for it.

Thanks

autopts/pybtp/btp/gap.py
```diff
@@ -362,7 +363,20 @@ def gap_padv_sync_lost_ev_(gap, data, data_len):
...
+    fmt = '<HBBBBB'
+    if len(data) < struct.calcsize(fmt):
+        raise BTPError("Invalid data length")
+
+    # Extract the data length from the header
+    _, _, _, _, _, length = struct.unpack_from(fmt, data[:struct.calcsize(fmt)])
+
+    padv_data = data[7:]
+
+    if length > 0 and len(padv_data) == length:
```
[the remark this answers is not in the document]
m-alperen-sener: it is a valid case if there is no padv_data, then periodic_adv_event_received is called without data.

autopts/ptsprojects/stack/layers/gap.py
```diff
@@ -220,12 +224,35 @@ def wait_for_peer_car_status(self, timeout):
...
-        if self.periodic_report_rxed:
+    def periodic_adv_event_received(self, data=None):
+        '''
+        Periodic Advertising event received.
+        With or without periodic advertising data.
+        '''
+        if data is not None:
+            self.periodic_adv_report_data = data
+
+        self.periodic_adv_report_rxed = True
```
Thalley, line 238: Consider adding a comment about the order here and why it is necessary to prevent the race condition you've mentioned

autopts/ptsprojects/stack/layers/gap.py
```diff
@@ -220,12 +224,35 @@ def wait_for_peer_car_status(self, timeout):
...
+
+    def wait_periodic_report(self, timeout, data=None):
+        '''
+        Wait for periodic advertising report. If data is provided, waits for a report that
+        includes periodic advertising data that matches the provided one.
+
+        Data might arrive before calling this function, then it returns immediately.
+        '''
```
Thalley, line 246: `data` here is an entire report, i.e. one or more LTV structures?

autopts/pybtp/btp/gap.py
```diff
@@ -362,7 +363,21 @@ def gap_padv_sync_lost_ev_(gap, data, data_len):
 def gap_padv_report_ev_(gap, data, data_len):
     logging.debug("%s", gap_padv_report_ev_.__name__)
     stack = get_stack()
-    stack.gap.periodic_report_rxed = True
+
+    fmt = '<HBBBBB'
+    header_len = struct.calcsize(fmt)
+    if len(data) < header_len:
+        raise BTPError("Invalid data length")
```
Thalley: Suggest to log the actual length as well

autopts/pybtp/btp/gap.py
```diff
@@ -362,7 +363,21 @@ def gap_padv_sync_lost_ev_(gap, data, data_len):
...
+
+    # Extract the data length from the header
+    _, _, _, _, _, length = struct.unpack_from(fmt, data[:header_len])
+
+    padv_data = data[header_len:]
```
Thalley, line 386: If `len(data) == header_len` (i.e. no data), does `padv_data` just become an empty list here?

autopts/ptsprojects/stack/layers/gap.py
m-alperen-sener: Hmmm I need to look into LTV structures. If it is in the GAP layer PADV test scope.

autopts/pybtp/btp/gap.py
```diff
@@ -373,7 +363,21 @@ def gap_padv_sync_lost_ev_(gap, data, data_len):
...
+
+    if length > 0 and len(padv_data) == length:
```
m-alperen-sener, line 388: > If len(data) == header_len (i.e. no data), does padv_data just become an empty list here?

non zero length is guaranteed here, isn't it?

Suggestion was quick path which I did delete after putting `length > 0 and len(padv_data) == length` 
because does not `length > 0`  mean `len(data) != header_len` ?

Let me know if I am missing a point.

autopts/pybtp/btp/gap.py
```diff
@@ -362,7 +363,21 @@ def gap_padv_sync_lost_ev_(gap, data, data_len):
...
+        stack.gap.periodic_adv_event_received(padv_data)
+    else:
+        stack.gap.periodic_adv_event_received()
```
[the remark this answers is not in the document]
m-alperen-sener: nope we need to avoid race condition which could overwtire data previously received.

autopts/pybtp/btp/gap.py
Thalley: If PA report can contain 0 bytes of data IIRC, so `header_len == len(data)` may be true, which would make `length == 0`, won't it?

autopts/ptsprojects/stack/layers/gap.py
m-alperen-sener: In test spec it is not clearly stated but looking at the GAP layer spec it only covers the event itself not the payload data type or its meaning. The test only verifies the payload as it is, not what type of data it is.

PTS actually sends a complete LTV data and we compare against it as whole. Currently tester is reporting the HCI event data as it is not the payload, so the length and the data seems checked correctly.

For other cases: where data might be truncated or fragmented; it is not scope of this test apparently and there is no explicit ICS for such corner cases in GAP; since GAP will eventually delegate those info to upper layer to handle.

Currently PTS is set to use payload that can fit into one HCI buffer which can be supported by most of the tester implementations, otherwise it should be increased accordingly in tester so they can get the data in one complete event IMO.

autopts/pybtp/btp/gap.py
m-alperen-sener: yes but isn't it protected by `length > 0` check?

autopts/ptsprojects/stack/layers/gap.py
Thalley: Doesn't the core spec mandate that all advertising data is LTV structures?

m-alperen-sener: 7.7.65.15 LE Periodic Advertising Report event
Defined as variable length data
https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-62/out/en/host-controller-interface/host-controller-interface-functional-specification.html#UUID-e216c26f-0383-c651-4b8d-1409b91c7e34

And in Part C I don't see any specification about the payload integrity.

autopts/pybtp/btp/gap.py
Thalley: Yes, when `padv_data = data[header_len:]` and `len(padv_data) == 0` then `length > 0` is also true (or should be, unless the event is corrupted. Having both conditions doesn't hurt, but ideally shouldn't be necessary as far as I can tell

autopts/ptsprojects/stack/layers/gap.py
m-alperen-sener: So coming back to original question; yes it is the whole data can be one or more LTVs but GAP does not check this, only delivers reassembled payload with data status.

autopts/pybtp/btp/gap.py
m-alperen-sener: when `len(padv_data) == 0`  then `length = 0` I can't see how length can be greater than 0, It is reported from controller. length is Data_length parameter in event  0x3E. I don't expect controller to not catch such corruption but this check is against the BTP transfer from tester to AutoPTS. But since empty list is possible when length=0, as you mentioned, blocked by length > 0 check.

Thalley (approved):
I don't like the changes in the iut_config file, but not sure what the best alternative is either

autopts/pybtp/btp/gap.py
Thalley: > when `len(padv_data) == 0` then `length = 0` I can't see how length can be greater than 0, It is reported from controller. length is Data_length parameter in event 0x3E. I don't expect controller to not catch such corruption but this check is against the BTP transfer from tester to AutoPTS. But since empty list is possible when length=0, as you mentioned, blocked by length > 0 check.

Isn't the simple answer "A bug in the host and/or the BTP handlers"? :D

m-alperen-sener: I don't know, I felt like I am missing a point. Maybe I am still don't know.

sjanc (approved):
looks good, please rebase (and sorry for delay)

m-alperen-sener:
rebased on master

sjanc (approved):

Issue #712, opened 2022-03-16
Add missing verification mechanisms to MMI handlers in all profiles


michal-narajowski:
Example of a handler with missing verification: https://github.com/intel/auto-pts/pull/700

PR #1724, opened 2026-03-13
l2cap: add mechanism to retrieve error code

piotrnarajowski:
As stated in L2CAP BTP docs, Opcode 0x82 - Disconected Event:

> result value is returned in the response only if remote rejected connection request

`wait_for_conn_req_reject_ev` is introduced if we specifically expect an L2CAP connection request to fail and need to obtain the status value from response. Although `class L2capChan` already has a field that is related to that (`self.disconn_reason`), if connection request has been rejected an instance of L2capChan was never created thus disconn_reason was empty.

This is needed for verification in testcases L2CAP/LE/CFC/BV-10-C, L2CAP/LE/CFC/BV-12-C, L2CAP/LE/CFC/BV-14-C, L2CAP/LE/CFC/BV-24-C

Related issue: https://github.com/auto-pts/auto-pts/issues/712

### DIFF
--- a/autopts/ptsprojects/mynewt/l2cap.py
+++ b/autopts/ptsprojects/mynewt/l2cap.py
@@ -17,7 +17,7 @@
 
 from autopts.client import get_unique_name
 from autopts.ptsprojects.mynewt.ztestcase import ZTestCase
-from autopts.ptsprojects.stack import L2cap, get_stack
+from autopts.ptsprojects.stack import L2CapChannelStatusCode, get_stack
 from autopts.ptsprojects.testcase import TestFunc
 from autopts.pybtp import btp
 from autopts.pybtp.types import Addr
@@ -136,7 +136,7 @@ def test_cases(ptses):
 
     pre_conditions_unacceptable_parameters = common + [
         TestFunc(stack.l2cap_init, le_psm, le_initial_mtu),
-        TestFunc(btp.l2cap_le_listen, le_psm, le_initial_mtu, L2cap.unacceptable_parameters)
+        TestFunc(btp.l2cap_le_listen, le_psm, le_initial_mtu, L2CapChannelStatusCode.UNACCEPTABLE_PARAMETERS)
     ]
 
     pre_conditions_gatt = common + [
--- a/autopts/ptsprojects/stack/layers/l2cap.py
+++ b/autopts/ptsprojects/stack/layers/l2cap.py
@@ -110,6 +110,7 @@ class L2cap:
         self.channels = []
         self.hold_credits = 0
         self.num_channels = 2
+        self.conn_req_reject_reason = None
 
     def chan_lookup_id(self, chan_id):
         """ lookup L2CAP channel with specified channel ID"""
@@ -210,6 +211,22 @@ class L2cap:
             return True
 
         return wait_for_event(timeout, lambda: not self.is_connected(chan_id) and not self.is_connecting(chan_id))
+
+    def wait_for_conn_req_reject_ev(self, timeout: int = 30):
+        """
+        Waits for a L2CAP connection request to be rejected.
+
+        Args:
+            timeout (int): The maximum time to wait for the event in seconds.
+
+        Returns:
+            bool: True if the event occurred within the timeout, False otherwise.
+        """
+
+        if self.conn_req_reject_reason:
+            return True
+
+        return wait_for_event(timeout, lambda: self.conn_req_reject_reason is not None)
 
     def wait_for_connection(self, chan_id, timeout=5):
         """ Wait for channel connection """
--- a/autopts/ptsprojects/stack/layers/l2cap.py
+++ b/autopts/ptsprojects/stack/layers/l2cap.py
@@ -104,17 +104,26 @@ class L2capChan:
 
 
 class L2cap:
-    """L2CAP layer - manages L2CAP channels """
+    """
+    L2CAP layer - manages L2CAP channels.
 
-    def __init__(self, psm, initial_mtu):
-        # PSM used for testing for Client role
+    Attributes:
+        psm (int): Protocol/Service Multiplexer used for testing client role.
+        initial_mtu (int): Initial Maximum Transmission Unit.
+        channels (list[L2capChan]): List of currently active L2CAP channels.
+        hold_credits (int): Number of flow control credits currently held.
+        num_channels (int): Default or maximum number of channels.
+        conn_req_reject_reason (L2CapChannelStatusCode | None): The reason a
+            connection request was rejected, or None if no rejection occurred.
+    """
+
+    def __init__(self, psm: int, initial_mtu: int):
         self.psm = psm
         self.initial_mtu = initial_mtu
-        self.channels = []
+        self.channels: list[L2capChan] = []
         self.hold_credits = 0
         self.num_channels = 2
         self.conn_req_reject_reason: L2CapChannelStatusCode | None = None
-        """L2CapChannelStatusCode | None: The reason the L2CAP connection request was rejected."""
 
     def chan_lookup_id(self, chan_id):
         """ lookup L2CAP channel with specified channel ID"""
--- a/autopts/ptsprojects/stack/layers/l2cap.py
+++ b/autopts/ptsprojects/stack/layers/l2cap.py
@@ -14,7 +14,7 @@
 # more details.
 #
 import logging
-from enum import Enum, auto
+from enum import Enum, IntEnum, auto
 
 from autopts.ptsprojects.stack.common import wait_for_event
 
@@ -24,6 +24,20 @@ class L2capChanState(Enum):
     CONNECTING = auto()
     CONNECTED = auto()
     DISCONNECTED = auto()
+
+
+class L2CapChannelStatusCode(IntEnum):
+    SUCCESS = 0x0000
+    LE_PSM_NOT_SUPPORTED = 0x0002
+    INSUFFICIENT_RESOURCES = 0x0004
+    INSUFFICIENT_AUTHENTICATION = 0x0005
+    INSUFFICIENT_AUTHORIZATION = 0x0006
+    INSUFFICIENT_ENCRYPTION_KEY_SIZE = 0x0007
+    INSUFFICIENT_ENCRYPTION = 0x0008
+    INVALID_SOURCE_CID = 0x0009
+    SOURCE_CID_ALREADY_ALLOCATED = 0x000a
+    UNACCEPTABLE_PARAMETERS = 0x000b
+    INVALID_PARAMETERS = 0x000c
 
 
 class L2capChan:
@@ -91,17 +105,6 @@ class L2capChan:
 
 class L2cap:
     """L2CAP layer - manages L2CAP channels """
-    connection_success = 0x0000
-    unknown_le_psm = 0x0002
-    no_resources = 0x0004
-    insufficient_authen = 0x0005
-    insufficient_author = 0x0006
-    insufficient_key_sz = 0x0007
-    insufficient_enc = 0x0008
-    invalid_source_cid = 0x0009
-    source_cid_already_used = 0x000a
-    unacceptable_parameters = 0x000b
-    invalid_parameters = 0x000c
 
     def __init__(self, psm, initial_mtu):
         # PSM used for testing for Client role
@@ -110,7 +113,8 @@ class L2cap:
         self.channels = []
         self.hold_credits = 0
         self.num_channels = 2
-        self.conn_req_reject_reason = None
+        self.conn_req_reject_reason: L2CapChannelStatusCode | None = None
+        """L2CapChannelStatusCode | None: The reason the L2CAP connection request was rejected."""
 
     def chan_lookup_id(self, chan_id):
         """ lookup L2CAP channel with specified channel ID"""
--- a/autopts/pybtp/btp/l2cap.py
+++ b/autopts/pybtp/btp/l2cap.py
@@ -19,7 +19,7 @@ import binascii
 import logging
 import struct
 
-from autopts.ptsprojects.stack import get_stack
+from autopts.ptsprojects.stack import L2CapChannelStatusCode, get_stack
 from autopts.pybtp import defs
 from autopts.pybtp.btp.btp import CONTROLLER_INDEX, btp_hdr_check, pts_addr_get, pts_addr_type_get
 from autopts.pybtp.btp.btp import get_iut_method as get_iut
@@ -400,11 +400,14 @@ def l2cap_disconnected_ev(l2cap, data, data_len):
 
     hdr_fmt = '<HBHB6s'
     res, chan_id, psm, bd_addr_type, bd_addr = struct.unpack_from(hdr_fmt, data)
-    result_str = l2cap_result_str[res]
-    l2cap.disconnected(chan_id, psm, bd_addr_type, bd_addr, result_str)
+    status = L2CapChannelStatusCode(res)
+    l2cap.disconnected(chan_id, psm, bd_addr_type, bd_addr, status)
+    if status != L2CapChannelStatusCode.SUCCESS:
+        # L2CAP connection request failed
+        l2cap.conn_req_reject_reason = status
 
     logging.debug("id:%r on psm:%r, addr %r type %r, res %r",
-                  chan_id, psm, bd_addr, bd_addr_type, result_str)
+                  chan_id, psm, bd_addr, bd_addr_type, status.name)
 
 
 def l2cap_data_rcv_ev(l2cap, data, data_len):
--- a/autopts/pybtp/btp/l2cap.py
+++ b/autopts/pybtp/btp/l2cap.py
@@ -154,18 +154,6 @@ def l2cap_conn_v2(bd_addr, bd_addr_type, psm, mtu=0, num=1, ecfc=0, hold_credit=
     return chan_ids
 
 
-l2cap_result_str = {0: "Success",
-                    2: "LE_PSM not supported",
-                    4: "Insufficient Resources",
-                    5: "insufficient authentication",
-                    6: "insufficient authorization",
-                    7: "insufficient encryption key size",
-                    8: "insufficient encryption",
-                    9: "Invalid Source CID",
-                    10: "Source CID already allocated",
-                    }
-
-
 def l2cap_conn_rsp():
     logging.debug("")
 
--- a/autopts/wid/l2cap.py
+++ b/autopts/wid/l2cap.py
@@ -17,7 +17,7 @@ import logging
 import re
 import time
 
-from autopts.ptsprojects.stack import get_stack
+from autopts.ptsprojects.stack import L2CapChannelStatusCode, get_stack
 from autopts.pybtp import btp, defs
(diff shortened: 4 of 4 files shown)


Thalley:
Not really sure if I'm the best at reviewing this - I'm not really involved with L2CAP :)

piotrnarajowski:
Applied copilot suggestions and implemented `pop_conn_req_reject_reason()` method that gets and clears `stack.l2cap.conn_req_reject_reason` flag

autopts/wid/l2cap.py
```diff
@@ -502,23 +499,51 @@ def hdl_wid_138(_: WIDParams):
...
 def hdl_wid_251(_: WIDParams):
-    # TODO: Fix to actually verify result of 'Insufficient Encryption' 0x0008 error
-    return get_stack().l2cap.wait_for_disconnection(0, 30)
+    """
+        Did Implementation Under Test(IUT) receive Connection refused 'Insufficient Encryption' 0x0008 error?
+    """
+    stack = get_stack()
+    stack.l2cap.wait_for_conn_req_reject_ev(30)
+    reason = stack.l2cap.pop_conn_req_reject_reason()
+    if reason == "insufficient authorization":
```
szymon-czapracki: Are those reasons passed as strings via btp? Is there no way to check the error codes that are passed to the IUT?

piotrnarajowski: It is not passed as strings. Please see `l2cap_disconnected_ev()` where line `result_str = l2cap_result_str.get(res, "Unknown Error")` is responsible for this

autopts/wid/l2cap.py
```diff
@@ -499,23 +499,51 @@ def hdl_wid_138(_: WIDParams):
...
```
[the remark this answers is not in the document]
piotrnarajowski: Fixed

autopts/ptsprojects/stack/layers/l2cap.py
```diff
@@ -172,6 +173,34 @@ def wait_for_disconnection(self, chan_id, timeout):
...
+            timeout (int): The maximum time to wait for the event in seconds.
+
+        Returns:
+            bool: True if the event occurred within the timeout, False otherwise.
+        """
+
+        if self.conn_req_reject_reason:
+            return True
+
+        return wait_for_event(timeout, lambda: self.conn_req_reject_reason)
```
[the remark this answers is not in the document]
piotrnarajowski: fair point

autopts/pybtp/btp/l2cap.py
```diff
@@ -396,8 +396,11 @@ def l2cap_disconnected_ev(l2cap, data, data_len):
 
     hdr_fmt = '<HBHB6s'
     res, chan_id, psm, bd_addr_type, bd_addr = struct.unpack_from(hdr_fmt, data)
-    result_str = l2cap_result_str[res]
+    result_str = l2cap_result_str.get(res, "Unknown Error")
     l2cap.disconnected(chan_id, psm, bd_addr_type, bd_addr, result_str)
+    if result_str != "Success":
+        # L2CAP connection request failed
```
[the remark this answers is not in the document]
piotrnarajowski: l2cap.disconnected returns when chan is None

autopts/wid/l2cap.py
```diff
@@ -499,23 +499,51 @@ def hdl_wid_138(_: WIDParams):
...
-    return get_stack().l2cap.wait_for_disconnection(0, 30)
+    """
+        Did Implementation Under Test(IUT) receive Connection refused 'Insufficient Encryption' 0x0008 error?
+    """
+    stack = get_stack()
+    stack.l2cap.wait_for_conn_req_reject_ev(30)
+    reason = stack.l2cap.pop_conn_req_reject_reason()
+    if reason == "insufficient encryption":
+        return True
+    return False
```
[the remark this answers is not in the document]
piotrnarajowski: L2Cap object is initialized at every test case so there is no chance to leave a stale reject reason.

autopts/ptsprojects/stack/layers/l2cap.py
```diff
@@ -107,6 +107,7 @@ def __init__(self, psm, initial_mtu):
         self.channels = []
         self.hold_credits = 0
         self.num_channels = 2
+        self.conn_req_reject_reason = None
```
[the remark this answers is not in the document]
piotrnarajowski: It is not present in `Stack.cleanup()` because cleanup calls service_init methods. The L2CAP init method requires data like psm and initial_mtu, that's why it's usually called in preconditions section.

autopts/pybtp/btp/l2cap.py
```diff
@@ -396,8 +396,11 @@ def l2cap_disconnected_ev(l2cap, data, data_len):
 
     hdr_fmt = '<HBHB6s'
     res, chan_id, psm, bd_addr_type, bd_addr = struct.unpack_from(hdr_fmt, data)
-    result_str = l2cap_result_str[res]
+    result_str = l2cap_result_str.get(res, "Unknown Error")
     l2cap.disconnected(chan_id, psm, bd_addr_type, bd_addr, result_str)
+    if result_str != "Success":
```
Thalley: Let's not use literal strings for comparison. 

Either define the L2CAP results as a string enum and use that, or use the defined actually defined integer values for reasons (from the core spec)

autopts/pybtp/btp/l2cap.py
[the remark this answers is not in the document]
Thalley: if `chan = l2cap.chan_lookup_id(chan_id)` is None, wouldn't that indicate a serious issue?

autopts/ptsprojects/stack/layers/l2cap.py
```diff
@@ -107,6 +107,7 @@ def __init__(self, psm, initial_mtu):
...
```
Thalley: Please add documentation for the newly added field since it's clearly accessible via `stack.l2cap.conn_req_reject_reason` (i.e. it's a public field).

autopts/ptsprojects/stack/layers/l2cap.py
```diff
@@ -172,6 +173,34 @@ def wait_for_disconnection(self, chan_id, timeout):
 
         return wait_for_event(timeout, lambda: not self.is_connected(chan_id))
 
+    def wait_for_conn_req_reject_ev(self, timeout):
```
Thalley: Suggest to 
```suggestion
    def wait_for_conn_req_reject_ev(self, timeout: int = 30):
```

to apply type hinting and use an appropriate default value so you don't have to call `wait_for_conn_req_reject_ev(30)` everywhere

autopts/wid/l2cap.py
```diff
@@ -499,23 +499,51 @@ def hdl_wid_138(_: WIDParams):
...
-    # TODO: Fix to actually verify result of 'Insufficient Authentication' 0x0005 error
-    return get_stack().l2cap.wait_for_disconnection(0, 30)
+    """
+        Did Implementation Under Test(IUT) receive Connection refused 'Insufficient Authentication' 0x0005 error
+    """
+    stack = get_stack()
+    stack.l2cap.wait_for_conn_req_reject_ev(30)
+    reason = stack.l2cap.pop_conn_req_reject_reason()
+    if reason == "insufficient authentication":
+        return True
```
Thalley: What's the reasoning behind `pop_conn_req_reject_reason`? 

I would expect the `reason` to only ever be populated once; what we do get from clearing / popping it, instead of just reading it?

Thalley (changes requested):

autopts/wid/l2cap.py
piotrnarajowski: It will be populated only once,  it looks like I put too much trust in the Copilot suggestion that was triggered here

autopts/pybtp/btp/l2cap.py
```diff
@@ -396,8 +396,11 @@ def l2cap_disconnected_ev(l2cap, data, data_len):
...
+        # L2CAP connection request failed
+        l2cap.conn_req_reject_reason = result_str
```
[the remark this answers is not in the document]
piotrnarajowski: Again, l2cap object is reinitialized at the start of every test case

autopts/pybtp/btp/l2cap.py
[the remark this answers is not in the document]
piotrnarajowski: `class L2Cap self.channels` list is populated with `L2capChan` objects. The `L2capChan` object is initialized only when connection has been succesful. So it doesn't indicate a serious issue.

autopts/wid/l2cap.py
Thalley: But do we gain anything from popping it so that the reason isn't available later? 

Keep in mind that something like a `reject_reason` should not be used to determine the state of the L2CAP channel IMO

piotrnarajowski: > But do we gain anything from popping it so that the reason isn't available later?

I think no. Again, I blindly accepted the copiloto suggestions triggered

> Keep in mind that something like a reject_reason should not be used to determine the state of the L2CAP channel IMO

What do you mean by that? In these modified wid handlers we should expect connection to fail.

autopts/pybtp/btp/l2cap.py
piotrnarajowski: This is gonna be trickier than I thought, currently there are two places where these L2Cap Channel status codes are defined, I'm not sure why though:

in `layers/l2cap.py`, only one of which is used
```
class L2cap:
    """L2CAP layer - manages L2CAP channels """
    connection_success = 0x0000
    unknown_le_psm = 0x0002
    no_resources = 0x0004
    insufficient_authen = 0x0005
    insufficient_author = 0x0006
    insufficient_key_sz = 0x0007
    insufficient_enc = 0x0008
    invalid_source_cid = 0x0009
    source_cid_already_used = 0x000a
    unacceptable_parameters = 0x000b
    invalid_parameters = 0x000c
```
in `types.py`

```
class L2CAPConnectionResponse:
    success = 0
    insufficient_authentication = 1
    insufficient_authorization = 2
    insufficient_encryption_key_size = 3
    insufficient_encryption = 4
    insufficient_secure_authentication = 5
```

autopts/wid/l2cap.py
Thalley: > What do you mean by that? In these modified wid handlers we should expect connection to fail.

It was a followup as to why I don't think it makes sense to pop it; popping it seems to treat it more like an event, rather than just a value :)

autopts/pybtp/btp/l2cap.py
piotrnarajowski: To clarify, the upper one contain actual status codes, the lower are just btp command parameters

autopts/wid/l2cap.py
piotrnarajowski: I ditched the pop method and introduced L2CAP results enum

autopts/pybtp/btp/l2cap.py
Thalley: Hmm, yeah this is a bit odd. 

Setting aside that neither of these are Enums or use appropriate naming for constants in Python, the top one seems to be the the BT Core Spec defined values, which we also have in Zephyr

```
#define BT_L2CAP_LE_SUCCESS             0x0000
#define BT_L2CAP_LE_ERR_PSM_NOT_SUPP    0x0002
#define BT_L2CAP_LE_ERR_NO_RESOURCES    0x0004
#define BT_L2CAP_LE_ERR_AUTHENTICATION  0x0005
#define BT_L2CAP_LE_ERR_AUTHORIZATION   0x0006
#define BT_L2CAP_LE_ERR_KEY_SIZE        0x0007
#define BT_L2CAP_LE_ERR_ENCRYPTION      0x0008
#define BT_L2CAP_LE_ERR_INVALID_SCID    0x0009
#define BT_L2CAP_LE_ERR_SCID_IN_USE     0x000A
#define BT_L2CAP_LE_ERR_UNACCEPT_PARAMS 0x000B
#define BT_L2CAP_LE_ERR_INVALID_PARAMS  0x000C
```

The bottom one seems to a different set of responses; are those defined specific for BTP, and in that case, is the `reason` value that we receive here the BTP ones or the Core Spec ones?

autopts/ptsprojects/stack/layers/l2cap.py
piotrnarajowski: Done

autopts/pybtp/btp/l2cap.py
piotrnarajowski: The bottom one are related to L2Cap Listen BTP Command: 
```
                Response parameter can be used to specify a response value used in
                L2CAP Connection Response to send non-success responses like
                Insufficient Authentication. If device bases its security requirements
                for creating L2CAP connection on PSM, this value should be 0x00.

                Valid Response values:
                        0x00 = Success
                        0x01 = Insufficient authentication
                        0x02 = Insufficient authorization
                        0x03 = Insufficient encryption key size
                        0x04 = Insufficient encryption
                        0x05 = Insufficient secure authentication
```

If I understand correctly, this can be used for test cases where IUT is expected to respond with specific error to lower testers connection request.

The case that my PR covers is opposite, i.e. enables to retrieve a response IUT has recevied after sending L2cap connection request

Thalley: > The case that my PR covers is opposite, i.e. enables to retrieve a response IUT has recevied after sending L2cap connection request

Right, but are the values that we receive here the ones defined by BTP or by the Core spec?

piotrnarajowski: The values we receive are the actual values the IUT receives in the L2CAP LE Connection response, as defined by the Core spec.

My changes introduce the `L2CapChannelStatusCodes(Enum)` class to represent these values in auto-pts.

Thalley: > The values we receive are the actual values the IUT receives in the L2CAP LE Connection response, as defined by the Core spec.

That begs the question on whether we should instead use the BTP defined ones, or modify the BTP defines ones
```
                Valid Response values:
                        0x00 = Success
                        0x01 = Insufficient authentication
                        0x02 = Insufficient authorization
                        0x03 = Insufficient encryption key size
                        0x04 = Insufficient encryption
                        0x05 = Insufficient secure authentication
```

to use the Core Spec defined ones? It's a mess to define the same semantic values (using different actual values) in two different ways, but that can be postponed to a different PR

autopts/ptsprojects/stack/layers/l2cap.py
```diff
@@ -26,6 +26,20 @@ class L2capChanState(Enum):
     DISCONNECTED = auto()
 
 
+class L2CapChannelStatusCodes(Enum):
```
Thalley:
```suggestion
class L2CapChannelStatusCodes(IntEnum):
```

autopts/ptsprojects/stack/layers/l2cap.py
```diff
@@ -110,6 +113,8 @@ def __init__(self, psm, initial_mtu):
         self.channels = []
         self.hold_credits = 0
         self.num_channels = 2
+        self.conn_req_reject_reason = None
+        """L2CapChannelStatusCodes: Represents a reason when an L2CAP connection request fails."""
```
Thalley: Not the correct way of documenting a field in a class :)

autopts/ptsprojects/stack/layers/l2cap.py
Thalley: What is the type of `conn_req_reject_reason`? `L2CapChannelStatusCodes`?

autopts/ptsprojects/stack/layers/l2cap.py
Thalley:
```suggestion
class L2CapChannelStatusCode(Enum):
```

Type names are usually not in plural

Thalley (changes requested):

autopts/ptsprojects/stack/layers/l2cap.py
piotrnarajowski: Yes, it's now stated in updated documentation

autopts/ptsprojects/stack/layers/l2cap.py
```diff
@@ -26,6 +26,20 @@ class L2capChanState(Enum):
...
+    INSUFFICIENT_RESOURCES = 0x0004
+    INSUFFICIENT_AUTHENTICATION = 0x0005
+    INSUFFICIENT_AUTHORIZATION = 0x0006
+    INSUFFICIENT_ENCRYPTION_KEY_SIZE = 0x0007
+    INSUFFICIENT_ENCRYPTION = 0x0008
+    INVALID_SOURCE_CID = 0x0009
+    SOURCE_CID_ALREADY_ALLOCATED = 0x000a
+    UNACCEPTABLE_PARAMETERS = 0x000b
+    INVALID_PARAMETERS = 0x000c
+
```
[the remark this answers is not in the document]
piotrnarajowski: Changed to IntEnum

autopts/ptsprojects/mynewt/l2cap.py
```diff
@@ -136,7 +136,7 @@ def test_cases(ptses):
 
     pre_conditions_unacceptable_parameters = common + [
         TestFunc(stack.l2cap_init, le_psm, le_initial_mtu),
-        TestFunc(btp.l2cap_le_listen, le_psm, le_initial_mtu, L2cap.unacceptable_parameters)
+        TestFunc(btp.l2cap_le_listen, le_psm, le_initial_mtu, L2CapChannelStatusCodes.UNACCEPTABLE_PARAMETERS)
```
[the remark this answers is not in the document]
piotrnarajowski: Changed to IntEnum

autopts/ptsprojects/stack/layers/l2cap.py
piotrnarajowski: Done

autopts/ptsprojects/stack/layers/l2cap.py
piotrnarajowski: Should be OK now

autopts/ptsprojects/stack/layers/l2cap.py
piotrnarajowski: Updated

autopts/ptsprojects/stack/layers/l2cap.py
```diff
@@ -110,6 +113,8 @@ def __init__(self, psm, initial_mtu):
         self.channels = []
         self.hold_credits = 0
         self.num_channels = 2
+        self.conn_req_reject_reason: L2CapChannelStatusCode | None = None
+        """L2CapChannelStatusCode | None: The reason the L2CAP connection request was rejected."""
```
Thalley: Please see https://github.com/auto-pts/auto-pts/blob/master/autopts/wid/bap.py#L78-L91 or other examples on how to properly document a field/attribute in a Python class

Thalley (approved):
Minor comment otherwise LGTM

autopts/ptsprojects/stack/layers/l2cap.py
piotrnarajowski: Added proper documentation for `class L2cap`

PR #1726, opened 2026-03-17
Phase1/gatt migration to gatt client api

kedMertens:
This PR implements the first step in migrating Zephyr GATT test automation away from the deprecated GATT BTP service and toward the GATT Client (`GATT_CL`) API.

Details are captured in the [Discord discussion](https://discord.com/channels/720317445772017664/1478668447775854755), and the related Zephyr-side context is tracked in [74485](https://github.com/zephyrproject-rtos/zephyr/issues/74485#issuecomment-3996027695).

The current implementation uses runtime service detection to decide whether the migrated path can be used. If the IUT reports `GATT_CL` support, the new API is used; otherwise the legacy GATT path is preserved.


autopts/ptsprojects/zephyr/gatt.py
```diff
@@ -376,6 +384,8 @@ def test_cases_client(pts):
     iut_device_name = get_unique_name(pts)
     stack = get_stack()
 
+    use_gattc = os.getenv("AUTOPTS_ZEPHYR_USE_GATTC") == "1"
```
sjanc: I'd check if GATT CL service is supported by  IUT and if it is use it instead of legacy GATT

kedMertens: will do

autopts/bot/iut_config/zephyr.py
```diff
@@ -128,6 +128,26 @@
         ]
     },
 
+    "gatt_cl_gar_bv10.conf": {
```
sjanc: why do we need those overlays?  if this is to enforce exactly 2 bearers being use it should probably be LT that enforces this

kedMertens: For GATT/CL/GAR/BV-10-C CONFIG_BT_EATT_MAX should be set to 1, the test checks that when both EATT and Unenhanced ATT supported, if one read is pending on one bearer next read will use another one.

autopts/ptsprojects/zephyr/gatt.py
```diff
@@ -27,6 +29,16 @@
 from autopts.wid.gatt import gattc_wid_hdl_multiple_indications
 
 
+def _set_optional_pixit(pts, project, name, value):
+    try:
+        pts.set_pixit(project, name, value)
+    except Fault as exc:
+        if "PTSCONTROL_E_PIXIT_PARAM_NOT_FOUND" in str(exc):
```
sjanc: I'm not fully sure if we should do this,  PTS tends to change/remove IXITs names and in such case we should properly check what needs to be updated instead of just skipping removed IXIT

also in long term we will reduce those to only non-default value (see https://github.com/auto-pts/auto-pts/issues/1716)

autopts/pybtp/defs.py
```diff
@@ -265,7 +265,8 @@ def BIT(bit):
 BTP_GATTC_CMD_WRITE_RELIABLE = 0x11
 BTP_GATTC_CMD_CFG_NOTIFY = 0x12
 BTP_GATTC_CMD_CFG_INDICATE = 0x13
-BTP_GATTC_CMD_READ_MULTIPLE_VAR = 0x14
+BTP_GATTC_CMD_EATT_CONNECT = 0x1f
```
sjanc, line 270: those commands are from GATT service

autopts/bot/iut_config/zephyr.py
sjanc: ok, I guess this is fine for now, however I think in long term we should have a way to force specific transport from BTP level,  unfortunately those commands don't have 'flags' so we would have to either add v2 variants or maybe have a top level command like 'GATT client bearers" or maybe "GATT client options" that  would allow to tune client behavior for tests

@Thalley WDYT?

Thalley: We can control how many EATT bearers we connect via `BTP_GATT_CMD_EATT_CONNECT` (takes the number of bearers as a parameter`, so even when `CONFIG_BT_EATT_MAX=5` we can still have the test only use the ATT bearer and 1 EATT bearer to ensure that it switches between those two. 

I don't see why we need to reconfigure the maximum number

kedMertens: I missed that one, will test.

autopts/pybtp/defs.py
kedMertens: It seems that there is no eatt_connect command counterpart in gatt client api, [doc](https://github.com/auto-pts/auto-pts/blob/master/doc/btp_gatt_client.txt). Should it be added? 
I've checked nimble implementation it doesn't have it in use.

autopts/ptsprojects/zephyr/gatt_wid.py
```diff
@@ -15,11 +15,16 @@
...
 
 log = logging.debug
 
 
 def gatt_wid_hdl(wid, description, test_case_name):
-    log(f'{gatt_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
-    return generic_wid_hdl(wid, description, test_case_name, [__name__, 'autopts.wid.gatt'])
+    use_gattc = get_stack().is_svc_supported("GATT_CL")
+    if test_case_name.startswith("GATT/CL/GAR/") and use_gattc:
+        return generic_wid_hdl(wid, description, test_case_name, [__name__, "autopts.wid.gatt_client", "autopts.wid.gatt"])
```
Thalley: What if `test_case_name.startswith("GATT/CL/GAR/")` and `use_gattc` is False?

autopts/ptsprojects/zephyr/gatt.py
```diff
@@ -381,7 +386,8 @@ def test_cases_client(pts):
             stack.gap.iut_addr_get_str())),
         TestFunc(btp.core_reg_svc_gatt),
         TestFunc(btp.set_pts_addr, pts_bd_addr, Addr.le_public),
-        TestFunc(stack.gatt_init)
+        TestFunc(stack.gatt_init),
+        TestFunc(init_gatt_client_if_supported)
```
Thalley: When is GATT client not supported?

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -39,6 +39,11 @@ def gatt_cl_mtu_exchanged_ev_(gatt_cl, data, data_len):
     logging.debug("%s %r", gatt_cl_mtu_exchanged_ev_.__name__, data)
 
     fmt = '<B6sB'
+    read_mult_var_fmt = '<B6sBBH'
+
+    if data_len >= struct.calcsize(read_mult_var_fmt):
+        gatt_cl_read_mult_var_rsp_ev_(gatt_cl, data, data_len)
+        return
```
Thalley: Why are we doing this in a MTU exchanged event?

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -429,8 +434,8 @@ def gatt_cl_read_uuid_rsp_ev_(gatt_cl, data, data_len):
                   gatt_cl_read_uuid_rsp_ev_.__name__,
                   addr_type, addr, status, data_length, value_length)
 
-    if status != 0:
-        add_to_verify_values(att_rsp_str[status])
+    if status != 0 and data_length == 0:
+        add_to_verify_values(att_rsp_str.get(status, f"ATT error 0x{status:02x}"))
```
Thalley: Not sure how `add_to_verify_values` is used, but don't we want to add information about empty reads as well?

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -534,33 +539,78 @@ def gatt_cl_read_mult_rsp_ev_(gatt_cl, data, data_len):
 def gatt_cl_read_mult_var_rsp_ev_(gatt_cl, data, data_len):
     logging.debug("%s %r", gatt_cl_read_mult_var_rsp_ev_.__name__, data)
 
-    fmt = '<B6sBH'
+    # Legacy format: <addr_type, addr, att_status, data_length>
+    legacy_fmt = '<B6sBH'
+    # Migrated Zephyr format: <addr_type, addr, btp_status, att_status, data_length>
+    migrated_fmt = '<B6sBBH'
```
Thalley, line 579: This is not a Zephyr specific file, so we should not have Zephyr specific behavior here, right?

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -534,33 +539,78 @@ def gatt_cl_read_mult_rsp_ev_(gatt_cl, data, data_len):
...
+        )
+        rp_data = data[migrated_size:]
+    elif data_len >= legacy_size:
+        addr_type, addr, att_status, data_length = struct.unpack_from(
+            legacy_fmt, data[:legacy_size]
+        )
+        status = 0
+        rp_data = data[legacy_size:]
+    else:
+        add_to_verify_values("Malformed READ_MULTIPLE_VAR response")
```
Thalley: Suggest to modify `att_rsp_str` to an enum and add this new condition to that and then use the enum instead of hardcoding a string

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -534,33 +539,78 @@ def gatt_cl_read_mult_rsp_ev_(gatt_cl, data, data_len):
...
-    logging.debug("%s received addr_type=%r addr=%r status=%r data_len=%r",
+    logging.debug(
+        "%s received addr_type=%r addr=%r status=%r att_status=%r data_len=%r",
                   gatt_cl_read_mult_var_rsp_ev_.__name__,
-                  addr_type, addr, status, data_length)
+                  addr_type, addr, status, att_status, data_length)
 
-    rp_data = data[struct.calcsize(fmt):]
+    if status != 0:
+        add_to_verify_values(f"BTP status 0x{status:02x}")
```
Thalley: Ditto; we should not supply hardcoded strings to `add_to_verify_values`, should we?

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -1005,6 +1055,54 @@ def gatt_cl_read_multiple_var(bd_addr_type, bd_addr, *hdls):
     gatt_cl_command_rsp_succ()
 
 
+def gatt_cl_eatt_connect(bd_addr, bd_addr_type, num=1):
+    logging.debug("%s %r %r %r", gatt_cl_eatt_connect.__name__, bd_addr, bd_addr_type, num)
+    iutctl = get_iut()
```
Thalley: Please add proper documentation for new (or changed) functions

autopts/wid/gatt_client.py
```diff
@@ -50,26 +50,33 @@ def hdl_wid_1(_: WIDParams):
...
     initiate GATT disconnect request to PTS.
     """
-    btp.gap_disconn(btp.pts_addr_get(), btp.pts_addr_type_get())
-    return get_stack().gap.wait_for_disconnection(30)
+    if "LT2" in params.test_case_name:
+        btp.gap_disconn(btp.lt2_addr_get(), btp.lt2_addr_type_get())
+        return get_stack().gap.wait_for_disconnection(30, btp.lt2_addr_get())
+    else:
+        btp.gap_disconn(btp.pts_addr_get(), btp.pts_addr_type_get())
+        return get_stack().gap.wait_for_disconnection(30)
```
Thalley: I wonder if it makes sense to add/use a function that returns either `pts_addr_get` or `lt2_addr_get` based on `test_case_name` so we don't need this `if ... else` so many places

autopts/wid/gatt_client.py
```diff
@@ -1329,11 +1336,121 @@ def hdl_wid_147(params: WIDParams):
...
+            btp.pts_addr_get(),
+            hdl1,
+            hdl2,
+        )
+        btp.gatt_cl_read_multiple_var(
+            btp.pts_addr_type_get(),
+            btp.pts_addr_get(),
+            hdl1,
+            hdl2,
+        )
```
Thalley: How does this ensure that an EATT bearer is being used?

autopts/wid/gatt_client.py
```diff
@@ -1329,11 +1336,121 @@ def hdl_wid_147(params: WIDParams):
...
+        return False
+
+    if not stack.gatt_cl.wait_for_verify_values(timeout=35, expected_count=2):
+        logging.error("%s timeout waiting for two read_multiple_var async events", hdl_wid_147.__name__)
+        return False
+
+    verify_values = btp.get_verify_values()
+    if len(verify_values) < 2:
+        logging.error("%s missing read_multiple_var verify values: %r", hdl_wid_147.__name__, verify_values)
+        return False
```
Thalley: Should we also not verify that the values are succesful? Looks like `add_to_verify_values` may also be called in case of erros

autopts/wid/gatt_client.py
```diff
@@ -1329,11 +1336,121 @@ def hdl_wid_147(params: WIDParams):
...
+    hdl1 = MMI.args[0]
+    hdl2 = MMI.args[1]
+
+    if not hdl1 or not hdl2:
+        logging.error("parsing error")
+        return False
+
+    stack = get_stack()
+    for attempt in range(2):
+        btp.clear_verify_values()
```
Thalley: Why do we need 2 attempts?

autopts/wid/gatt_client.py
```diff
@@ -1329,11 +1336,121 @@ def hdl_wid_147(params: WIDParams):
...
+        if len(verify_values) < 2:
+            logging.error("%s missing read_multiple_var verify values: %r", hdl_wid_148.__name__, verify_values)
+            return False
+
+        statuses = [item[0] if isinstance(item, tuple) else item for item in verify_values[:2]]
+
+        if all(status == "" for status in statuses):
+            return True
+
+        if attempt == 0 and all(status == "unlikely error" for status in statuses):
```
Thalley: Please avoid using hardcoded strings for comparison

autopts/wid/gatt_client.py
```diff
@@ -1348,6 +1465,101 @@ def hdl_wid_150(params: WIDParams):
...
+    def _try_eatt_connect(num):
+        try:
+            btp.gatt_cl_eatt_connect(
+                btp.pts_addr_get(),
+                btp.pts_addr_type_get(),
+                num,
+            )
+            return None
+        except Exception as err:
+            return err
```
Thalley: Can we return exceptions in WID handlers?

autopts/wid/gatt_client.py
```diff
@@ -1348,6 +1465,101 @@ def hdl_wid_150(params: WIDParams):
...
+                        logging.debug(
+                            "%s EATT disconnect cleanup failed for %s: %r",
+                            hdl_wid_400.__name__,
+                            params.test_case_name,
+                            disconn_err,
+                        )
+                        break
+                    sleep(0.5)
+                    continue
+                sleep(0.5)
```
Thalley: Why do we need sleeps here?

autopts/wid/gatt_client.py
```diff
@@ -1348,6 +1465,101 @@ def hdl_wid_150(params: WIDParams):
...
+            if disconn_err is not None:
+                logging.debug(
+                    "%s EATT disconnect cleanup failed: %r",
+                    hdl_wid_400.__name__,
+                    disconn_err,
+                )
+                break
+            sleep(0.5)
+            continue
+        sleep(0.5)
```
Thalley: Ditto w.r.t. sleeps

Thalley (changes requested):

autopts/pybtp/btp/gatt_cl.py
kedMertens: yeah you mean that on empty read you will still get a meaningful result, will address.

autopts/pybtp/btp/gatt_cl.py
kedMertens: My bad, I'll reword

autopts/pybtp/btp/gatt_cl.py
kedMertens: will do

autopts/pybtp/btp/gatt_cl.py
kedMertens: will do

autopts/wid/gatt_client.py
kedMertens: will rework, I think would be cleaner

autopts/wid/gatt_client.py
kedMertens: I don't think there is a way to observe which bearer was in use. Do you have an idea in mind how to do that?

autopts/wid/gatt_client.py
kedMertens: Agreed. I added explicit status checks on the collected verify values so the handler now fails on error

autopts/wid/gatt_client.py
kedMertens: Sure

autopts/wid/gatt_client.py
kedMertens: was forcing to make it pass, and forgot to remove/rework

autopts/wid/gatt_client.py
```diff
@@ -1348,6 +1499,88 @@ def hdl_wid_150(params: WIDParams):
...
+    can accept L2CAP_CREDIT_BASED_CONNECTION_REQ from PTS.
+    """
+
+    def _try_eatt_connect(num):
+        try:
+            btp.gatt_cl_eatt_connect(
+                btp.pts_addr_get(),
+                btp.pts_addr_type_get(),
+                num,
+            )
```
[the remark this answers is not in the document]
kedMertens: will not address, wid docstring must match what PTS requests, can be found in GATT ATS: 
{400,%s,GATT}Please prepare IUT into an L2CAP Credit Based
Connection connectable mode using LE signaling
channel.\n\nDescription: Verify that the Implementation Under Test
(IUT) can accept L2CAP_CREDIT_BASED_CONNECTION_REQ
from PTS.

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -1005,6 +1068,55 @@ def gatt_cl_read_multiple_var(bd_addr_type, bd_addr, *hdls):
...
+                    status,
+                )
+                logging.debug("%s sending legacy GATT EATT_CONNECT", gatt_cl_eatt_connect.__name__)
+                iutctl.btp_socket.send(
+                    defs.BTP_SERVICE_ID_GATT, defs.BTP_GATT_CMD_EATT_CONNECT, CONTROLLER_INDEX, data=data_ba
+                )
+                tuple_hdr, _ = iutctl.btp_socket.read()
+                logging.debug("%s legacy GATT EATT_CONNECT response: %r", gatt_cl_eatt_connect.__name__, tuple_hdr)
+                btp_hdr_check(tuple_hdr, defs.BTP_SERVICE_ID_GATT)
+                return
```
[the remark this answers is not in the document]
kedMertens: About that one I'm not sure, will need feedback

autopts/ptsprojects/zephyr/gatt.py
```diff
@@ -372,6 +372,15 @@ def test_cases_client(pts):
     iut_device_name = get_unique_name(pts)
     stack = get_stack()
 
+    def init_gatt_client_if_supported():
```
Thalley: The naming is quite inconsistent. Somethings `GATT client` is `gatt_client`, sometimes it's `gatt_cl` and sometimes it's even `gattc`. Consider making the naming consistent

autopts/pybtp/btp/gatt_cl.py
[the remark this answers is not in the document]
Thalley: I think Copilot has a point here.

If we always read the supported commands to verify that the IUT supports the "new" API, why do we need a fallback in the first place?

autopts/wid/gatt_client.py
```diff
@@ -27,6 +27,26 @@
 log = logging.debug
 
 indication_subbed_already = False
+LT2_CASE_TOKEN = "LT2"
```
Thalley: This seems like something that is more generic than GATT

autopts/wid/gatt_client.py
```diff
@@ -27,6 +27,26 @@
...
+VERIFY_STATUS_OK = ""
+VERIFY_STATUS_UNLIKELY_ERROR = "unlikely error"
```
Thalley: Consider putting these into an enum

autopts/wid/gatt_client.py
```diff
@@ -27,6 +27,26 @@
...
+READ_MULT_VAR_EXPECTED_EVENTS = 2
+READ_MULT_VAR_TIMEOUT = 35
+EATT_CONNECT_RETRIES = 3
+
+
+def _is_lt2_case(test_case_name):
+    return LT2_CASE_TOKEN in test_case_name
```
Thalley: Ditto; this is generic and not GATT related. It would be a useful thing to have any use throughout the entire codebase though

autopts/wid/gatt_client.py
```diff
@@ -27,6 +27,26 @@
...
+
+
+def _peer_addr_and_type(test_case_name):
+    if _is_lt2_case(test_case_name):
+        return btp.lt2_addr_get(), btp.lt2_addr_type_get()
+    return btp.pts_addr_get(), btp.pts_addr_type_get()
+
+
+def _wait_pts_connection(timeout=2):
+    return get_stack().gap.wait_for_connection(timeout=timeout, addr=btp.pts_addr_get())
```
Thalley: Ditto; not GATT specific

autopts/wid/gatt_client.py
Thalley: Depends on the IUT perhaps? 

But we cannot ensure that, then there's a logical fallacy here. The test very explicitly requires this, so to pass this test, any IUT must provide some way to ensure this, otherwise we might end up with random failures

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -1005,6 +1068,55 @@ def gatt_cl_read_multiple_var(bd_addr_type, bd_addr, *hdls):
     gatt_cl_command_rsp_succ()
 
 
+def gatt_cl_eatt_connect(bd_addr, bd_addr_type, num=1):
+    """Connect EATT channels using GATTC API, with fallback to legacy GATT API."""
```
Thalley: Please provide full documentation for all new functions, and ideally also type hinting :)

autopts/wid/gatt_client.py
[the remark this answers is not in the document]
Thalley: I think the point here is that there's a mismatch between the current docstring and the implementation. 

The implementation here does not do what the docstring says; instead of preparing the IUT for an incoming request for EATT, we send one instead

Thalley (changes requested):

kedMertens:
Thanks for review, will address concerns

autopts/wid/gatt_client.py
kedMertens: should then gatt_cl_read_multiple_var be extended to return used bearer? will it be acceptable @sjanc, @Thalley?

Thalley: I'm more wondering if the request should contain some way to control the EATT bearer. This will probably be an issue for Zephyr as the API doesn't allow this (but does allow to force a request to be either ATT or EATT), so for `CONFIG_BT_EATT_MAX=1` it could work for Zephyr. 

But that would require the IUT to also send some kind of identifier back to the AutoPTS when calling sending the EATT Connect Cmd

So possible options here, ordered by how well I think they would work:

1. Use identifiers for EATT/ATT bearers to control which bearer is used for a request
2. Use a boolean flag to indicate whether the IUT should use an ATT or EATT bearer for the request
3. Add a boolean flag in the response to indicate which bearer was used
4. Do nothing and hope for the best

kedMertens: I'll give a try and aim for the first one

Thalley: Just keep in mind that I don't think e.g. Zephyr currently has that concept (although I do think it's planned)

kedMertens:
Sorry, it takes awhile. I did address comments and rebase on top of latest main. 
One open topic is confirmation of bearer usage, I did try to resolve as point it out previously in discussions there is no current api in Zephyr. Talked with maintainers and didn't get full picture yet, except that it was asked multiple times and might require major work to have it. I've checked how mynewt is currently handling it, as it is using migrated api, and current approach is aligned. Can we start with that?

autopts/ptsprojects/mynewt/gatt_client_wid.py
```diff
@@ -24,8 +24,8 @@
 log = logging.debug
 
 
-def gattc_wid_hdl(wid, description, test_case_name):
-    log(f'{gattc_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
+def gatt_client_wid_hdl(wid, description, test_case_name):
+    log(f'{gatt_client_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
```
Thalley:
```suggestion
    log(f'{wid}, {description}, {test_case_name}')
```

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -171,6 +197,29 @@ def gatt_cl_dec_disc_rsp(data, attr_type):
...
+btp_rsp_str = {
+    defs.BTP_STATUS_FAILED: VerifyValueStatus.BTP_FAILED.value,
+    defs.BTP_STATUS_UNKNOWN_CMD: VerifyValueStatus.BTP_UNKNOWN_CMD.value,
+    defs.BTP_STATUS_NOT_READY: VerifyValueStatus.BTP_NOT_READY.value,
+}
+
+
+def _att_status_str(status: int) -> str:
+    """Return a stable verification string for an ATT status."""
+    return att_rsp_str.get(
```
Thalley: Please document all arguments for all new functions too :)

autopts/pybtp/types.py
```diff
@@ -531,6 +531,17 @@ class GATTErrorCodes:
...
 
+class VerifyValueStatus(str, Enum):
+    OK = att_rsp_str[0]
+    UNLIKELY_ERROR = att_rsp_str[GATTErrorCodes.unlikely_error]
+    MALFORMED_READ_MULTIPLE_VAR_RESPONSE = "Malformed READ_MULTIPLE_VAR response"
+    BTP_FAILED = "BTP failed"
+    BTP_UNKNOWN_CMD = "BTP unknown command"
+    BTP_NOT_READY = "BTP not ready"
+    UNKNOWN_BTP_STATUS = "BTP status 0x{status:02x}"
+    UNKNOWN_ATT_ERROR = "ATT error 0x{status:02x}"
```
Thalley: Oh I really do not like this :sweat_smile: 

This is a mix of lookups in a global dictionary, literal strings and formatted strings. 

We should _never_ use formatted strings for comparison like this. 
Using a dictionary for some of them also seems like a bad idea.

Finally some of these are related to ATT and some are BTP and some are something else?

autopts/wid/common.py
```diff
@@ -5,6 +5,20 @@
...
 
+LT2_CASE_TOKEN = "LT2"
+
+
+def is_lt2_case(test_case_name: str) -> bool:
+    return LT2_CASE_TOKEN in test_case_name
+
+
+def peer_addr_and_type(test_case_name: str) -> tuple[str, int]:
+    """Return peer address and address type for the active PTS link."""
```
Thalley: Please add proper and full documentation

autopts/wid/gatt_client.py
```diff
@@ -16,21 +16,23 @@
...
 from autopts.pybtp import btp
-from autopts.pybtp.types import GATTErrorCodes, IOCap, WIDParams
+from autopts.pybtp.types import GATTErrorCodes, IOCap, VerifyValueStatus, WIDParams
 from autopts.wid import generic_wid_hdl
+from autopts.wid.common import peer_addr_and_type
 
 log = logging.debug
 
 indication_subbed_already = False
+READ_MULT_VAR_EXPECTED_EVENTS = 2
```
Thalley: This value seems specific to some tests; seems odd to define a global constant for the number of read multiple variable length expected events :o

autopts/wid/gatt_client.py
Thalley: There still doesn't seem to be any control of which bearers are used here

autopts/wid/gatt_client.py
```diff
@@ -1329,11 +1332,140 @@ def hdl_wid_147(params: WIDParams):
...
+            logging.error("%s read_multiple_var failed, ATT status: %r", hdl_wid_147.__name__, status)
+            return False
+
     return True
 
 
+def hdl_wid_148(params: WIDParams):
+    """
+    Please send two Read Multiple Variable Length characteristic requests using these handles: 'XXXX'O 'XXXX'O
+    Required Bearers are "EATT" bearers.
```
Thalley: The code does not match this. This says to use EATT bearers, but we only connect a single EATT bearer

autopts/wid/gatt_client.py
```diff
@@ -1348,6 +1480,60 @@ def hdl_wid_150(params: WIDParams):
...
+    try:
+        sec_level = get_stack().gap.gap_wait_for_sec_lvl_change(
+            level=2,
+            timeout=10,
+            addr=addr,
+        )
+    except Exception as err:
+        logging.error(
+            "%s security level wait failed for %s: %r",
+            hdl_wid_400.__name__,
```
Thalley:
```suggestion
            "ecurity level wait failed for %s: %r",
```

autopts/wid/gatt_client.py
```diff
@@ -1348,6 +1480,60 @@ def hdl_wid_150(params: WIDParams):
...
+            params.test_case_name,
+            err,
+        )
+        return False
+
+    if sec_level != 2:
+        logging.error(
+            "%s required security level 2 for %s, got %r",
+            hdl_wid_400.__name__,
```
Thalley:
```suggestion
        logging.error(
            "required security level 2 for %s, got %r",
```

autopts/wid/gatt_client.py
```diff
@@ -1348,6 +1480,60 @@ def hdl_wid_150(params: WIDParams):
...
+        )
+        return False
+
+    return True
+
+
+def hdl_wid_402(params: WIDParams):
+    """
+    Please initiate an L2CAP Credit Based Connection using LE signaling
+    channel to the PTS.
```
Thalley, line 1491: Is this really the same as EATT?

autopts/wid/gatt_client.py
```diff
@@ -1329,11 +1332,140 @@ def hdl_wid_147(params: WIDParams):
...
+            logging.debug("%s got ATT 0x0e on first read_multiple_var attempt, retrying once", hdl_wid_148.__name__)
+            logging.debug(
+                "%s waiting for link ready before second read_multiple_var attempt",
+                hdl_wid_148.__name__,
+            )
+            if not stack.gap.wait_for_connection(timeout=10, addr=btp.pts_addr_get()):
+                logging.error(
+                    "%s retry aborted: no connection to PTS before second attempt after ATT 0x0e",
+                    hdl_wid_148.__name__,
+                )
```
Thalley, line 1400:
```suggestion
                logging.error(
                    "retry aborted: no connection to PTS before second attempt after ATT 0x0e",
                )
```

The logging system automatically adds the function name

autopts/wid/gatt_client.py
```diff
@@ -1329,11 +1332,140 @@ def hdl_wid_147(params: WIDParams):
...
+
+        if all(status == VerifyValueStatus.OK.value for status in statuses):
+            return True
+
+        if attempt == 0 and all(status == VerifyValueStatus.UNLIKELY_ERROR.value for status in statuses):
+            logging.debug("%s got ATT 0x0e on first read_multiple_var attempt, retrying once", hdl_wid_148.__name__)
+            logging.debug(
+                "%s waiting for link ready before second read_multiple_var attempt",
+                hdl_wid_148.__name__,
+            )
```
Thalley, line 1396:
```suggestion
            logging.debug(
                "waiting for link ready before second read_multiple_var attempt",
            )
```

autopts/wid/gatt_client.py
```diff
@@ -1329,11 +1332,140 @@ def hdl_wid_147(params: WIDParams):
...
+        statuses = [
+            item[0] if isinstance(item, tuple) else item
+            for item in verify_values[:READ_MULT_VAR_EXPECTED_EVENTS]
+        ]
+
+        if all(status == VerifyValueStatus.OK.value for status in statuses):
+            return True
+
+        if attempt == 0 and all(status == VerifyValueStatus.UNLIKELY_ERROR.value for status in statuses):
+            logging.debug("%s got ATT 0x0e on first read_multiple_var attempt, retrying once", hdl_wid_148.__name__)
```
Thalley:
```suggestion
            logging.debug("got ATT 0x0e on first read_multiple_var attempt, retrying once"
```

Thalley: Basically applies to all cases where you've manually added the function name in the log statement :D

Thalley (changes requested):

autopts/wid/gatt_client.py
kedMertens: Good point. It is not the same in general: EATT uses LE Credit Based L2CAP channels, but an LE Credit Based L2CAP channel is not necessarily EATT.

In this WID, PTS is asking for the L2CAP setup as part of the GATT/CL/GAR EATT procedure, so I used the GATT Client `EATT_CONNECT` command rather than the generic L2CAP CoC API. I’ll clarify that in the WID comment/docstring so it’s clear this is EATT-specific setup, not a generic L2CAP CoC handler.

autopts/wid/gatt_client.py
```diff
@@ -1329,11 +1331,140 @@ def hdl_wid_147(params: WIDParams):
     hdl1 = MMI.args[0]
     hdl2 = MMI.args[1]
 
-    btp.gatt_cl_read_multiple_var(btp.pts_addr_type_get(), btp.pts_addr_get(), hdl1, hdl2)
-    btp.gatt_cl_read_multiple_var(btp.pts_addr_type_get(), btp.pts_addr_get(), hdl1, hdl2)
+    if not hdl1 or not hdl2:
+        logging.error("missing read_multiple_var handles for %s: %r", params.test_case_name, MMI.args)
+        return False
```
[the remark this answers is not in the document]
kedMertens: good catch

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -171,6 +199,38 @@ def gatt_cl_dec_disc_rsp(data, attr_type):
...
+    """Return a stable verification string for an ATT status.
+
+    Args:
+        status: ATT status code from a GATT Client BTP event.
+
+    Returns:
+        Known ATT status text, or a stable fallback for unknown ATT status
+        codes.
+    """
+    return att_rsp_str.get(status, ATTVerifyStatus.UNKNOWN_ERROR.value)
```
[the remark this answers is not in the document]
kedMertens: Good catch. I’ll switch GATT_CL to use the shared ATT response mapping from autopts.pybtp.types so known ATT errors stay consistent with the GATT path.

autopts/wid/gatt_client.py
```diff
@@ -1326,14 +1328,143 @@ def hdl_wid_147(params: WIDParams):
...
+    Please send two Read Multiple Variable Length characteristic requests using these handles: 'XXXX'O 'XXXX'O
+    Required Bearers are "EATT" bearers.
+
+    Description: Verify that the Implementation Under Test (IUT) can receive multiple characteristics.
+    """
+    MMI.reset()
+    MMI.parse_description(params.description)
+
+    if len(MMI.args) < 2 or not MMI.args[0] or not MMI.args[1]:
+        logging.error("missing read_multiple_var handles for %s: %r", params.test_case_name, MMI.args)
```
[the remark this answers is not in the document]
kedMertens: will rework with helpers

Thalley:
@sjanc Does AutoPTS allow "fixup" commits? Else @kedMertens will need to squash the commits :)

kedMertens:
> @sjanc Does AutoPTS allow "fixup" commits? Else @kedMertens will need to squash the commits :)

I just hoped, that with fixups it would be easier to see what I addressed. Definitely to squash before merge.

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -39,6 +50,24 @@ def gatt_cl_mtu_exchanged_ev_(gatt_cl, data, data_len):
     logging.debug("%r", data)
 
     fmt = '<B6sB'
+    read_mult_var_fmt = '<B6sBBH'
+
+    if data_len >= struct.calcsize(read_mult_var_fmt):
+        # Some Zephyr builds emit READ_MULTIPLE_VAR_RP payload with MTU_EXCHANGED event opcode.
+        # Reroute only when the payload shape is valid for READ_MULTIPLE_VAR to avoid false positives.
```
Thalley, line 57: This sounds odd.

Can you elaborate on this?

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -39,6 +50,24 @@ def gatt_cl_mtu_exchanged_ev_(gatt_cl, data, data_len):
...
+        )
+        known_btp_statuses = {
+            defs.BTP_STATUS_SUCCESS,
+            defs.BTP_STATUS_FAILED,
+            defs.BTP_STATUS_UNKNOWN_CMD,
+            defs.BTP_STATUS_NOT_READY,
+        }
+        if btp_status in known_btp_statuses and data_len == struct.calcsize(read_mult_var_fmt) + payload_len:
+            gatt_cl_read_mult_var_rsp_ev_(gatt_cl, data, data_len)
+            return
```
Thalley, line 70: What if `btp_status not in known_btp_statuses`? Why do we need to check this in the first place?

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -156,19 +185,36 @@ def gatt_cl_dec_disc_rsp(data, attr_type):
...
-               }
+def _att_status_str(status: int) -> str:
+    """Return a stable verification string for an ATT status.
+
+    Args:
+        status: ATT status code from a GATT Client BTP event.
+
+    Returns:
+        Known ATT status text, or a stable fallback for unknown ATT status
+        codes.
```
Thalley, line 196:
```suggestion
    Returns:
        Known ATT status text, or a stable fallback for unknown ATT status codes.
```

The configured line length for AutoPTS is 127

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -156,19 +185,36 @@ def gatt_cl_dec_disc_rsp(data, attr_type):
...
+
+def _btp_status_str(status: int) -> str:
+    """Return a stable verification string for a BTP status.
+
+    Args:
+        status: BTP status code returned by the IUT.
+
+    Returns:
+        Known BTP status text, or a stable fallback for unknown BTP status
+        codes.
```
Thalley, line 209:
```suggestion
    Returns:
        Known BTP status text, or a stable fallback for unknown BTP status codes.
```

autopts/pybtp/btp/gatt_cl.py
```diff
@@ -422,7 +468,12 @@ def gatt_cl_read_uuid_rsp_ev_(gatt_cl, data, data_len):
                   addr_type, addr, status, data_length, value_length)
 
     if status != 0:
-        add_to_verify_values(att_rsp_str[status])
+        add_to_verify_values(_att_status_str(status))
+        return
+
+    if data_length == 0:
+        add_to_verify_values(ATTVerifyStatus.OK.value)
```
Thalley, line 475: Why not just 

```suggestion
        add_to_verify_values(status)
        return

    if data_length == 0:
        add_to_verify_values(ATTVerifyStatus.OK)
```

autopts/pybtp/types.py
```diff
@@ -531,6 +531,23 @@ class GATTErrorCodes:
     value_not_allowed = 0x13
 
 
+class ATTVerifyStatus(str, Enum):
```
Thalley, line 534: Why not
```suggestion
class ATTVerifyStatus(StrEnum):
```

autopts/wid/common.py
```diff
@@ -5,6 +5,35 @@
...
+def is_lt2_case(test_case_name: str) -> bool:
+    """Return whether a test case name targets the LT2 PTS instance.
+
+    Args:
+        test_case_name: Full PTS test case name.
+
+    Returns:
+        True if the test case is for LT2, otherwise False.
+    """
+    return LT2_CASE_TOKEN in test_case_name
```
Thalley, line 20: I'm wondering if this should be more strict and be
```suggestion
LT2_CASE_TOKEN = "_LT2"


def is_lt2_case(test_case_name: str) -> bool:
    """Return whether a test case name targets the LT2 PTS instance.

    Args:
        test_case_name: Full PTS test case name.

    Returns:
        True if the test case is for LT2, otherwise False.
    """
    return test_case_name.endswith(LT2_CASE_TOKEN)
```

to avoid any false positives from e.g. `PROFILE_LT2_SOMETHING`

autopts/wid/common.py
```diff
@@ -5,6 +5,35 @@
...
+    Args:
+        test_case_name: Full PTS test case name used to select LT1 or LT2.
+
+    Returns:
+        A tuple containing the peer Bluetooth address and LE address type.
+    """
+    if is_lt2_case(test_case_name):
+        return btp.lt2_addr_get(), btp.lt2_addr_type_get()
+
+    return btp.pts_addr_get(), btp.pts_addr_type_get()
```
Thalley, line 35: Outside the scope of this PR, but we should consider a BT_ADDR_LE class that has the BD_ADDR and TYPE in a single object

autopts/wid/gatt_client.py
```diff
@@ -16,21 +16,22 @@
...
 from autopts.pybtp import btp
-from autopts.pybtp.types import GATTErrorCodes, IOCap, WIDParams
+from autopts.pybtp.types import ATTVerifyStatus, GATTErrorCodes, IOCap, WIDParams
 from autopts.wid import generic_wid_hdl
+from autopts.wid.common import peer_addr_and_type
 
 log = logging.debug
 
 indication_subbed_already = False
+READ_MULT_VAR_TIMEOUT = 35
```
Thalley, line 30: Is this based on ATT_TIMEOUT = 30 (seconds) + 5?

autopts/wid/gatt_client.py
```diff
@@ -16,21 +16,22 @@
...
 
 
-def gatt_cl_wid_hdl(wid, description, test_case_name):
-    log(f'{gatt_cl_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
+def gatt_client_wid_hdl(wid, description, test_case_name):
+    log(f'{gatt_client_wid_hdl.__name__}, {wid}, {description}, {test_case_name}')
```
Thalley, line 34:
```suggestion
    log(f'{wid}, {description}, {test_case_name}')
```

autopts/wid/gatt_client.py
```diff
@@ -1316,22 +1318,121 @@ def hdl_wid_141(params: WIDParams):
...
+                addr,
+                hdl1,
+                hdl2,
+            )
+            btp.gatt_cl_read_multiple_var(
+                addr_type,
+                addr,
+                hdl1,
+                hdl2,
+            )
```
Thalley, line 1368: This does still not ensure that EATT is used, and there's no documentation nor comments about why the code is as it is

autopts/wid/gatt_client.py
```diff
@@ -1316,22 +1318,121 @@ def hdl_wid_141(params: WIDParams):
...
+
+        verify_values = btp.get_verify_values()
+        if len(verify_values) < expected_events:
+            logging.error("missing read_multiple_var verify values: %r", verify_values)
+            return False
+
+        statuses = [
+            item[0] if isinstance(item, tuple) else item
+            for item in verify_values[:expected_events]
+        ]
```
Thalley, line 1385: This looks nasty. Why does `verify_values` either return a tuple or a single value?

autopts/wid/gatt_client.py
```diff
@@ -1316,22 +1318,121 @@ def hdl_wid_141(params: WIDParams):
...
+    Please send two Read Multiple Variable Length characteristic requests using these handles: 'XXXX'O 'XXXX'O
+    Required Bearers are "EATT" bearers.
+
+    Description: Verify that the Implementation Under Test (IUT) can receive multiple characteristics.
+    """
+    return _send_gar_read_multiple_var_requests(
+        params,
+        eatt_bearers=2,
+        retry_on_unlikely_error=True,
+    )
```
Thalley, line 1435:
```suggestion
    return _send_gar_read_multiple_var_requests(params, eatt_bearers=2, retry_on_unlikely_error=True)
```

Please configure your IDE and line length to match https://github.com/auto-pts/auto-pts/blob/master/pyproject.toml#L2

This applies many places :) Consider using `ruff` to help format things properly

autopts/wid/gatt_client.py
```diff
@@ -1348,6 +1449,62 @@ def hdl_wid_150(params: WIDParams):
...
+    mode using LE signaling channel.
+
+    Description: Verify that the Implementation Under Test (IUT)
+    can accept L2CAP_CREDIT_BASED_CONNECTION_REQ from PTS.
+    """
+
+    addr, _ = peer_addr_and_type(params.test_case_name)
+
+    try:
+        sec_level = get_stack().gap.gap_wait_for_sec_lvl_change(
```
Thalley, line 1464: What if the `lvl_change` event was already handled and popped?

autopts/wid/gatt_client.py
```diff
@@ -1348,6 +1449,62 @@ def hdl_wid_150(params: WIDParams):
...
+    except Exception as err:
+        logging.error(
+            "security level wait failed for %s: %r",
+            params.test_case_name,
+            err,
+        )
+        return False
+
+    if sec_level != 2:
+        logging.error(
```
Thalley, line 1478: The description does not mention anything about security; why are we checking that here at all? 
What if the security level is 3 or 4?

autopts/wid/gatt_client.py
Thalley: Hmm, the docstrings of `wid`s are usually just the description from PTS. Here you are mixing your own comment with it, making it difficult to determine what is verbatim from PTS and what is written by us. 

Not sure what the best way to handle this is, but the current one seems confusing

Thalley (changes requested):

PR #1748, opened 2026-04-02
btp: simplify logging by removing manual function name references.

### LINKED ISSUE #1344: Logging: Automatically format log statements to have the function name
A lot of places we have log statements as `logging.debug("%s", <function>.__name__)`.

If we want to log the names of functions, we should instead instruct the logger to have that as the format. See https://docs.python.org/3/library/logging.html#logrecord-attributes and search for `funcName`.

agata-ponitka:
Use automatic function name in logs instead of __name__.

fixes https://github.com/auto-pts/auto-pts/issues/1344

### DIFF
--- a/autopts/client.py
+++ b/autopts/client.py
@@ -81,6 +81,13 @@ AUTO_PTS_LOCAL = "AUTO_PTS_LOCAL" in os.environ
 
 def logger_log(self, *args, **kwargs):
     with log_lock:
+        # Add one frame so logs show the real caller (not logger_log).
+        # Adjust stacklevel (kwarg or 7th arg) without duplicating it.
+        if len(args) > 6:
+            args = list(args)
+            args[6] += 1
+        else:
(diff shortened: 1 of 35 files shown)


autopts/client.py
```diff
@@ -80,6 +80,11 @@
 
 def logger_log(self, *args, **kwargs):
     with log_lock:
+        if len(args) > 6:
+            args = list(args)
+            args[6] += 1
+        else:
+            kwargs['stacklevel'] = kwargs.get('stacklevel', 1) + 1
```
Thalley, line 90: What is the purpose of this?

Thalley (approved):

autopts/client.py
agata-ponitka: Without the stacklevel increment, every log line would report "logger_log" as the function name and "client.py" as the source location, regardless of where the log is called.

Thalley: Can you please add that as a comment to the code?

sjanc (approved):

PR #1751, opened 2026-04-08
btp: gap: implement configure_security_mode BTP command

piotrnarajowski:
Implement BTP command that sets IUT into desired security mode and level.

Related issue: #712


autopts/wid/gap.py
```diff
@@ -1331,14 +1331,24 @@ def hdl_wid_225(_: WIDParams):
         Please configures the IUT into LE Secure Connections Only: Mode 1 Level 3.
         Press OK to continue.
     """
+    stack = get_stack()
+
+    if (stack.supported_cmds['GAP'] >> defs.BTP_GAP_CMD_CONFIGURE_SECURITY_MODE) & 1:
```
Thalley: We should consider having a helper function for this, so we can do something like 

```python
if stack.supports_cmd(defs.BTP_GAP_CMD_CONFIGURE_SECURITY_MODE):
```

or something like that

piotrnarajowski: Yeah, also thought about that

piotrnarajowski: Added helper function

szymon-czapracki (approved):

autopts/pybtp/btp/gap.py
```diff
@@ -1172,6 +1174,22 @@ def gap_read_ctrl_info():
     __gap_current_settings_update(_curr_set)
 
 
+def gap_configure_security_mode(mode, level):
+    logging.debug("%s", gap_configure_security_mode.__name__)
```
```suggestion
    logging.debug("")
```

Since there's an incoming PR to automatically add the function name :)

autopts/pybtp/btp/gap.py
```diff
@@ -170,6 +170,8 @@
     "set_ead_key_material": (defs.BTP_SERVICE_ID_GAP, defs.BTP_GAP_CMD_SET_EAD_KEY_MATERIAL, CONTROLLER_INDEX),
     "encrypt_ead_data": (defs.BTP_SERVICE_ID_GAP, defs.BTP_GAP_CMD_ENCRYPT_EAD_DATA, CONTROLLER_INDEX),
     "decrypt_ead_data": (defs.BTP_SERVICE_ID_GAP, defs.BTP_GAP_CMD_DECRYPT_EAD_DATA, CONTROLLER_INDEX),
+    "configure_security_mode": (defs.BTP_SERVICE_ID_GAP,
+                                defs.BTP_GAP_CMD_CONFIGURE_SECURITY_MODE, CONTROLLER_INDEX),
```
```suggestion
    "configure_security_mode": (defs.BTP_SERVICE_ID_GAP, defs.BTP_GAP_CMD_CONFIGURE_SECURITY_MODE, CONTROLLER_INDEX),
```

The configured line length for the project is 127

autopts/ptsprojects/stack/stack.py
```diff
@@ -97,6 +97,9 @@ def is_svc_supported(self, svc):
         svc_value = self.supported_svcs_cmds.get(svc, {}).get("service", 0)
         return (self.supported_svcs & svc_value) > 0
 
+    def is_command_supported(self, service: str, opcode: int):
+        return (self.supported_cmds[service] >> opcode) & 1
```
Thalley: Please add docstring :)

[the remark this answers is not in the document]
Thalley: A better solution is to use an enum for the services, instead of using string literals :)

autopts/wid/gap.py
```diff
@@ -1331,14 +1331,24 @@ def hdl_wid_225(_: WIDParams):
         Please configures the IUT into LE Secure Connections Only: Mode 1 Level 3.
         Press OK to continue.
     """
+    stack = get_stack()
+
+    if stack.is_command_supported('GAP', defs.BTP_GAP_CMD_CONFIGURE_SECURITY_MODE):
+        btp.gap_configure_security_mode(1, 3)
```
Thalley: Suggest to define enums for the mode and levels :)

autopts/wid/gap.py
```diff
@@ -1331,14 +1331,24 @@ def hdl_wid_225(_: WIDParams):
...
+
     return True
 
 
 def hdl_wid_226(_: WIDParams):
     """
-        Please configures the IUT into LE Secure Connections Only: Mode 1 Level 3.
+        Please configures the IUT into LE Secure Connections Only: Mode 1 Level 4.
         Press OK to continue.
```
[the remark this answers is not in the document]
piotrnarajowski: I think this is actuall text PTS is displaying so I will leave this as is.

autopts/pybtp/btp/gap.py
piotrnarajowski: Done

autopts/pybtp/btp/gap.py
piotrnarajowski: Done

autopts/ptsprojects/stack/stack.py
piotrnarajowski: Done

autopts/pybtp/btp/gap.py
```diff
@@ -1172,6 +1174,22 @@ def gap_read_ctrl_info():
...
 
 
+def gap_configure_security_mode(mode, level):
+    logging.debug("%s", gap_configure_security_mode.__name__)
+
+    iutctl = get_iut()
+
+    data = bytearray()
+    mode_ba = struct.pack('b', mode)
+    level_ba = struct.pack('b', level)
```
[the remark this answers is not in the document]
piotrnarajowski: Done

doc/btp_gap.txt
```diff
@@ -892,6 +892,18 @@ Commands and responses:
...
+
+                Controller Index:       <controller id>
+                Command parameters:     Mode (1 octet)
+                                        Level (1 octet)
+                Return Parameters:      <none>
+
+                This command can be used to configure IUT into desired
+                LE Security Mode and Level.
+
+                In case of an error, the error response will be returned
```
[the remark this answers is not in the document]
piotrnarajowski: Done

autopts/wid/gap.py
piotrnarajowski: Hmm... not sure how to do this in this context. Did you mean something like that?
```
class SecurityModeLevel(Enum):
MODE_1_LEVEL_1 = (1, 1)
MODE_1_LEVEL_2 = (1, 2)

etc...
```

Thalley: I was thinking a `BtSecurityMode(IntEnum)` and a `BtSecurityLevel(IntEnum)`

We can make an additional (data)class that defines all possible combinations as well, as not all modes work with all levels. 

What I wanted to avoid is literal integers like `1` and `3` to indicate specific modes and levels (although for this specific case, given that modes and levels are denoted by integers in the first place, is arguably OK, but then `gap_configure_security_mode` should at least document the valid values and throw and error if not provided)

autopts/ptsprojects/stack/stack.py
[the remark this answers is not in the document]
piotrnarajowski: Service ID's are already defined as constants in defs.py. Moving to an enum and deleting old defines from defs.py would involve a very large refactor. On the other hand, introducing an enum while keeping the old defines might create confusion.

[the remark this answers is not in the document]
Thalley: It would be a large, but fairly trivial, refactor. I would also say that it's outside the scope of this PR, but it was a better solution that what Copilot suggested IMO. Although the suggestion from Copilot is something we can have until then

piotrnarajowski: OK, now they are implemented correctly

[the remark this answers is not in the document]
piotrnarajowski: Went with copilot suggestion

autopts/wid/gap.py
piotrnarajowski: Security modes and levels are now documented in btp_gap.txt and their usage is verified in `gap_configure_security_mode`

autopts/pybtp/btp/gap.py
```diff
@@ -1172,6 +1173,41 @@ def gap_read_ctrl_info():
...
+    logging.debug("")
+
+    valid_security_levels = {
+        1: [1, 2, 3, 4],
+        2: [1, 2],
+        3: [1, 2, 3]
+    }
+
+    if level not in valid_security_levels.get(mode, []):
+        raise ValueError(f"Error: Mode {mode} with Level {level} is not a supported combination.")
```
Thalley: This may also throw an `KeyError` if `mode` is e.g. `5`. Is that intentionally not handled here? It's not necessarily an issue, as the call will get the exception/error in that case

autopts/pybtp/btp/gap.py
```diff
@@ -1172,6 +1173,41 @@ def gap_read_ctrl_info():
     __gap_current_settings_update(_curr_set)
 
 
+def gap_configure_security_mode(mode: int, level: int) -> None:
+    """
+    Configures IUT into specified security mode and level
+
+    Args:
+        mode (int): The security mode to configure (1, 2, or 3).
+        level (int): The security level to configure (allowed values depend on the mode).
```
Thalley: Suggest to define the valid `level` values as well

Thalley: One of the reasons why I suggested enums are that not only does it handle all value checking automatically, but it also allows us to properly define and document each mode and level

This is on par with most code in AutoPTS so I think it's OK, but I do still think that, when a small subset of values are allowed and properly defined by the Bluetooth specifications, an enum would have been better :)

autopts/wid/gap.py
piotrnarajowski: Recent changes introduce `class SecurittyModeLevel(Enum).` Having all that information inside one enum is IMO the best approach, because it defines mode and level combinations and handles value checking. Creating two separate enums for mode and level and then making additional class that defines possible combinations is an overkill

autopts/pybtp/btp/gap.py
piotrnarajowski: New approach is introduced, please see comment above

autopts/pybtp/types.py
```diff
@@ -751,3 +751,17 @@ class BearerTech(IntEnum):
...
+    MODE_1_LEVEL_2 = (1, 2)
+    MODE_1_LEVEL_3 = (1, 3)
+    MODE_1_LEVEL_4 = (1, 4)
+
+    MODE_2_LEVEL_1 = (2, 1)
+    MODE_2_LEVEL_2 = (2, 2)
+
+    MODE_3_LEVEL_1 = (3, 1)
+    MODE_3_LEVEL_2 = (3, 2)
+    MODE_3_LEVEL_3 = (3, 3)
```
Thalley, line 767: Looks good - Some things that can be considered is to use NamedTuples or properties if we want to access the individual values, e.g. 

```python
from enum import Enum
from typing import NamedTuple

class SecurityTuple(NamedTuple):
    mode: int
    level: int

class SecurityModeLevel(Enum):
    MODE_1_LEVEL_1 = SecurityTuple(1, 1)
    MODE_1_LEVEL_2 = SecurityTuple(1, 2)
    MODE_1_LEVEL_3 = SecurityTuple(1, 3)
    MODE_1_LEVEL_4 = SecurityTuple(1, 4)

    MODE_2_LEVEL_1 = SecurityTuple(2, 1)
    MODE_2_LEVEL_2 = SecurityTuple(2, 2)

    MODE_3_LEVEL_1 = SecurityTuple(3, 1)
    MODE_3_LEVEL_2 = SecurityTuple(3, 2)
    MODE_3_LEVEL_3 = SecurityTuple(3, 3)

# Usage
member = SecurityModeLevel.MODE_1_LEVEL_1
print(member.value.mode)   # 1
print(member.value.level)  # 1
```

or 

```python
class SecurityModeLevel(Enum):
    MODE_1_LEVEL_1 = (1, 1)
    MODE_1_LEVEL_2 = (1, 2)
    MODE_1_LEVEL_3 = (1, 3)
    MODE_1_LEVEL_4 = (1, 4)

    MODE_2_LEVEL_1 = (2, 1)
    MODE_2_LEVEL_2 = (2, 2)

    MODE_3_LEVEL_1 = (3, 1)
    MODE_3_LEVEL_2 = (3, 2)
    MODE_3_LEVEL_3 = (3, 3)

    @property
    def mode(self):
        return self.value[0]
    
    @property
    def level(self):
        return self.value[1]

# Usage
member = SecurityModeLevel.MODE_1_LEVEL_1
print(member.mode)   # 1
print(member.level)  # 1
```

But we can postpone that until we have a usecase for it

autopts/wid/gap.py
```diff
@@ -1331,14 +1332,24 @@ def hdl_wid_225(_: WIDParams):
...
+    if stack.is_command_supported('GAP', defs.BTP_GAP_CMD_CONFIGURE_SECURITY_MODE):
+        btp.gap_configure_security_mode(SecurityModeLevel.MODE_1_LEVEL_3)
+
     return True
 
 
 def hdl_wid_226(_: WIDParams):
     """
-        Please configures the IUT into LE Secure Connections Only: Mode 1 Level 3.
+        Please configures the IUT into LE Secure Connections Only: Mode 1 Level 4.
```
Thalley, line 1349: Just noticed now that the level here changed; is it Level 4 for all tests, or will it be Level 3 for some given the previous text?

piotrnarajowski: I think there was a typo in description inside `hdl_wid_226`. Probably I should review this. Or we could look for the modes and level in `params.description` to make it futureproof if anything changes

autopts/pybtp/btp/gap.py
```diff
@@ -1172,6 +1174,33 @@ def gap_read_ctrl_info():
...
+    """
+    Configures IUT into specified security mode and level
+
+    Args:
+        mode_level (SecurityModeLevel): Security mode and level combination
+
+    Returns:
+        None
+    """
+    logging.debug("")
```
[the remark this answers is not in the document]
piotrnarajowski: there's an incoming PR to automatically add the function name hence the change

autopts/wid/gap.py
Thalley: The following tests use this WID handler: 

GAP/SEC/SEM/BV-21-C 
GAP/SEC/SEM/BV-22-C 
GAP/SEC/SEM/BV-26-C 
GAP/SEC/SEM/BV-27-C 
GAP/SEC/SEM/BV-58-C 
GAP/SEC/SEM/BV-61-C 
GAP/SEC/SEM/BV-64-C 
GAP/SEC/SEM/BV-67-C

We should verify, and if the level may change between 3 and 4, we should catch that with a regex or something

piotrnarajowski: All listed tests share same description that requires configuring IUT to security mode 1 level 4. So it was a typo then, or maybe a it changed with PTS update some time ago, hard to tell

autopts/pybtp/types.py
piotrnarajowski: Added properties

autopts/ptsprojects/stack/stack.py
```diff
@@ -97,6 +97,24 @@ def is_svc_supported(self, svc):
...
+
+        Returns:
+            bool: True if command is supported by IUT, False if not
+        """
+        try:
+            return bool((self.supported_cmds[service] >> opcode) & 1)
+        except KeyError:
+            log(f'{service} BTP Service not found')
+            return False
+
```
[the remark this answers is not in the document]
piotrnarajowski: Implemented these suggestions

Thalley (approved):

autopts/wid/gap.py
```diff
@@ -1331,14 +1332,24 @@ def hdl_wid_225(_: WIDParams):
         Please configures the IUT into LE Secure Connections Only: Mode 1 Level 3.
         Press OK to continue.
     """
+    stack = get_stack()
+
+    if stack.is_command_supported('GAP', defs.BTP_GAP_CMD_CONFIGURE_SECURITY_MODE):
+        btp.gap_configure_security_mode(SecurityModeLevel.MODE_1_LEVEL_3)
```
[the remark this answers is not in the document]
piotrnarajowski: IMO we should prioritize instructions in the MMI in this case, @Thalley can you resolve this?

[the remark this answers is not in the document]
Thalley: From the core spec:

> For certain services that require LE security mode 1, levels 2 or 3, a device may enforce
the use of LE Secure Connections pairing before those services can be used.

So we do not need Level 4 for LE Secure connections only. That being said, perhaps that needs to be an argument to the `gap_configure_security_mode`, i.e. add flags to only allow LE Secure Connections, only allow OOB data, etc. 

Maybe mode and level isn't granular enough for these tests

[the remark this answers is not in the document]
piotrnarajowski: Decided to add flags parameter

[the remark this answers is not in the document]
Thalley: You've added flags as `Flags are reserved for future use.`, but in this case we need to provide an LE Secure Connections flag, so suggest to define that flag and use it here

Thalley (changes requested):
We already need to define and use a flag for "LE Secure Connections only"

autopts/wid/gap.py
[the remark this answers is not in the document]
piotrnarajowski: Do you mean something like:
```
Bit 0 - Secure Connections Only: False
Bit 1 - Secure Connections Only: True
```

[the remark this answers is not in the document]
Thalley: Yes, but I'd just define it as a single bit, i.e. 

```
Bit 0 - Secure Connections Only
```

So that if Bit 0 is set, Secure Connections Only must be set/allowed :)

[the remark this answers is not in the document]
piotrnarajowski: defined `class SecurityFlags` and updated `gap_configure_security_mode()` accordingly

autopts/pybtp/types.py
```diff
@@ -751,3 +751,29 @@ class BearerTech(IntEnum):
...
+    def mode(self):
+        return self.value[0]
+
+    @property
+    def level(self):
+        return self.value[1]
+
+
+class SecurityFlags(IntFlag):
+    LE_SECURE_CONNECTIONS_ONLY = defs.BIT(0)
```
Thalley, line 780: Should we have something like
```suggestion
class SecurityFlags(IntFlag):
    NONE = 0
    LE_SECURE_CONNECTIONS_ONLY = defs.BIT(0)
```

So that we can do `flags: SecurityFlags = SecurityFlags.NONE`

piotrnarajowski: Done

Thalley (approved):
Nice work :)

piotrnarajowski:
Resolved conflicts

piotrnarajowski:
Resolved conflicts

piotrnarajowski:
@sjanc PTAL

piotrnarajowski:
Rebased and deleted commit implementing is_command_supported since it is already implemented by https://github.com/auto-pts/auto-pts/pull/1813

sjanc (approved):

PR #1767, opened 2026-04-27
Bluez: Add BAP/UCL support

fdanis-oss:
This adds a new `Setup` command used to configure the ASCS service for IUT stack managing states internally like `BlueZ`.

This allows to set the audio profile to use for BAP/UCL/SCC tests requiring specific profiles, and to desynchronized
linked transports to be able to acquire streams separately for BAP/UCL/STR/BV-54[36]-C.

### DIFF
--- a/autopts/ptsprojects/bluez/iutctl.py
+++ b/autopts/ptsprojects/bluez/iutctl.py
@@ -58,6 +58,7 @@ class IUTCtl:
         self.iut_process = None
         self.audio_profile = None
         self.audio_process = None
+        self.desynchronized = False
 
         self.stack = Stack()
         self.stack.synch_init()
@@ -136,7 +137,7 @@ class IUTCtl:
                 latency = 1
             elif self.audio_profile == "high-reliability":
                 latency = 3
-            vendor.vendor_ascs_setup(latency)
+            vendor.vendor_ascs_setup(latency, self.desynchronized)
             return
         elif self.external_audio != "wireplumber":
             return
@@ -173,6 +174,9 @@ class IUTCtl:
 
     def set_audio_profile(self, profile):
         self.audio_profile = profile
+
+    def set_desynchronized(self, desynchronized):
+        self.desynchronized = desynchronized
 
 
 def get_iut():
--- a/autopts/ptsprojects/bluez/iutctl.py
+++ b/autopts/ptsprojects/bluez/iutctl.py
@@ -19,6 +19,7 @@ import os
 import shlex
 import subprocess
 
+import autopts.ptsprojects.bluez.vendor as vendor
 from autopts.ptsprojects.stack import Stack
 from autopts.pybtp import defs
 from autopts.pybtp.iutctl_common import BTP_ADDRESS, BTPSocketSrv, BTPWorker
@@ -126,7 +127,18 @@ class IUTCtl:
         return self.external_audio
 
     def start_audio(self):
-        if self.external_audio != "wireplumber":
+        if self.external_audio is None:
+            if not (self.stack.supported_cmds['VENDOR'] & defs.BIT(vendor.BTP_VENDOR_CMD_ASCS_SETUP)):
+                raise BTPInitError("Vendor ASCS setup command not supported")
+
+            latency = 0
+            if self.audio_profile == "low-latency":
+                latency = 1
+            elif self.audio_profile == "high-reliability":
+                latency = 3
+            vendor.vendor_ascs_setup(latency)
+            return
+        elif self.external_audio != "wireplumber":
             return
 
         logging.debug("Starting external audio with profile: %s", self.audio_profile)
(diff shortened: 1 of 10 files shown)


fdanis-oss:
@sjanc @Thalley Could you please review this PR?

doc/btp_ascs.txt
```diff
@@ -192,6 +192,18 @@ Commands and responses:
...
 
+        Opcode 0x0c: Setup
+
+                Controller Index:       <controller id>
+                Command parameters:     Target Latency (1 octet)
+                                        Desynchronized (1 octet)
+                Response parameters:    <None>
+
+                This command is used to configure the ASCS service for IUT stack
+                managing states internally, e.g. BlueZ.
```
Thalley: This seems very specific to a single platform. 

Term like `Desynchronized` is not defined by ASCS

Target Latency is something that should exist for `Opcode 0x02: ASCS Configure Codec` (which unfortunately was not added when that command was added). 

Since you are adding support for BAP/UCL, any new commands should really be in the BAP service, instead of ASCS IMO

autopts/pybtp/btp/ascs.py
```diff
@@ -211,6 +212,18 @@ def ascs_preconfig_qos(cig_id, cis_id, sdu_interval, framing, max_sdu,
     ascs_command_rsp_succ()
 
 
+def ascs_setup(target_latency, desynchronized):
+    logging.debug(f"{ascs_setup.__name__}")
```
Thalley:
```suggestion
    logging.debug("")
```

Thalley (dismissed):

doc/btp_ascs.txt
fdanis-oss: Yes, `Desynchronized` is no defined in LE Audio profiles, it just allows to request that audio endpoints are not acquired synchronously for a stack which managed this internally and can't be controlled from btpclient on this point. And yes currently only BlueZ.

`Opcode 0x02: ASCS Configure Codec` is already used during bap operation but way to late for BlueZ.

This new command is used to both pass specific configurations but also start/register the audio endpoints, which iiuc will be also requested to pass the ASCS tests.

fdanis-oss: Is there a way to create new command specific to a platform for a service?

Thalley: No idea; I think the general idea behind BTW is that it's platform agnostic. @sjanc ?

Thalley: Maybe one idea could be to have like a "vendor specific data" when initializing the individual BTP services?

autopts/ptsprojects/bluez/iutctl.py
```diff
@@ -126,7 +127,18 @@ def get_external_audio_support(self):
...
-        if self.external_audio != "wireplumber":
+        if self.external_audio is None:
+            if not (self.stack.supported_cmds['ASCS'] & defs.BIT(defs.BTP_ASCS_CMD_SETUP)):
+                raise BTPInitError("ASCS setup command not supported")
+
+            latency = 0
+            if self.audio_profile == "low-latency":
+                latency = 1
+            elif self.audio_profile == "high-reliability":
+                latency = 3
```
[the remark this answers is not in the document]
fdanis-oss: Fallback to `latency = 0` is expected when no specific latency is requested.

autopts/ptsprojects/bluez/iutctl.py
```diff
@@ -126,7 +127,18 @@ def get_external_audio_support(self):
         return self.external_audio
 
     def start_audio(self):
-        if self.external_audio != "wireplumber":
+        if self.external_audio is None:
+            if not (self.stack.supported_cmds['ASCS'] & defs.BIT(defs.BTP_ASCS_CMD_SETUP)):
```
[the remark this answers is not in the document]
fdanis-oss: If `self.stack.supported_cmds` doesn't contain an "ASCS" entry it's expected this to raise `KeyError`

doc/btp_ascs.txt
fdanis-oss: @sjanc Do you have any advice on this, please?

Thalley (commented):
I like this approach much better :) I think it may also be useful for other cases; the only thing we need to keep in mind is to ensure that we don't rely too much on the vendor specific commands ^_^

Thalley:
@fdanis-oss Do you know if the Codecoup bot can run a set of tests to validate this PR? Do you have a small subset of BAP/UCL tests that you use to verify your changes?

fdanis-oss:
@Thalley I don't know for bot run, but this also needs some changes in BlueZ btpclient.

I was able to pass the BAP/CL and BAP/UCL tests, except the one requesting multiple PTS instances as I don't have access to them today.

sjanc (approved):

Issue #1756, opened 2026-04-10
Figure out how to handle tests from IOPT suite


sjanc:
Those are defined in separate test set 'Interoperable Product Testing (IOPT)' and all of them start with IOPT prefix, but are testing specific services, eg  IOPT/DIS/SR/GATTDB/BV-01-I. Those tests seems to only validate existence and correctness of services/characteristics.

This means that we may have to map those to overlays somehow, either:
- have single overlay with "all enabled" for all IOPT tests , may simply include other overlays 
- have each IOPT/FOO test(s) mapped to specific overlay in service overlay (where needed)

PR #1769, opened 2026-04-30
bot/iut_config: Map IOPT test cases to zephyr overlays

szymon-czapracki:
Add IOPT LE Audio prefixes to the LE Audio overlay mapping and route IOPT Mesh cases through the Mesh overlay.
This prevents IOPT cases from falling back to the default prj.conf during AutoPTS runs.

### DIFF
--- a/autopts/bot/iut_config/zephyr.py
+++ b/autopts/bot/iut_config/zephyr.py
@@ -151,9 +151,32 @@ iut_config = {
             # The overlay file exists in zephyr repo. Leave this empty.
         },
         "test_cases": [
-            'AICS', 'ASCS', 'BAP', 'BASS', 'CAP', 'CAS', 'CCP', 'CSIP', 'CSIS', 'GMAP',
-            'GMCS', 'GTBS', 'HAP', 'HAS', 'IAS', 'MCP', 'MCS', 'MICP', 'MICS', 'PACS',
-            'PBP', 'TBS', 'TMAP', 'VCP', 'VCS', 'VOCS',
+            'AICS', 'IOPT/AICS',
+            'ASCS', 'IOPT/ASCS',
+            'BAP', 'IOPT/BAP',
+            'BASS', 'IOPT/BASS',
+            'CAP', 'IOPT/CAP',
+            'CAS', 'IOPT/CAS',
+            'CCP', 'IOPT/CCP',
+            'CSIP', 'IOPT/CSIP',
+            'CSIS', 'IOPT/CSIS',
+            'GMAP', 'IOPT/GMAP',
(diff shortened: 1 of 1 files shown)


autopts/bot/iut_config/zephyr.py
```diff
@@ -154,6 +154,9 @@
             'AICS', 'ASCS', 'BAP', 'BASS', 'CAP', 'CAS', 'CCP', 'CSIP', 'CSIS', 'GMAP',
             'GMCS', 'GTBS', 'HAP', 'HAS', 'IAS', 'MCP', 'MCS', 'MICP', 'MICS', 'PACS',
             'PBP', 'TBS', 'TMAP', 'VCP', 'VCS', 'VOCS',
+            'IOPT/AICS', 'IOPT/ASCS', 'IOPT/BASS', 'IOPT/CAS', 'IOPT/CSIS', 'IOPT/HAS',
+            'IOPT/IAS', 'IOPT/GMCS', 'IOPT/MICS', 'IOPT/PACS', 'IOPT/PBP', 'IOPT/TBS',
+            'IOPT/GTBS', 'IOPT/TMAP', 'IOPT/VCS', 'IOPT/VOCS', 'IOPT/OTS',
```
Thalley: Can this be made more generic? Is this also something that exists for non-audio profiles like GAP? 

I'm wondering if we should always just treat `IOPT/XYZ` as `XYZ` in all cases, and not just for LE Audio for Zephyr here?

szymon-czapracki: IOPT tests are usually basic checks that verify whether a service exists and exposes the expected characteristics. For LE Audio-related IOPT tests, the audio overlay seems like the right fit. 

There are IOPT tests for non-audio services too: `IOPT/MESH` is added to the Mesh group because it needs the Mesh configuration, while `IOPT/DIS` does not need a separate entry because DIS is already enabled by the regular `prj.conf`. I also found IOPT tests for HRS and BAS, but AutoPTS does not support them yet, and they are not in the Zephyr test plan.


I would prefer to keep this explicit rather than adding generic `IOPT/XYZ -> XYZ` handling. 
Some IOPT cases may need service-specific tweaks, and I do not think the extra recognition logic adds much value right now.

Thalley (approved):
No blockers from me, although I'm not 100% sure about the solution :)

sjanc:
> No blockers from me, although I'm not 100% sure about the solution :)

this is not perfect but allows to correctly apply overlays also for individual tests

sjanc (approved):

szymon-czapracki:
Resolves https://github.com/auto-pts/auto-pts/issues/1756

PR #1773, opened 2026-05-12
Fix new oob test cases

m-alperen-sener:
- Update IXITs for v8.3.0
- Fix for legacy pairing OOB method tests

### DIFF
--- a/autopts/wid/sm.py
+++ b/autopts/wid/sm.py
@@ -14,6 +14,7 @@
 #
 
 import logging
+import os
 import re
 
 from autopts.ptsprojects.stack import get_stack
@@ -199,10 +200,22 @@ def hdl_wid_146(_: WIDParams):
     return True
 
 
-def hdl_wid_147(_: WIDParams):
+def hdl_wid_147(params: WIDParams):
     """
     Please enter 16 bytes IUT's OOB Data (confirmation).
     """
+    if params.test_case_name in ['SM/CEN/OOB/BV-10-C', 'SM/PER/OOB/BV-11-C']:
+        # This behaviour is required by core v6.3 and later.
+        # Ref: Core_v6.3 Vol 3, Part H section 2.3.5.4
+        # The TK value shall be a 128-bit random number.
+        # However this can not be forced on the Host SM layer, since OOB is not limited to be supplied
+        # any application level method.
+        # if test case is ble legacy pairing with OOB data, we need to generate a 16bytes random number
+        # each time this WID is called to set the OOB data.
+        random_number = ''.join(f'{byte:02x}' for byte in os.urandom(16))
+        btp.gap_oob_legacy_set_data(random_number)
+        return random_number
+
     r, c = btp.gap_oob_sc_get_local_data()
     return c
 
--- a/autopts/ptsprojects/zephyr/gap.py
+++ b/autopts/ptsprojects/zephyr/gap.py
@@ -82,10 +82,6 @@ def set_pixits(ptses):
 
     # Set GAP common PIXIT values
     pts.set_pixit("GAP", "TSPX_bd_addr_iut", "DEADBEEFDEAD")
-    pts.set_pixit("GAP", "TSPX_bd_addr_PTS", "C000DEADBEEF")
-    pts.set_pixit("GAP", "TSPX_broadcaster_class_of_device", "100104")
-    pts.set_pixit("GAP", "TSPX_observer_class_of_device", "100104")
-    pts.set_pixit("GAP", "TSPX_central_class_of_device", "100104")
     pts.set_pixit("GAP", "TSPX_security_enabled", "FALSE")
     pts.set_pixit("GAP", "TSPX_delete_link_key", "FALSE")
     pts.set_pixit("GAP", "TSPX_iut_setup_att_over_br_edr", "FALSE")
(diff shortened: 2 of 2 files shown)


autopts/wid/sm.py
```diff
@@ -199,10 +200,22 @@ def hdl_wid_146(_: WIDParams):
...
     Please enter 16 bytes IUT's OOB Data (confirmation).
     """
+    if params.test_case_name in ['SM/CEN/OOB/BV-10-C', 'SM/PER/OOB/BV-11-C']:
+        # This behaviour is required by core v6.3 and later.
+        # Ref: Core_v6.3 Vol 3, Part H section 2.3.5.4
+        # The TK value shall be a 128-bit random number.
+        # However this can not be forced on the Host SM layer, since OOB is not limited to be supplied
+        # any application level method.
+        # if test case is ble legacy pairing with OOB data, we need to generate a 16bytes random number
+        # each time this WID is called to set the OOB data.
```
Thalley, line 214: How does PTS validate that this is a random number? I'm not opposed to this change at all, but I'm just wondering if they just change if it is changed and then just increment the value for each call :D

autopts/wid/sm.py
```diff
@@ -199,10 +200,22 @@ def hdl_wid_146(_: WIDParams):
...
+        random_number = ''.join(f'{byte:02x}' for byte in os.urandom(16))
+        btp.gap_oob_legacy_set_data(random_number)
+        return random_number
+
     r, c = btp.gap_oob_sc_get_local_data()
```
Thalley, line 219: It looks kind of... weird, that we can only _set_ but not _get_ legacy OOB data, and similarly we can only _get_ but not _set_ LESC OOB data, but I guess there's a reason for it?

Thalley:
@m-alperen-sener Please let me know if I should trigger the bot for these tests, although I guess they may fail until Codecoup updates the PTS version to 8.3? (cc @sjanc )

sjanc:
yeap, we are currently at 8.12 so those won't run at all (since they are not yet present)

autopts/wid/sm.py
m-alperen-sener: I think it is the nature OBB, it is distributed out of band not generated locally so we never need to get it from device.

LESC part; I don't know that part, but guess it is due to R,C value pair.

autopts/wid/sm.py
m-alperen-sener: it run test 5 times and checks against the used OOB data, I did not tested the test itself to understand, because I am against this test cases actually, they are useless IMO.

OOB data is never under responsibility of ble host  let alone SM layer, for qualification this test case adds no value whatsoever. Any restriction that is put on OOB can not prevent user to do whatever they want; and ble host can never check if OOB is random: it requires infinite memory.

maybe we should create a errata for this; it is OK to in specification to state and dictate OOB to be random, but not a test case for it. Maybe it should be category D test case.

Any input?

autopts/wid/sm.py
sjanc: > I think it is the nature OBB, it is distributed out of band not generated locally so we never need to get it from device.

this is exactly that,  for legacy OOB pairing one set full TK value (on both sides),  while for LESC you get rand/confirm to be sent to remote via OOB (either one peer or both)

autopts/wid/sm.py
sjanc: I'm fine with how this is done now, point of this test is not to test TK randomness but pairing process :)

image

Thalley (approved):
No blockers from me

KyraLengfeld (approved):

sjanc:
why those random IXITs were removed? I check few random and they all exist in PTS 8.13 workspace..

sjanc:
scratch that, it looks like those IXIT exists only if converting existing workspace, but new one created from scratch have those IXITs removed

Thalley:
> scratch that, it looks like those IXIT exists only if converting existing workspace, but new one created from scratch have those IXITs removed

@sjanc How do you feel about requesting that PTS adds some prefix or infix to separate the IXIT values defined by the specs, and the IXIT values that PTS define outside of the specs? 

That way we may be able to automatically verify and remove all non-PTS IXIT values, and then maybe make it easier to determine if a PTS IXIT value should be removed or not as well

sjanc:
PTS throws exception if we try to set non existing IXIT so it is easy to spot :)

I'd rather continue work on https://github.com/auto-pts/auto-pts/pull/1720 where we only set IXITs if it is necessary to use non-default value

Thalley:
> PTS throws exception if we try to set non existing IXIT so it is easy to spot :)
> 
> I'd rather continue work on #1720 where we only set IXITs if it is necessary to use non-default value

Sure, but more than once I've tried to figure out what the IXIT value means, and trying to determine this from reading the specs and test specs, only to then figure out that it's a PTS specific IXIT value that is all-too-often poorly documented :P

sjanc:
> Sure, but more than once I've tried to figure out what the IXIT value means, and trying to determine this from reading the specs and test specs, only to then figure out that it's a PTS specific IXIT value that is all-too-often poorly documented :P

yeah, and I think they are slowly moving towards this

Thalley:
> > Sure, but more than once I've tried to figure out what the IXIT value means, and trying to determine this from reading the specs and test specs, only to then figure out that it's a PTS specific IXIT value that is all-too-often poorly documented :P
> 
> yeah, and I think they are slowly moving towards this

Can you elaborate? Moving towards marking PTS specific IXIT values, or making more of their own (undocumented) IXIT values?

sjanc:
they no longer add new IXITs without first updating TS (at least this was the case in my support tickets)

m-alperen-sener:
Is there a timeline for master workspace v8.3.0 migration?

sjanc:
> Is there a timeline for master workspace v8.3.0 migration?

https://github.com/auto-pts/auto-pts/pull/1778

sjanc (approved):

PR #1791, opened 2026-06-11
zephyr: Sync workspace with latest qualification plan

sjanc:
This includes Core Spec 6.3 and thus enables some additional tests.


sjanc:
based on https://github.com/zephyrproject-rtos/zephyr/pull/111036

Thalley (approved):
A much easier diff to read than the Zephyr PR :D

sjanc:
> A much easier diff to read than the Zephyr PR :D

yeap, I've created support ticket for this, will see how it goes  :)

Thalley:
> > A much easier diff to read than the Zephyr PR :D
> 
> yeap, I've created support ticket for this, will see how it goes :)

Worst case is that we make a small XML script that will just parse and order everything manually. Should be pretty easy to do

Issue #1317, opened 2024-12-06
Deprecate `BAP_SEND`


Thalley:
The `BAP_SEND` command has a few issues:
1) It assumes unicast given that it uses an ASE ID and an address
2) It has a maximum length of 255 which is too small for some of the BAP configurations when there are multiple channels in a single stream
3) It is used in a way so that we are only sending a few times (and it's unclear whether each of these are SDUs or part of SDUs). 

It is a lot easier for the IUT to determine how and what to send, and without the latency of autopts. See also the following Discord discussion: https://discord.com/channels/720317445772017664/1314275603091820544

PR #1799, opened 2026-07-01
BAP: Add explicit BSRC STR config mapping

ciascai:
Use explicit AC12/AC13/AC14 STR mappings and
typed config IDs. Update BV-35 to 50 AC14 handling
to scale max SDU by channel count.
Sets BAP PIXIT streaming confirmation method to
“By Playing” for both PTS instances.

BAP/BSRC/STR/BV-01-C
BAP/BSRC/STR/BV-02-C
BAP/BSRC/STR/BV-03-C
BAP/BSRC/STR/BV-04-C
BAP/BSRC/STR/BV-05-C
BAP/BSRC/STR/BV-06-C
BAP/BSRC/STR/BV-07-C
BAP/BSRC/STR/BV-08-C
BAP/BSRC/STR/BV-09-C
BAP/BSRC/STR/BV-10-C
BAP/BSRC/STR/BV-11-C
BAP/BSRC/STR/BV-12-C
BAP/BSRC/STR/BV-13-C
BAP/BSRC/STR/BV-14-C
BAP/BSRC/STR/BV-15-C
BAP/BSRC/STR/BV-16-C

BAP/BSRC/STR/BV-35-C
BAP/BSRC/STR/BV-36-C
BAP/BSRC/STR/BV-37-C
BAP/BSRC/STR/BV-38-C
BAP/BSRC/STR/BV-39-C
BAP/BSRC/STR/BV-40-C
BAP/BSRC/STR/BV-41-C
BAP/BSRC/STR/BV-42-C
BAP/BSRC/STR/BV-43-C
BAP/BSRC/STR/BV-44-C
BAP/BSRC/STR/BV-45-C
BAP/BSRC/STR/BV-46-C
BAP/BSRC/STR/BV-47-C
BAP/BSRC/STR/BV-48-C
BAP/BSRC/STR/BV-49-C
BAP/BSRC/STR/BV-50-C

BAP/BSRC/STR/BV-18-C
BAP/BSRC/STR/BV-19-C
BAP/BSRC/STR/BV-20-C
BAP/BSRC/STR/BV-21-C
BAP/BSRC/STR/BV-22-C
BAP/BSRC/STR/BV-23-C
BAP/BSRC/STR/BV-24-C
BAP/BSRC/STR/BV-25-C
BAP/BSRC/STR/BV-26-C
BAP/BSRC/STR/BV-27-C
BAP/BSRC/STR/BV-28-C
BAP/BSRC/STR/BV-29-C
BAP/BSRC/STR/BV-30-C
BAP/BSRC/STR/BV-31-C
BAP/BSRC/STR/BV-32-C
BAP/BSRC/STR/BV-33-C

Testing
nRF5340 Audio DK
nRF54L15 DK

### DIFF
--- a/autopts/wid/bap.py
+++ b/autopts/wid/bap.py
@@ -73,6 +73,30 @@ class AudioConfigId(Enum):
     AC_10 = auto()
     AC_11_i = auto()
     AC_11_ii = auto()
+    AC_12 = auto()
+    AC_13 = auto()
+    AC_14 = auto()
+
+
+class BroadcastSetName(str, Enum):
+    """Broadcast set names defined by BAP."""
+
+    SET_8_1_1 = "8_1_1"
+    SET_8_2_1 = "8_2_1"
+    SET_16_1_1 = "16_1_1"
+    SET_16_2_1 = "16_2_1"
+    SET_24_1_1 = "24_1_1"
+    SET_24_2_1 = "24_2_1"
+    SET_32_1_1 = "32_1_1"
+    SET_32_2_1 = "32_2_1"
+    SET_441_1_1 = "441_1_1"
+    SET_441_2_1 = "441_2_1"
+    SET_48_1_1 = "48_1_1"
+    SET_48_2_1 = "48_2_1"
+    SET_48_3_1 = "48_3_1"
+    SET_48_4_1 = "48_4_1"
+    SET_48_5_1 = "48_5_1"
+    SET_48_6_1 = "48_6_1"
 
 
 @dataclass(frozen=True)
@@ -326,6 +350,152 @@ AC_CONFIGS: Final[list[AudioConfigTest]] = [
 
 # Dictionary representing the above list for easier and O(1) lookup
 AC_CONFIGS_DICT: Final[dict[str, AudioConfigTest]] = {cfg.test_case_str: cfg for cfg in AC_CONFIGS}
+
+
+BROADCAST_AUDIO_LOCATION_STEREO: Final[int] = (defs.PACS_AUDIO_LOCATION_FRONT_LEFT | defs.PACS_AUDIO_LOCATION_FRONT_RIGHT)
+BROADCAST_AUDIO_LOCATION_MONO: Final[int] = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+
+
+@dataclass(frozen=True)
+class BroadcastAudioConfigTest:
+    """Declarative broadcast STR test configuration.
+
+    Attributes:
+        test_case_str (str): PTS test case identifier, for example "BAP/BSRC/STR/BV-31-C".
+        set_name (BroadcastSetName): set name.
+        audio_config_id (AudioConfigId): BAP audio configuration identifier used by the testcase.
+        channel_allocation (int): PACS audio location bitfield used in codec configuration.
+        streams_per_subgroup (int): Number of BIS streams configured per subgroup.
+    """
+
+    test_case_str: str
+    set_name: BroadcastSetName
+    audio_config_id: AudioConfigId
+    channel_allocation: int = BROADCAST_AUDIO_LOCATION_MONO
+    streams_per_subgroup: int = 1
+
+
+BROADCAST_STR_CONFIGS: Final[tuple[BroadcastAudioConfigTest, ...]] = (
+    # AC12: 1 BIS, mono (BV-01..BV-16)
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-01-C", BroadcastSetName.SET_8_1_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-02-C", BroadcastSetName.SET_8_2_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-03-C", BroadcastSetName.SET_16_1_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-04-C", BroadcastSetName.SET_16_2_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-05-C", BroadcastSetName.SET_24_1_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-06-C", BroadcastSetName.SET_24_2_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-07-C", BroadcastSetName.SET_32_1_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-08-C", BroadcastSetName.SET_32_2_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-09-C", BroadcastSetName.SET_441_1_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-10-C", BroadcastSetName.SET_441_2_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-11-C", BroadcastSetName.SET_48_1_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-12-C", BroadcastSetName.SET_48_2_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-13-C", BroadcastSetName.SET_48_3_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-14-C", BroadcastSetName.SET_48_4_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-15-C", BroadcastSetName.SET_48_5_1, AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-16-C", BroadcastSetName.SET_48_6_1, AudioConfigId.AC_12),
+    # AC13: 2 BIS, stereo (BV-18..BV-33)
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-18-C", BroadcastSetName.SET_8_1_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-19-C", BroadcastSetName.SET_8_2_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-20-C", BroadcastSetName.SET_16_1_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-21-C", BroadcastSetName.SET_16_2_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-22-C", BroadcastSetName.SET_24_1_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-23-C", BroadcastSetName.SET_24_2_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-24-C", BroadcastSetName.SET_32_1_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-25-C", BroadcastSetName.SET_32_2_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-26-C", BroadcastSetName.SET_441_1_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-27-C", BroadcastSetName.SET_441_2_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-28-C", BroadcastSetName.SET_48_1_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-29-C", BroadcastSetName.SET_48_2_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-30-C", BroadcastSetName.SET_48_3_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-31-C", BroadcastSetName.SET_48_4_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-32-C", BroadcastSetName.SET_48_5_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-33-C", BroadcastSetName.SET_48_6_1, AudioConfigId.AC_13, BROADCAST_AUDIO_LOCATION_STEREO, 2
+    ),
+    # AC14: 1 BIS, stereo, max SDU scales by channel count (BV-35..BV-50)
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-35-C", BroadcastSetName.SET_8_1_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-36-C", BroadcastSetName.SET_8_2_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-37-C", BroadcastSetName.SET_16_1_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-38-C", BroadcastSetName.SET_16_2_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-39-C", BroadcastSetName.SET_24_1_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-40-C", BroadcastSetName.SET_24_2_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-41-C", BroadcastSetName.SET_32_1_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-42-C", BroadcastSetName.SET_32_2_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-43-C", BroadcastSetName.SET_441_1_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-44-C", BroadcastSetName.SET_441_2_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-45-C", BroadcastSetName.SET_48_1_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-46-C", BroadcastSetName.SET_48_2_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-47-C", BroadcastSetName.SET_48_3_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-48-C", BroadcastSetName.SET_48_4_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-49-C", BroadcastSetName.SET_48_5_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-50-C", BroadcastSetName.SET_48_6_1, AudioConfigId.AC_14, BROADCAST_AUDIO_LOCATION_STEREO, 1
+    ),
+)
+
+BROADCAST_STR_CONFIGS_DICT: Final[dict[str, BroadcastAudioConfigTest]] = {
+    cfg.test_case_str: cfg for cfg in BROADCAST_STR_CONFIGS
+}
 
 
 def ascs_config_codec(addr_type: int, addr: str, ase_id: int, coding_format: int,
@@ -664,11 +834,32 @@ BAS_CONFIG_SETTINGS = {
 }
 
 
+@dataclass(frozen=True)
+class BasConfig:
+    """Broadcast Audio Stream QoS/codec mapping.
+
+    Attributes:
+        set_name (str): Key into CODEC_CONFIG_SETTINGS.
+        sdu_interval_us (int): SDU interval in microseconds.
+        framing (int): Framing mode used by BAP (0 unframed, 1 framed)
+        max_sdu_octets (int): Maximum SDU size in octets for a single stream.
+        rtn (int): Retransmission number.
+        max_transport_latency_ms (int): Maximum transport latency in milliseconds.
+    """
+
+    set_name: str
+    sdu_interval_us: int
+    framing: int
+    max_sdu_octets: int
+    rtn: int
+    max_transport_latency_ms: int
+
+
 def hdl_wid_114(params: WIDParams):
     """
     Please advertise with Broadcast Audio Announcement (0x1852) service data
     """
-    configurations = {
+    scc_configurations = {
         'BAP/BSRC/SCC/BV-01-C': '8_1_1',
         'BAP/BSRC/SCC/BV-02-C': '8_2_1',
         'BAP/BSRC/SCC/BV-03-C': '16_1_1',
@@ -702,43 +893,11 @@ def hdl_wid_114(params: WIDParams):
         'BAP/BSRC/SCC/BV-31-C': '48_5_2',
         'BAP/BSRC/SCC/BV-32-C': '48_6_2',
         'BAP/BSRC/SCC/BV-38-C': '16_2_1',
-        # Cases with 1 BIS:
-        'BAP/BSRC/STR/BV-01-C': '8_1_1',
-        'BAP/BSRC/STR/BV-02-C': '8_2_1',
-        'BAP/BSRC/STR/BV-03-C': '16_1_1',
-        'BAP/BSRC/STR/BV-04-C': '16_2_1',
-        'BAP/BSRC/STR/BV-05-C': '24_1_1',
-        'BAP/BSRC/STR/BV-06-C': '24_2_1',
-        'BAP/BSRC/STR/BV-07-C': '32_1_1',
-        'BAP/BSRC/STR/BV-08-C': '32_2_1',
-        'BAP/BSRC/STR/BV-09-C': '441_1_1',
-        'BAP/BSRC/STR/BV-10-C': '441_2_1',
-        'BAP/BSRC/STR/BV-11-C': '48_1_1',
-        'BAP/BSRC/STR/BV-12-C': '48_2_1',
-        'BAP/BSRC/STR/BV-13-C': '48_3_1',
-        'BAP/BSRC/STR/BV-14-C': '48_4_1',
-        'BAP/BSRC/STR/BV-15-C': '48_5_1',
-        'BAP/BSRC/STR/BV-16-C': '48_6_1',
-        # Cases with 2 BISes:
-        'BAP/BSRC/STR/BV-18-C': '8_1_1',
-        'BAP/BSRC/STR/BV-19-C': '8_2_1',
-        'BAP/BSRC/STR/BV-20-C': '16_1_1',
-        'BAP/BSRC/STR/BV-21-C': '16_2_1',
-        'BAP/BSRC/STR/BV-22-C': '24_1_1',
-        'BAP/BSRC/STR/BV-23-C': '24_2_1',
-        'BAP/BSRC/STR/BV-24-C': '32_1_1',
-        'BAP/BSRC/STR/BV-25-C': '32_2_1',
-        'BAP/BSRC/STR/BV-26-C': '441_1_1',
-        'BAP/BSRC/STR/BV-27-C': '441_2_1',
-        'BAP/BSRC/STR/BV-28-C': '48_1_1',
-        'BAP/BSRC/STR/BV-29-C': '48_2_1',
-        'BAP/BSRC/STR/BV-30-C': '48_3_1',
-        'BAP/BSRC/STR/BV-31-C': '48_4_1',
-        'BAP/BSRC/STR/BV-32-C': '48_5_1',
-        'BAP/BSRC/STR/BV-33-C': '48_6_1',
     }
 
     stack = get_stack()
+
+    is_broadcast_str_test = params.test_case_name.startswith('BAP/BSRC/STR')
 
     if stack.bap.hdl_wid_114_cnt == 0:
         broadcast_id = stack.bap.broadcast_id
@@ -747,8 +906,34 @@ def hdl_wid_114(params: WIDParams):
     else:
         raise ValueError("hdl_wid_114 is not 0 or 1")
 
-    if params.test_case_name in configurations:
-        qos_set_name = configurations[params.test_case_name]
+    channel_allocation = BROADCAST_AUDIO_LOCATION_MONO
+    streams_per_subgroup = 1
+
+    if is_broadcast_str_test:
+        broadcast_str_cfg = BROADCAST_STR_CONFIGS_DICT.get(params.test_case_name)
+        if broadcast_str_cfg is None:
+            raise ValueError(f"Unsupported STR testcase config: {params.test_case_name}")
+
+        set_name = broadcast_str_cfg.set_name
+        channel_allocation = broadcast_str_cfg.channel_allocation
+        streams_per_subgroup = broadcast_str_cfg.streams_per_subgroup
+        coding_format = 0x06
+        vid = 0x0000
+        cid = 0x0000
+
+        channel_count = count_1_bits(channel_allocation)
+        if streams_per_subgroup <= 0:
+            raise ValueError(f"Invalid streams_per_subgroup={streams_per_subgroup}")
+        if channel_count == 0:
+            raise ValueError(f"Invalid channel_allocation=0x{channel_allocation:X}")
+        if channel_count % streams_per_subgroup != 0:
+            raise ValueError(
+                "Invalid broadcast STR config: channel_count "
+                f"{channel_count} is not divisible by streams_per_subgroup {streams_per_subgroup}"
+            )
+        channel_ratio = channel_count // streams_per_subgroup
+    elif params.test_case_name in scc_configurations:
+        qos_set_name = scc_configurations[params.test_case_name]
         coding_format = 0x06
         vid = 0x0000
         cid = 0x0000
@@ -758,21 +943,22 @@ def hdl_wid_114(params: WIDParams):
         vid = 0xffff
         cid = 0xffff
 
-    codec_set_name, *qos_config = BAS_CONFIG_SETTINGS[qos_set_name]
+    if is_broadcast_str_test:
+        qos_set_name = set_name
+
+    bas_config = BasConfig(*BAS_CONFIG_SETTINGS[qos_set_name])
+    max_sdu_octets = bas_config.max_sdu_octets
+
+    if is_broadcast_str_test:
+        max_sdu_octets *= int(channel_ratio)
 
     (sampling_freq, frame_duration, octets_per_frame) = \
-        CODEC_CONFIG_SETTINGS[codec_set_name]
-    audio_locations = 0x01
+        CODEC_CONFIG_SETTINGS[bas_config.set_name]
     frames_per_sdu = 0x01
 
     codec_ltvs_bytes = create_lc3_ltvs_bytes(sampling_freq, frame_duration,
-                                             audio_locations, octets_per_frame,
+                                             channel_allocation, octets_per_frame,
                                              frames_per_sdu)
-
-    streams_per_subgroup = 1
-    tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-    if tc_num >= 18:
-        streams_per_subgroup = 2
 
     presentation_delay = 40000
     subgroups = 1
@@ -780,11 +966,21 @@ def hdl_wid_114(params: WIDParams):
     if params.test_case_name.startswith("BAP/BSRC/SCC/BV-38-C"):
         btp.bap_broadcast_source_setup_v2(broadcast_id, streams_per_subgroup, subgroups,
                                           coding_format, vid, cid, codec_ltvs_bytes,
-                                          *qos_config, presentation_delay)
+                                          bas_config.sdu_interval_us,
+                                          bas_config.framing,
+                                          max_sdu_octets,
+                                          bas_config.rtn,
+                                          bas_config.max_transport_latency_ms,
+                                          presentation_delay)
     else:
         broadcast_id = btp.bap_broadcast_source_setup(streams_per_subgroup, subgroups,
                                                       coding_format, vid, cid, codec_ltvs_bytes,
-                                                      *qos_config, presentation_delay)
+                                                      bas_config.sdu_interval_us,
+                                                      bas_config.framing,
+                                                      max_sdu_octets,
+                                                      bas_config.rtn,
+                                                      bas_config.max_transport_latency_ms,
+                                                      presentation_delay)
         stack.bap.set_broadcast_id(broadcast_id)
 
     btp.bap_broadcast_adv_start(broadcast_id)
--- a/test/unittests.py
+++ b/test/unittests.py
@@ -395,6 +395,69 @@ class MyTestCase(unittest.TestCase):
         except Exception as e:
             self.fail(f"Function raised unexpected exception: {e}")
 
+    def test_broadcast_str_ac12_config_mapping(self):
+        from autopts.wid.bap import BROADCAST_STR_CONFIGS, AudioConfigId, BroadcastSetName
+
+        configs = [cfg for cfg in BROADCAST_STR_CONFIGS if cfg.audio_config_id is AudioConfigId.AC_12]
+
+        assert len(configs) == 16
+        assert [cfg.set_name for cfg in configs] == list(BroadcastSetName)
+        assert [cfg.test_case_str for cfg in configs] == [
+            f"BAP/BSRC/STR/BV-{idx:02d}-C" for idx in range(1, 17)
+        ]
+
+        for cfg in configs:
+            assert cfg.channel_allocation == defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+            assert cfg.streams_per_subgroup == 1
+            bits = defs.PACS_AUDIO_LOCATION_FRONT_LEFT.bit_count()
+            assert bits % cfg.streams_per_subgroup == 0
+
+    def test_broadcast_str_ac13_config_mapping(self):
+        from autopts.wid.bap import (
+            BROADCAST_AUDIO_LOCATION_STEREO,
+            BROADCAST_STR_CONFIGS,
+            AudioConfigId,
+            BroadcastSetName,
+        )
+
+        configs = [cfg for cfg in BROADCAST_STR_CONFIGS if cfg.audio_config_id is AudioConfigId.AC_13]
+        expected_qos_set_names = list(BroadcastSetName)
+
+        assert len(configs) == len(expected_qos_set_names)
+        assert [cfg.set_name for cfg in configs] == expected_qos_set_names
+        assert [cfg.test_case_str for cfg in configs] == [
+            f"BAP/BSRC/STR/BV-{idx:02d}-C" for idx in range(18, 34)
+        ]
+
+        for cfg in configs:
+            assert cfg.channel_allocation == BROADCAST_AUDIO_LOCATION_STEREO
+            assert cfg.streams_per_subgroup == 2
+            bits = cfg.channel_allocation.bit_count()
+            assert bits % cfg.streams_per_subgroup == 0
+
+    def test_broadcast_str_ac14_config_mapping(self):
+        from autopts.wid.bap import (
+            BROADCAST_AUDIO_LOCATION_STEREO,
+            BROADCAST_STR_CONFIGS,
+            AudioConfigId,
+            BroadcastSetName,
+        )
+
+        configs = [cfg for cfg in BROADCAST_STR_CONFIGS if cfg.audio_config_id is AudioConfigId.AC_14]
+        expected_qos_set_names = list(BroadcastSetName)
+
+        assert len(configs) == len(expected_qos_set_names)
+        assert [cfg.set_name for cfg in configs] == expected_qos_set_names
+        assert [cfg.test_case_str for cfg in configs] == [
+            f"BAP/BSRC/STR/BV-{idx:02d}-C" for idx in range(35, 51)
+        ]
+
+        for cfg in configs:
+            assert cfg.channel_allocation == BROADCAST_AUDIO_LOCATION_STEREO
+            assert cfg.streams_per_subgroup == 1
+            bits = cfg.channel_allocation.bit_count()
+            assert bits % cfg.streams_per_subgroup == 0
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/autopts/ptsprojects/zephyr/bap.py
+++ b/autopts/ptsprojects/zephyr/bap.py
@@ -54,6 +54,7 @@ def set_pixits(ptses):
     pts.set_pixit("BAP", "TSPX_iut_ATT_transport", "ATT Bearer on LE Transport")
     pts.set_pixit("BAP", "TSPX_VS_Codec_ID", "ffff")
     pts.set_pixit("BAP", "TSPX_VS_Company_ID", "ffff")
+    pts.set_pixit("BAP", "TSPX_STREAMING_DATA_CONFIRMATION_METHOD", "By Playing")
     pts.set_pixit("BAP", "TSPX_broadcast_code", BROADCAST_CODE)
     """
     TODO this should be set to -1 but PTS is return error when setting this with RPC so leave
@@ -80,6 +81,7 @@ def set_pixits(ptses):
     pts2.set_pixit("BAP", "TSPX_iut_ATT_transport", "ATT Bearer on LE Transport")
     pts2.set_pixit("BAP", "TSPX_VS_Codec_ID", "ffff")
     pts2.set_pixit("BAP", "TSPX_VS_Company_ID", "ffff")
+    pts2.set_pixit("BAP", "TSPX_STREAMING_DATA_CONFIRMATION_METHOD", "By Playing")
     pts2.set_pixit("BAP", "TSPX_broadcast_code", BROADCAST_CODE)
 
 


autopts/wid/bap.py
```diff
@@ -763,18 +780,22 @@ def hdl_wid_114(params: WIDParams):
     (sampling_freq, frame_duration, octets_per_frame) = \
         CODEC_CONFIG_SETTINGS[codec_set_name]
     audio_locations = 0x01
-    frames_per_sdu = 0x01
+    is_str_ac14 = re.match(r'^BAP/BSRC/STR/BV-(3[5-9]|4[0-9]|50)-C$', params.test_case_name)
+    if is_str_ac14:
+        # AC14 uses one BIS carrying FL|FR channels.
+        audio_locations = 0x00000003
+    # AC14 (single BIS, stereo) should advertise two codec frames per SDU.
+    frames_per_sdu = 0x02 if is_str_ac14 else 0x01
```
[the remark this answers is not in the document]
ciascai: We intentionally keep QoS (including MAX SDU) from the selected BAS profile (8_1_1, 8_2_1, etc), because those profile values are what is expected. 

`frames_per_sdu = 2` here is used to represent de AC14 stereo framing semantics for a single BIS, while the QoS profile remains unchanged by design. With this exact combination, BV-35 through BV-50 now pass end-to-end (including channel-count validation), so I'd prefer not to change the MAX SDU

ciascai:
https://github.com/zephyrproject-rtos/zephyr/pull/112553

autopts/wid/bap.py
```diff
@@ -736,6 +736,23 @@ def hdl_wid_114(params: WIDParams):
         'BAP/BSRC/STR/BV-31-C': '48_4_1',
         'BAP/BSRC/STR/BV-32-C': '48_5_1',
         'BAP/BSRC/STR/BV-33-C': '48_6_1',
+        # Cases with 1 BIS and 2 channels (AC14):
```
Thalley: Did you have a look at https://github.com/auto-pts/auto-pts/blob/master/autopts/wid/bap.py#L55-L328 and see how we handle this for unicast? I'm wondering if we can apply/extend those to also cover the broadcast audio configurations, but maybe it's too much when there are only 3 configs for broadcast?

autopts/wid/bap.py
```diff
@@ -763,18 +780,22 @@ def hdl_wid_114(params: WIDParams):
     (sampling_freq, frame_duration, octets_per_frame) = \
         CODEC_CONFIG_SETTINGS[codec_set_name]
     audio_locations = 0x01
-    frames_per_sdu = 0x01
+    is_str_ac14 = re.match(r'^BAP/BSRC/STR/BV-(3[5-9]|4[0-9]|50)-C$', params.test_case_name)
+    if is_str_ac14:
+        # AC14 uses one BIS carrying FL|FR channels.
+        audio_locations = 0x00000003
```
Thalley: We could consider adding the channel allocation to the `configurations` dictionary.

autopts/wid/bap.py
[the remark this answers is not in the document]
Thalley: This is actually the opposite. Broadcast Audio Configuration uses a single frame per SDU per channel. The naming here is a bit off, but the actual definition of this value is `Codec_Frame_Blocks_Per_SDU` and indicates how many frames per channel per SDU there is. 

For the AC_14 tests, we do need to double the SDU to have 2 channels. 

The SDU size is in fact number of bits in `Audio_Channel_Allocation` * `Octets_Per_Codec_Frame` *  `Codec_Frame_Blocks_Per_SDU`.

The BAP configs define the `Octets_Per_Codec_Frame` (although they have typed it as `Octets per_Codec_Frame` ...) and by default Codec_Frame_Blocks_Per_SDU is 1, so to set the BAP config SDU size we just need to multiple the `Octets_Per_Codec_Frame` (e.g. 40 for 16_2) by number of channels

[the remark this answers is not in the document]
Thalley: So this line should just remain unchanged, but we do need to configure the right SDU size elsewhere

autopts/wid/bap.py
```diff
@@ -763,18 +780,22 @@ def hdl_wid_114(params: WIDParams):
...
                                              frames_per_sdu)
 
     streams_per_subgroup = 1
-    tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-    if tc_num >= 18:
+    if re.match(r'^BAP/BSRC/STR/BV-(1[8-9]|2[0-9]|3[0-3])-C$', params.test_case_name):
         streams_per_subgroup = 2
 
-    presentation_delay = 40000
+    presentation_delay = 10000 if is_str_ac14 else 40000
```
Thalley: AC_14 does not define a different presentation delay

autopts/wid/bap.py
```diff
@@ -763,18 +780,22 @@ def hdl_wid_114(params: WIDParams):
...
+    frames_per_sdu = 0x02 if is_str_ac14 else 0x01
 
     codec_ltvs_bytes = create_lc3_ltvs_bytes(sampling_freq, frame_duration,
                                              audio_locations, octets_per_frame,
                                              frames_per_sdu)
 
     streams_per_subgroup = 1
-    tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-    if tc_num >= 18:
+    if re.match(r'^BAP/BSRC/STR/BV-(1[8-9]|2[0-9]|3[0-3])-C$', params.test_case_name):
```
Thalley: The above suggestions to better defined the audio configurations would help make this easier. From a glance it's nearly impossible to conclude that this regex means "Audio configuration 13"

Thalley (changes requested):

autopts/wid/bap.py
ciascai: I did look at the unicast `AudioConfig` helpers, and I agree the general idea is cleaner. For this change I kept the broadcast handling lighter because the broadcast STR cases only need a small amount of classification (18..33 for 2 BIS and 35..50 for AC14), so pulling in the full unicast model felt like more structure than we need right now.

autopts/wid/bap.py
```diff
@@ -736,6 +736,23 @@ def hdl_wid_114(params: WIDParams):
...
+        'BAP/BSRC/STR/BV-41-C': ('32_1_1', 0x00000003),
+        'BAP/BSRC/STR/BV-42-C': ('32_2_1', 0x00000003),
+        'BAP/BSRC/STR/BV-43-C': ('441_1_1', 0x00000003),
+        'BAP/BSRC/STR/BV-44-C': ('441_2_1', 0x00000003),
+        'BAP/BSRC/STR/BV-45-C': ('48_1_1', 0x00000003),
+        'BAP/BSRC/STR/BV-46-C': ('48_2_1', 0x00000003),
+        'BAP/BSRC/STR/BV-47-C': ('48_3_1', 0x00000003),
+        'BAP/BSRC/STR/BV-48-C': ('48_4_1', 0x00000003),
+        'BAP/BSRC/STR/BV-49-C': ('48_5_1', 0x00000003),
+        'BAP/BSRC/STR/BV-50-C': ('48_6_1', 0x00000003),
```
Thalley: If applying the channel allocation like this, you should do it for all the items in the dictionary. AC_13 will probably also use LEFT + RIGHT, and AC_12 can use whatever (or mono/0). 

Please also use the already defined channel allocation values like `PACS_AUDIO_LOCATION_FRONT_LEFT` (even though they should be defined as a non-PACS value...).

autopts/wid/bap.py
```diff
@@ -747,8 +764,13 @@ def hdl_wid_114(params: WIDParams):
     else:
         raise ValueError("hdl_wid_114 is not 0 or 1")
 
+    audio_locations = 0x01
```
Thalley: Please use the terminology channel allocation instead of audio location

autopts/wid/bap.py
```diff
@@ -760,18 +782,24 @@ def hdl_wid_114(params: WIDParams):
...
+    tc_match = re.search(r'BV-(\d+)-C$', params.test_case_name)
+    tc_num = int(tc_match.group(1)) if (is_bsrc_str and tc_match) else -1
+
     (sampling_freq, frame_duration, octets_per_frame) = \
         CODEC_CONFIG_SETTINGS[codec_set_name]
-    audio_locations = 0x01
+    is_str_ac14 = is_bsrc_str and 35 <= tc_num <= 50
     frames_per_sdu = 0x01
+    if is_str_ac14:
+        qos_config[2] *= count_1_bits(audio_locations)
```
Thalley: It still contains this messy and documented format where we kind of just implicitly apply changes to values based on the test case name, instead of the channel allocation. 

I think it would be better to do this in a generic approach, where we define a generic way of setting the values based on some defined values per test case. 

This _can_ work, but please do document it properly

autopts/wid/bap.py
```diff
@@ -760,18 +782,24 @@ def hdl_wid_114(params: WIDParams):
...
 
     codec_ltvs_bytes = create_lc3_ltvs_bytes(sampling_freq, frame_duration,
                                              audio_locations, octets_per_frame,
                                              frames_per_sdu)
 
     streams_per_subgroup = 1
-    tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-    if tc_num >= 18:
+    is_two_bis_str = is_bsrc_str and 18 <= tc_num <= 33
+    if is_two_bis_str:
```
Thalley: Please document the rationale behind this

autopts/wid/bap.py
```diff
@@ -747,8 +764,13 @@ def hdl_wid_114(params: WIDParams):
...
         raise ValueError("hdl_wid_114 is not 0 or 1")
 
+    audio_locations = 0x01
     if params.test_case_name in configurations:
-        qos_set_name = configurations[params.test_case_name]
+        configuration = configurations[params.test_case_name]
+        if isinstance(configuration, tuple):
+            qos_set_name, audio_locations = configuration
+        else:
+            qos_set_name = configuration
```
Thalley: If you apply my suggestion above, this would just be `qos_set_name, audio_locations = configurations[params.test_case_name]`

Thalley (changes requested):
Consider if we need something like `cap_broadcast_source_setup_stream` but for BAP. That way we can here in BTP/AutoPTS control the channel allocation of the BIS and other useful things in a IUT-generic way. From your other PR it's not strictly needed as we can infer the BIS codec config data based on the channel allocation

autopts/wid/bap.py
```diff
@@ -747,6 +764,30 @@ def hdl_wid_114(params: WIDParams):
...
+    tc_match = re.search(r'BV-(\d+)-C$', params.test_case_name)
+    tc_num = int(tc_match.group(1)) if (is_bsrc_str and tc_match) else -1
+
+    # BAP/BSRC/STR test numbering follows the PTS audio-config groups:
+    # BV-18..33 are AC13 (2 BIS) and BV-35..50 are AC14 (1 BIS, 2 channels).
+    # We use these ranges to select stereo channel allocation consistently.
+    is_str_ac13 = is_bsrc_str and 18 <= tc_num <= 33
+    is_str_ac14 = is_bsrc_str and 35 <= tc_num <= 50
+    if is_str_ac13 or is_str_ac14:
+        audio_locations = stereo_audio_locations
```
Thalley: One alternative here is to also in the `hdl_wid_114` define e.g. ` 'BAP/BSRC/STR/BV-35-C': ('8_1_1', defs.PACS_AUDIO_LOCATION_FRONT_LEFT | defs.PACS_AUDIO_LOCATION_FRONT_RIGHT, AudioConfigId.AC_14)`

It's effectively the same, but I am honestly not a big fan of implicitly extracting the AC and channel allocation based on the number in the test. Right now if a `BAP/BSRC/STR/BV-51-C` is being added, it's implicitly AC_12 with mono. 

Can you share your reasons for the the rationale of not explicitly declaring the AC and channel allocation in a declarative manner (i.e. in the `dict`), compared to relying on the number in the test case name? Maybe I'm missing something obvious

autopts/wid/bap.py
```diff
@@ -762,16 +803,19 @@ def hdl_wid_114(params: WIDParams):
 
     (sampling_freq, frame_duration, octets_per_frame) = \
         CODEC_CONFIG_SETTINGS[codec_set_name]
-    audio_locations = 0x01
     frames_per_sdu = 0x01
+    if is_str_ac14:
+        qos_config[2] *= count_1_bits(audio_locations)
```
Thalley: Please add a comment explaining this

autopts/wid/bap.py
```diff
@@ -791,11 +835,6 @@ def hdl_wid_114(params: WIDParams):
 
     btp.bap_broadcast_source_start(broadcast_id)
 
-    data = bytearray([j for j in range(0, 41)])
-
-    for _ in range(1, 100):
-        _safe_bap_send(0, data)
```
Thalley, line 797: Why was this removed?

autopts/wid/bap.py
```diff
@@ -747,6 +764,30 @@ def hdl_wid_114(params: WIDParams):
...
 
+    # Use PACS-defined location bits instead of raw literals to keep channel
+    # allocation explicit and consistent with the rest of the stack.
+    # Default is mono (front-left); specific STR AC ranges below override to
+    # stereo by combining front-left and front-right.
+    mono_audio_locations = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+    stereo_audio_locations = (
+        defs.PACS_AUDIO_LOCATION_FRONT_LEFT | defs.PACS_AUDIO_LOCATION_FRONT_RIGHT
+    )
+    audio_locations = mono_audio_locations
```
Thalley: As previously requested, please use the terminology `chan_alloc`/`channel_allocation` instead, as that is what this value reflects

autopts/wid/bap.py
```diff
@@ -747,6 +764,30 @@ def hdl_wid_114(params: WIDParams):
...
+
+    # Extract BV number once from the testcase name so we can map STR cases to
+    # their AC group behavior (channel allocation, BIS count, and QoS tweaks).
+    is_bsrc_str = params.test_case_name.startswith("BAP/BSRC/STR/")
+    tc_match = re.search(r'BV-(\d+)-C$', params.test_case_name)
+    tc_num = int(tc_match.group(1)) if (is_bsrc_str and tc_match) else -1
```
Thalley: If `tc_num == -1` then we just silently continue?

Thalley (changes requested):

sjanc:
and there is typo in commit message :)  (copilot is not checking commit messages?:)

autopts/wid/bap.py
ciascai: `tc_num == -1` is intentional as a sentinel for non-matching STR testcases. It makes the AC13/AC14 range checks evaluate false, so we silently skip the special channel-allocation logic for anything outside `BAP/BSRC/STR/BV-`

autopts/wid/bap.py
ciascai: Your point is fair. The number-based split was mainly a shortcut to avoid duplicating a large table while I was aligning the STR cases with the existing BV group structure. I don't have a strong reason to prefer it over the declarative dict version. The clearer approach is to encode the testcase to AC/channel-allocation mapping directly in `hdl_wid_114`, including the AC id, so the intent is obvious and adding something like `BV-51-C` won't silently inherit the wrong default behavior.

autopts/wid/bap.py
ciascai: I removed that hardcoded burst because it was unconditional and always targeted ASE 0, which is too brittle for this flow. The current code sends ISO data only to ASEs that are actually in STREAMING, so it stays testcase-driven instead of assuming a fixed ASE ID and pushing a blind repeat loop.

Also it was a workaround to keep the sink fed with ISO data while the testcase flow was still fragile. After the Zephyr-side changes, that extra burst is no longer needed

autopts/wid/bap.py
```diff
@@ -762,16 +805,21 @@ def hdl_wid_114(params: WIDParams):
 
     (sampling_freq, frame_duration, octets_per_frame) = \
         CODEC_CONFIG_SETTINGS[codec_set_name]
-    audio_locations = 0x01
     frames_per_sdu = 0x01
+    if is_str_ac14:
+        # AC14 uses stereo channel_allocation, so scale the QoS max SDU by the
+        # number of allocated channels.
+        qos_config[2] *= count_1_bits(channel_allocation)
```
Thalley: Lines like this is exactly why I want to refactor it to use some proper classes, instead of accessing tuples by a magic number index :(

autopts/wid/bap.py
Thalley: So `tc_num == -1` mean either AC12 or unknown. That doesn't seem like a good design. 

I _really_ think we should refactor the above dictionary to contain the AC. Not only does that add some very explicit information that makes it easier to debug, it also works a lot better if additional tests are added. E.g. If 3 new tests are added, 

'BAP/BSRC/STR/BV-51-C'
'BAP/BSRC/STR/BV-52-C'
'BAP/BSRC/STR/BV-53-C'

Where BV-51 is AC12, BV-52 is AC13 and BV-53 is AC14, then none of the assumptions nor logic below works, as that has a strong dependency that AC12 tests always have a lower TC number than AC13, and similar for AC13 and 14.

autopts/wid/bap.py
Thalley: Not sure I understand this. This is broadcast => There are no ASEs (ASEs are a unicast concept). 

I agree that this won't work for AC13 since it has 2 BIS and we only request sending to BIS_0 here, so something should be changed for TX. 

Whatever Zephyr may do, does not really come into play here, since this is not a Zephyr specific file, and this may break testing for other non-Zephyr platforms.

autopts/wid/bap.py
Thalley: I think the amount of regexes and checks below are far heavier than just adding the audio configuration here by now.

Thalley (changes requested):

autopts/wid/bap.py
ciascai: done

autopts/wid/bap.py
```diff
@@ -328,6 +331,205 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+@dataclass(frozen=True)
+class BroadcastAudioConfigTest:
+    """Declarative broadcast STR test configuration."""
+
+    test_case_str: str
+    qos_set_name: str
+    audio_config_id: AudioConfigId
+    channel_allocation: int = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+    streams_per_subgroup: int = 1
+    scale_max_sdu_by_channel_count: bool = False
```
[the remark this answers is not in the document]
ciascai: This is resolved

autopts/wid/bap.py
```diff
@@ -758,43 +988,62 @@ def hdl_wid_114(params: WIDParams):
...
+                                                      bas_config.framing,
+                                                      max_sdu_octets,
+                                                      bas_config.rtn,
+                                                      bas_config.max_transport_latency_ms,
+                                                      presentation_delay)
         stack.bap.set_broadcast_id(broadcast_id)
 
+    btp.bap_broadcast_source_start(broadcast_id)
+
     btp.bap_broadcast_adv_start(broadcast_id)
```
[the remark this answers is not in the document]
ciascai: resolved

autopts/wid/bap.py
```diff
@@ -747,6 +976,17 @@ def hdl_wid_114(params: WIDParams):
...
+    mono_channel_allocation = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+    channel_allocation = mono_channel_allocation
+    streams_per_subgroup = 1
+    scale_max_sdu_by_channel_count = False
+
+    broadcast_str_cfg = BROADCAST_STR_CONFIGS_DICT.get(params.test_case_name)
+    if broadcast_str_cfg is not None:
+        channel_allocation = broadcast_str_cfg.channel_allocation
+        streams_per_subgroup = broadcast_str_cfg.streams_per_subgroup
+        scale_max_sdu_by_channel_count = broadcast_str_cfg.scale_max_sdu_by_channel_count
```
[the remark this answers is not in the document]
ciascai: resolved

autopts/wid/bap.py
```diff
@@ -328,6 +331,205 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+    streams_per_subgroup: int = 1
+    scale_max_sdu_by_channel_count: bool = False
+
+
+BROADCAST_STR_CONFIGS: Final[list[BroadcastAudioConfigTest]] = [
+    # AC12: 1 BIS, mono
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-01-C", "8_1_1", AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-02-C", "8_2_1", AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-03-C", "16_1_1", AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-04-C", "16_2_1", AudioConfigId.AC_12),
```
Thalley: Suggest to define `"8_1_1"`, etc. instead of duplicating literal strings several places, as that can be quite error prone. 

Not 100% sure on what the best approach for that is, but instead of having the string "8_1_1" appear here, for AC_13 and for AC_14, as well as in the `QOS_CONFIG_SETTINGS`, there should be some way avoid that. 

I see that for the unicast tests, the set name (e.g. "8_1_1") is part of the WID description. That's not the case for broadcast, and we need to define it per test like this?

autopts/wid/bap.py
```diff
@@ -328,6 +331,205 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-11-C", "48_1_1", AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-12-C", "48_2_1", AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-13-C", "48_3_1", AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-14-C", "48_4_1", AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-15-C", "48_5_1", AudioConfigId.AC_12),
+    BroadcastAudioConfigTest("BAP/BSRC/STR/BV-16-C", "48_6_1", AudioConfigId.AC_12),
+    # AC13: 2 BIS, stereo
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-18-C", "8_1_1", AudioConfigId.AC_13,
+        defs.PACS_AUDIO_LOCATION_FRONT_LEFT | defs.PACS_AUDIO_LOCATION_FRONT_RIGHT, 2
```
Thalley: Consider if you want to define `defs.PACS_AUDIO_LOCATION_FRONT_LEFT | defs.PACS_AUDIO_LOCATION_FRONT_RIGHT` here locally as `BAP_LOCATION_STEREO = defs.PACS_AUDIO_LOCATION_FRONT_LEFT | defs.PACS_AUDIO_LOCATION_FRONT_RIGHT` or something to slightly reduce this duplication :)

autopts/wid/bap.py
Thalley: Do we need `scale_max_sdu_by_channel_count`? Wouldn't it always be `sdu *= streams_per_subgroup / count_1_bits(channel_allocation)`? (assuming that `channel_allocation` is always non-0 and we only have a single subgroup and that the number of channels is a multiple of the number of streams)

autopts/wid/bap.py
```diff
@@ -758,43 +991,62 @@ def hdl_wid_114(params: WIDParams):
...
 
-    for _ in range(1, 100):
-        _safe_bap_send(0, data)
+            for _ in range(1, 100):
+                for bis_id in stack.gap.big_bis_data_path_setup:
+                    _safe_gap_bis_broadcast(bis_id, '00')
+        else:
+            data = bytearray([j for j in range(0, 41)])
+            for _ in range(1, 100):
+                _safe_bap_send(0, data)
```
Thalley: Do we really need this change? The AC_13 tests, which are the ones affected by this, is passing without this change already? 

I agree that the code is not reflecting the audio configuration, but it seems like a separate issue to fix (if at all: https://github.com/auto-pts/auto-pts/issues/1317)

autopts/wid/bap.py
```diff
@@ -758,43 +991,62 @@ def hdl_wid_114(params: WIDParams):
...
+                                                      bas_config.framing,
+                                                      max_sdu_octets,
+                                                      bas_config.rtn,
+                                                      bas_config.max_transport_latency_ms,
+                                                      presentation_delay)
         stack.bap.set_broadcast_id(broadcast_id)
 
+    btp.bap_broadcast_source_start(broadcast_id)
+
     btp.bap_broadcast_adv_start(broadcast_id)
```
Thalley: Why the change in order between `bap_broadcast_source_start` and `bap_broadcast_adv_start`?

autopts/wid/bap.py
```diff
@@ -758,43 +991,62 @@ def hdl_wid_114(params: WIDParams):
...
-    codec_set_name, *qos_config = BAS_CONFIG_SETTINGS[qos_set_name]
+    bas_config = BasConfig(*BAS_CONFIG_SETTINGS[qos_set_name])
 
     (sampling_freq, frame_duration, octets_per_frame) = \
-        CODEC_CONFIG_SETTINGS[codec_set_name]
-    audio_locations = 0x01
+        CODEC_CONFIG_SETTINGS[bas_config.codec_set_name]
     frames_per_sdu = 0x01
+    max_sdu_octets = bas_config.max_sdu_octets
+    if scale_max_sdu_by_channel_count:
```
Thalley: Minor: We don't really need this local variable and could just use `broadcast_str_cfg.scale_max_sdu_by_channel_count`

autopts/wid/bap.py
```diff
@@ -702,52 +922,65 @@ def hdl_wid_114(params: WIDParams):
...
     else:
         raise ValueError("hdl_wid_114 is not 0 or 1")
 
-    if params.test_case_name in configurations:
+    channel_allocation = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+    streams_per_subgroup = 1
+    scale_max_sdu_by_channel_count = False
+
+    broadcast_str_cfg = BROADCAST_STR_CONFIGS_DICT.get(params.test_case_name)
+    if broadcast_str_cfg is not None:
```
Thalley: If `broadcast_str_cfg` is None, isn't that a unsupported test case then? We should fail loudly in that case, instead of defaulting to some other values. That would also significantly reduce the number of local variables here if we ensure that `broadcast_str_cfg` is always non-None later

autopts/wid/bap.py
```diff
@@ -702,52 +922,65 @@ def hdl_wid_114(params: WIDParams):
...
+            scale_max_sdu_by_channel_count,
+        )
+        if expected_cfg is not None and actual_cfg != expected_cfg:
+            logging.warning(
+                "Broadcast STR config mismatch for %s (%s): expected=%s actual=%s",
+                params.test_case_name,
+                broadcast_str_cfg.audio_config_id,
+                expected_cfg,
+                actual_cfg,
+            )
```
Thalley: This seems like it's some kind of embedded unit test in the middle of this function?

I'm all for adding unittests, but they should be added as proper unittests then :) We have a few already, so feel free to expand on those to add this verification

Thalley (changes requested):
Overall good, with some suggestions

autopts/wid/bap.py
ciascai: BV-46 and 2 other tests  fails before STR TX: PTS sees adv, sends Create Sync (OK), then times out waiting for periodic sync (~15.3s). So this is a PA sync timing issue. Keeping source_start before adv_start is intentional to avoid the adv-first race and keep BV-46 stable and 2 other tests

Thalley: Hmm, that's interesting. Please add that as a comment that this is a workaround for PTS, as either order should really work. 

`bap_broadcast_source_start` is not related to the PA at all (it relates to the BIG). The PA is started in `bap_broadcast_adv_start`. Did you mean the BIG sync fails? Otherwise this may be related to a scheduling issue on the IUT. 

Additionally, this is not an issue with most of the currently passing tests. Does it only affect a specific subset of tests?

autopts/wid/bap.py
ciascai: The streams/channels ratio idea is elegant, but it does not match all of our broadcast STR cases. In particular, AC13 and AC14 are both stereo in channel_allocation, but they intentionally do different things for SDU sizing:

1. AC13 (2 BIS, stereo): no extra SDU scaling
2. AC14 (1 BIS, stereo): SDU must be scaled by channel count

So a single generic formula would either be wrong for AC14 or hide behavior behind assumptions. Keeping this flag makes the testcase-specific intent explicit and safer.

autopts/wid/bap.py
ciascai: Agreed. I added a fail check for STR only. If a BAP/BSRC/STR testcase has no entry in BROADCAST_STR_CONFIGS_DICT, we now raise ValueError. Non-STR paths are intentionally unchanged.

Thalley: So this function is used for both STR and non-STR tests? :o If we don't do the lookup, how do we know if we are setting the right AC and QoS then, or are any values valid there since they are non-STR?

ciascai: Yes, this WID is shared by STR and non-STR. STR uses `BROADCAST_STR_CONFIGS_DICT`. Non-STR uses explicit per-test mappings/fallbacks in this function, so it does not rely on STR AC mappings by design.

autopts/wid/bap.py
Thalley: Sorry, it should have been `sdu *= count_1_bits(channel_allocation) / streams_per_subgroup` 

For
AC_12 = `sdu *= 1 / 1` = `sdu *= 1` 
AC_13 = `sdu *= 2 / 2` = `sdu *= 1` 
AC_14 = `sdu *= 2 / 1` = `sdu *= 2`

autopts/wid/bap.py
Thalley: I see. I guess all the SCC are implicitly AC-12. We could treat them in the same way to avoid having this split here. 

If we keep the split here, then we should ensure that we only allow `broadcast_str_cfg == None` if it's a SCC and not STR test, right?

autopts/wid/bap.py
Thalley: @ciascai The comment needs to be self-contained; right now it's leaning too much on the context of these Github comments. 

Instead of 
```
    # Workaround for observed PTS timing instability in a subset of BSRC/STR
    # cases (including BV-46): starting source before PA advertising reduces
    # periodic sync timeouts seen .
```

Suggest something like 
```
    # We have observed instabilities with PTS if we start advertising with EXT_ADV
    # and PER_ADV without the BIG in some tests. That instability seem to be fixed by
    # starting the BIG before the advertising. 
```

That being said, I think we should investigate this further. As mentioned, this does not seem to affect the current tests right, so it's something that happened with the other changes from this PR?

autopts/wid/bap.py
```diff
@@ -758,43 +847,65 @@ def hdl_wid_114(params: WIDParams):
...
+    btp.bap_broadcast_adv_start(broadcast_id)
+
+    if params.test_case_name.startswith('BAP/BSRC/STR'):
+        sleep(2)
 
-    for _ in range(1, 100):
-        _safe_bap_send(0, data)
+        data = bytearray([j for j in range(0, 41)])
+        for _ in range(1, 100):
+            _safe_bap_send(0, data)
```
Thalley: Can't we just revert all the lines to the previous version? 

And just FYI, Zephyr is not using `_safe_bap_send` at all, so this shouldn't affect your local testing at all :)

test/unittests.py
```diff
@@ -395,6 +395,64 @@ def test_audio_pack_metadata_program_info(self):
         except Exception as e:
             self.fail(f"Function raised unexpected exception: {e}")
 
+    def test_broadcast_str_ac12_config_shape(self):
```
Thalley: Just curios on the naming here: Is `shape` a specific term used for unittesting that I'm unaware of?

test/unittests.py
```diff
@@ -395,6 +395,64 @@ def test_audio_pack_metadata_program_info(self):
...
+        assert len(configs) == len(BROADCAST_QOS_SET_NAMES)
+        assert [cfg.qos_set_name for cfg in configs] == list(BROADCAST_QOS_SET_NAMES)
+        assert [cfg.test_case_str for cfg in configs] == [
+            f"BAP/BSRC/STR/BV-{idx:02d}-C" for idx in range(1, 17)
+        ]
+
+        for cfg in configs:
+            assert cfg.channel_allocation == defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+            assert cfg.streams_per_subgroup == 1
+            assert defs.PACS_AUDIO_LOCATION_FRONT_LEFT.bit_count() // cfg.streams_per_subgroup == 1
```
Thalley: Instead of using `//`, suggest to ensure that the number of bits is a multiple of `streams_per_subgroup`

autopts/wid/bap.py
```diff
@@ -758,43 +847,65 @@ def hdl_wid_114(params: WIDParams):
...
+        if streams_per_subgroup <= 0:
+            raise ValueError(f"Invalid streams_per_subgroup={streams_per_subgroup}")
+        if channel_count == 0:
+            raise ValueError(f"Invalid channel_allocation=0x{channel_allocation:X}")
+        if channel_count % streams_per_subgroup != 0:
+            raise ValueError(
+                "Invalid broadcast STR config: channel_count "
+                f"{channel_count} is not divisible by streams_per_subgroup {streams_per_subgroup}"
+            )
+        max_sdu_octets *= channel_count // streams_per_subgroup
```
Thalley: Suggest to do simply use `/`; if this is not an integer result then we should abort instead of silently flooring it 

```suggestion
        max_sdu_octets *= channel_count / streams_per_subgroup
```

autopts/wid/bap.py
```diff
@@ -328,6 +331,89 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+            BroadcastAudioConfigTest(
+                f"BAP/BSRC/STR/BV-{index:02d}-C",
+                qos_set_name,
+                AudioConfigId.AC_14,
+                BAP_LOCATION_STEREO,
+                1,
+            )
+        )
+
+    return configs
```
Thalley: This is going back to a previous issue I mentioned: If there are new tests added, then this solution scales poorly. I understand why you add this, but I think that the explicit list of test case mapping to location and AC makes it way easier to reverse lookup what is being used for a specific test case.

autopts/wid/bap.py
```diff
@@ -328,6 +331,89 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+    "32_2_1",
+    "441_1_1",
+    "441_2_1",
+    "48_1_1",
+    "48_2_1",
+    "48_3_1",
+    "48_4_1",
+    "48_5_1",
+    "48_6_1",
+)
```
Thalley: This is sort of a weird data type, isn't it? A long tuple of strings? :P

autopts/wid/bap.py
ciascai: Yes, agreed. `None` should only be allowed for SCC, not STR. I’ll keep the split but make that explicit in the guard.

autopts/wid/bap.py
ciascai: Yes, big sync fails. Commented this workaround

autopts/wid/bap.py
```diff
@@ -328,6 +331,89 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+
+
+@dataclass(frozen=True)
+class BroadcastAudioConfigTest:
+    """Declarative broadcast STR test configuration."""
+
+    test_case_str: str
+    qos_set_name: str
+    audio_config_id: AudioConfigId
+    channel_allocation: int = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
```
Thalley: Hmm, `PACS_AUDIO_LOCATION_PROHIBITED` should actually be `PACS_AUDIO_LOCATION_MONO`. I'm wondering if it makes more sense to use MONO when we only have a single stream as the default?

Thalley (changes requested):

autopts/wid/bap.py
ciascai: Yes, fair point. It’s just a compact lookup table, but I can make it more explicit if you want.

PS: it was your suggestion 👎

test/unittests.py
ciascai: No, it’s just shorthand for the tuple structure. I can rename it to something clearer.

autopts/wid/bap.py
Thalley: > PS: it was your suggestion 👎

I didn't suggest to use a tuple :D I didn't specify any specific data type, but an `StrEnum` would probably have been an ideal choice

autopts/wid/bap.py
```diff
@@ -758,39 +898,60 @@ def hdl_wid_114(params: WIDParams):
...
+            raise ValueError(f"Invalid streams_per_subgroup={streams_per_subgroup}")
+        if channel_count == 0:
+            raise ValueError(f"Invalid channel_allocation=0x{channel_allocation:X}")
+        channel_ratio = channel_count / streams_per_subgroup
+        if not channel_ratio.is_integer():
+            raise ValueError(
+                "Invalid broadcast STR config: channel_count "
+                f"{channel_count} is not divisible by streams_per_subgroup {streams_per_subgroup}"
+            )
+        max_sdu_octets = int(max_sdu_octets * channel_ratio)
```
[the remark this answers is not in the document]
ciascai: @Thalley you complained when I've done that :P
I think we can ignore this comment, thoughts?

autopts/wid/bap.py
```diff
@@ -328,6 +331,148 @@ def get_server(self, test_case: str) -> AudioConfigServer:
 AC_CONFIGS_DICT: Final[dict[str, AudioConfigTest]] = {cfg.test_case_str: cfg for cfg in AC_CONFIGS}
 
 
+BAP_LOCATION_STEREO: Final[int] = (
+    defs.PACS_AUDIO_LOCATION_FRONT_LEFT | defs.PACS_AUDIO_LOCATION_FRONT_RIGHT
+)
+BAP_LOCATION_MONO: Final[int] = getattr(
+    defs, "PACS_AUDIO_LOCATION_MONO", defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+)
```
Thalley: Why isn't this just

```suggestion
BAP_LOCATION_MONO: Final[int] = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
```

autopts/wid/bap.py
```diff
@@ -328,6 +331,148 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+
+
+@dataclass(frozen=True)
+class BroadcastAudioConfigTest:
+    """Declarative broadcast STR test configuration."""
```
Thalley: Please add documentation of the fields

autopts/wid/bap.py
Thalley: And given that we only use it 1 place, this (redefinition) does not serve that much of a purpose

autopts/wid/bap.py
```diff
@@ -328,6 +331,148 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-29-C", "48_2_1", AudioConfigId.AC_13, BAP_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-30-C", "48_3_1", AudioConfigId.AC_13, BAP_LOCATION_STEREO, 2
+    ),
+    BroadcastAudioConfigTest(
+        "BAP/BSRC/STR/BV-31-C", "48_4_1", AudioConfigId.AC_13, BAP_LOCATION_STEREO, 2
+    ),
```
Thalley, line 438: This project uses 127 line length; all of these seem to be able to fit on a single line, right?

autopts/wid/bap.py
```diff
@@ -664,6 +809,16 @@ def hdl_wid_104(params: WIDParams):
...
 
+@dataclass(frozen=True)
+class BasConfig:
+    codec_set_name: str
+    sdu_interval_us: int
+    framing: int
+    max_sdu_octets: int
+    rtn: int
+    max_transport_latency_ms: int
+
```
Thalley, line 856: Please document all fields

autopts/wid/bap.py
Thalley: Since you've now decouple the STR and SCC configurations, consider renaming this to make that more explicit that this is for SCC tests only

autopts/wid/bap.py
```diff
@@ -702,52 +857,37 @@ def hdl_wid_114(params: WIDParams):
...
-        'BAP/BSRC/STR/BV-32-C': '48_5_1',
-        'BAP/BSRC/STR/BV-33-C': '48_6_1',
     }
 
     stack = get_stack()
 
+    is_broadcast_str_test = params.test_case_name.startswith('BAP/BSRC/STR')
+
+    if is_broadcast_str_test:
+        stack.gap.big_bis_data_path_setup = []
```
```suggestion
```

Unused

autopts/wid/bap.py
```diff
@@ -702,52 +857,37 @@ def hdl_wid_114(params: WIDParams):
...
         raise ValueError("hdl_wid_114 is not 0 or 1")
 
-    if params.test_case_name in configurations:
+    channel_allocation = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+    streams_per_subgroup = 1
+
+    broadcast_str_cfg = BROADCAST_STR_CONFIGS_DICT.get(params.test_case_name)
+    if is_broadcast_str_test:
+        if broadcast_str_cfg is None:
+            raise ValueError(f"Unsupported STR testcase config: {params.test_case_name}")
```
```suggestion
    if is_broadcast_str_test:
        broadcast_str_cfg = BROADCAST_STR_CONFIGS_DICT.get(params.test_case_name)
        if broadcast_str_cfg is None:
            raise ValueError(f"Unsupported STR testcase config: {params.test_case_name}")
```

autopts/wid/bap.py
Thalley: Can't we put this in the existing `if` for `broadcast_str_cfg`?

Thalley: Do we need these additional run-time checks with your added unittests?

Thalley: The cast to `int` seem unnecessary with the `is_integer` check

autopts/wid/bap.py
Thalley: There is still something off. Your comment mentions PA sync, but this change can only affect the BIG sync. 

Have you invested what causes this? This change seems unnecessary.

Thalley (changes requested):

autopts/wid/bap.py
ciascai: Agreed, I’ll move it into the existing STR branch. I’d still keep the runtime checks because unit tests cover today’s mappings, but guards give fail-fast protection for future config changes

autopts/wid/bap.py
```diff
@@ -702,53 +808,49 @@ def hdl_wid_114(params: WIDParams):
...
+            raise ValueError(f"Invalid streams_per_subgroup={streams_per_subgroup}")
+        if channel_count == 0:
+            raise ValueError(f"Invalid channel_allocation=0x{channel_allocation:X}")
+        channel_ratio = channel_count / streams_per_subgroup
+        if not channel_ratio.is_integer():
+            raise ValueError(
+                "Invalid broadcast STR config: channel_count "
+                f"{channel_count} is not divisible by streams_per_subgroup {streams_per_subgroup}"
+            )
+        sdu_scale = int(channel_ratio)
```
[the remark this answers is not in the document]
ciascai: We don't need this

autopts/wid/bap.py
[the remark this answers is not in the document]
Thalley: I think so too, yes. Your unit tests does cover this, right? 

There may be some standard practices in Python that I'm not aware of (such as perhaps preferring `//` over `/`?), but I don't think there's any risk here

autopts/wid/bap.py
```diff
@@ -328,6 +331,88 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+    Attributes:
+        test_case_str: PTS test case identifier (for example "BAP/BSRC/STR/BV-31-C").
+        qos_set_name: BAS/QoS set name key (for example "48_4_1").
+        audio_config_id: BAP audio configuration identifier used by the testcase.
+        channel_allocation: PACS audio location bitfield used in codec configuration.
+        streams_per_subgroup: Number of BIS streams configured per subgroup.
+    """
+
+    test_case_str: str
+    qos_set_name: str
```
Thalley: I would still consider defining the `8_1_1`, `8_2_1`, etc. in a StrEnum or similar and use that. Since the sets are clearly defined in BAP, using defined values, instead of literal strings, may reduce the risk of typos, as well as giving a reference to the very limited subset of allowed values.

Thalley: Such a definition could also be used for the other places where we use this literal string constants (e.g. `QOS_CONFIG_SETTINGS` and `BAS_CONFIG_SETTINGS` and `hdl_wid_114.configurations`), but some of these may be better to update afterwards to reduce scope of this PR

autopts/wid/bap.py
```diff
@@ -664,11 +749,32 @@ def hdl_wid_104(params: WIDParams):
 }
 
 
+@dataclass(frozen=True)
+class BasConfig:
+    """Broadcast Audio Stream QoS/codec mapping.
+
+    Attributes:
+        codec_set_name: Key into CODEC_CONFIG_SETTINGS (for example "16_2").
```
Thalley: Another reason to use the a defined type: Here you reference `16_2` but it's used as `bas_config = BasConfig(*BAS_CONFIG_SETTINGS[qos_set_name])` where `qos_set_name` is defined as `qos_set_name: BAS/QoS set name key (for example "48_4_1").` which is clearly not the same (aside from having 2 different names, one is using the `Codec Capability Setting` from BAP (e.g. `16_2`) and the is using the `Set Name` (e.g. `16_2_1`).

autopts/wid/bap.py
```diff
@@ -664,11 +749,32 @@ def hdl_wid_104(params: WIDParams):
...
+        sdu_interval_us: SDU interval in microseconds.
+        framing: Framing mode value used by BAP (0x00 unframed, 0x01 framed).
+        max_sdu_octets: Maximum SDU size in octets for a single stream.
+        rtn: Retransmission number.
+        max_transport_latency_ms: Maximum transport latency in milliseconds.
+    """
+
+    codec_set_name: str
+    sdu_interval_us: int
+    framing: int
```
Thalley, line 852: Consider using a bool for `framing` or define a type for it since it only has two supported values

autopts/wid/bap.py
```diff
@@ -702,53 +808,49 @@ def hdl_wid_114(params: WIDParams):
...
     if stack.bap.hdl_wid_114_cnt == 0:
         broadcast_id = stack.bap.broadcast_id
     elif stack.bap.hdl_wid_114_cnt == 1:
         broadcast_id = stack.bap.broadcast_id_2
     else:
         raise ValueError("hdl_wid_114 is not 0 or 1")
 
-    if params.test_case_name in configurations:
-        qos_set_name = configurations[params.test_case_name]
+    channel_allocation = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
```
Thalley: In `BroadcastAudioConfigTest` you default to `BAP_LOCATION_MONO`, but here you default to `PACS_AUDIO_LOCATION_FRONT_LEFT`. Should these be the same?

autopts/wid/bap.py
```diff
@@ -328,6 +331,88 @@ def get_server(self, test_case: str) -> AudioConfigServer:
 AC_CONFIGS_DICT: Final[dict[str, AudioConfigTest]] = {cfg.test_case_str: cfg for cfg in AC_CONFIGS}
 
 
+BAP_LOCATION_STEREO: Final[int] = (defs.PACS_AUDIO_LOCATION_FRONT_LEFT | defs.PACS_AUDIO_LOCATION_FRONT_RIGHT)
+BAP_LOCATION_MONO: Final[int] = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
```
Thalley: We may need to consider naming these so that it's clearer that they are for broadcast only? Unicast seems to assign channel allocation significantly different (using `get_audio_locations_from_pac`)? 

Alternatively, we can keep them as is, but then consider a followup for unicast, although I guess using the PAC records for unicast is more appropriate

autopts/wid/bap.py
```diff
@@ -328,6 +331,88 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+@dataclass(frozen=True)
+class BroadcastAudioConfigTest:
+    """Declarative broadcast STR test configuration.
+
+    Attributes:
+        test_case_str: PTS test case identifier (for example "BAP/BSRC/STR/BV-31-C").
+        qos_set_name: BAS/QoS set name key (for example "48_4_1").
+        audio_config_id: BAP audio configuration identifier used by the testcase.
+        channel_allocation: PACS audio location bitfield used in codec configuration.
+        streams_per_subgroup: Number of BIS streams configured per subgroup.
```
Thalley: Please add the type here as well, similar to 
```
class AudioConfigTest:
    """
    Class that represents a unicast Audio Configuration test case

    Attributes:
        test_case_str (str): String that represents a test case, e.g. "BAP/UCL/STR/BV-523-C"
        audio_config (BapAudioConfig): A BAP audio configuration
        mono (bool): Whether the test will setup a mono stream or not
        connect_sink_in_qos (bool): Whether the test will connect the sink CIS in the QoS configured
                                    state (True) or in the enabling state (False).
        connect_source_in_qos (bool): Whether the test will connect the source CIS in the QoS
                                      configured state (True) or in the enabling state (False).
    """

```

autopts/wid/bap.py
```diff
@@ -664,11 +749,32 @@ def hdl_wid_104(params: WIDParams):
...
+class BasConfig:
+    """Broadcast Audio Stream QoS/codec mapping.
+
+    Attributes:
+        codec_set_name: Key into CODEC_CONFIG_SETTINGS (for example "16_2").
+        sdu_interval_us: SDU interval in microseconds.
+        framing: Framing mode value used by BAP (0x00 unframed, 0x01 framed).
+        max_sdu_octets: Maximum SDU size in octets for a single stream.
+        rtn: Retransmission number.
+        max_transport_latency_ms: Maximum transport latency in milliseconds.
```
Thalley: Please add the type here as well, similar to 
```
class AudioConfigTest:
    """
    Class that represents a unicast Audio Configuration test case

    Attributes:
        test_case_str (str): String that represents a test case, e.g. "BAP/UCL/STR/BV-523-C"
        audio_config (BapAudioConfig): A BAP audio configuration
        mono (bool): Whether the test will setup a mono stream or not
        connect_sink_in_qos (bool): Whether the test will connect the sink CIS in the QoS configured
                                    state (True) or in the enabling state (False).
        connect_source_in_qos (bool): Whether the test will connect the source CIS in the QoS
                                      configured state (True) or in the enabling state (False).
    """

```

autopts/wid/bap.py
```diff
@@ -758,39 +860,47 @@ def hdl_wid_114(params: WIDParams):
...
 
-    codec_set_name, *qos_config = BAS_CONFIG_SETTINGS[qos_set_name]
+    bas_config = BasConfig(*BAS_CONFIG_SETTINGS[qos_set_name])
 
     (sampling_freq, frame_duration, octets_per_frame) = \
-        CODEC_CONFIG_SETTINGS[codec_set_name]
-    audio_locations = 0x01
+        CODEC_CONFIG_SETTINGS[bas_config.codec_set_name]
     frames_per_sdu = 0x01
+    max_sdu_octets = bas_config.max_sdu_octets * sdu_scale
```
Thalley: Since we only need to do this for `STR` tests, consider removing the `sdu_scale` variable and only use the `max_sdu_octets` variable

Thalley (changes requested):

autopts/wid/bap.py
ciascai: renamed them

autopts/wid/bap.py
ciascai: I think that this will be better to do it in another PR as a refactor for all of them

autopts/wid/bap.py
ciascai: yeah, my bad

autopts/wid/bap.py
```diff
@@ -664,11 +834,32 @@ def hdl_wid_104(params: WIDParams):
...
+        rtn (int): Retransmission number.
+        max_transport_latency_ms (int): Maximum transport latency in milliseconds.
+    """
+
+    codec_set_name: str
+    sdu_interval_us: int
+    framing: bool
+    max_sdu_octets: int
+    rtn: int
+    max_transport_latency_ms: int
```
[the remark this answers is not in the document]
ciascai: BAS_CONFIG_SETTINGS will be updated in another PR, we don't want to lose the scope of this PR

autopts/wid/bap.py
```diff
@@ -702,53 +893,47 @@ def hdl_wid_114(params: WIDParams):
...
+        if streams_per_subgroup <= 0:
+            raise ValueError(f"Invalid streams_per_subgroup={streams_per_subgroup}")
+        if channel_count == 0:
+            raise ValueError(f"Invalid channel_allocation=0x{channel_allocation:X}")
+        channel_ratio = channel_count / streams_per_subgroup
+        if not channel_ratio.is_integer():
+            raise ValueError(
+                "Invalid broadcast STR config: channel_count "
+                f"{channel_count} is not divisible by streams_per_subgroup {streams_per_subgroup}"
+            )
```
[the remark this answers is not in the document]
ciascai: @Thalley ?

autopts/wid/bap.py
```diff
@@ -758,39 +943,50 @@ def hdl_wid_114(params: WIDParams):
...
         broadcast_id = btp.bap_broadcast_source_setup(streams_per_subgroup, subgroups,
                                                       coding_format, vid, cid, codec_ltvs_bytes,
-                                                      *qos_config, presentation_delay)
+                                                      bas_config.sdu_interval_us,
+                                                      bas_config.framing,
+                                                      max_sdu_octets,
+                                                      bas_config.rtn,
+                                                      bas_config.max_transport_latency_ms,
+                                                      presentation_delay)
         stack.bap.set_broadcast_id(broadcast_id)
```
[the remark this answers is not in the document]
ciascai: This is also out of scope for this PR IMO

autopts/wid/bap.py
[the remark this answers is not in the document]
Thalley: We should not implicitly convert between `bool` and `int`

autopts/wid/bap.py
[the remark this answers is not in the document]
Thalley: :shrug: 

Either way really works IMO, but if we are already doing `%` to check for multiple, then we also know that the division is an integer and `//` can be used. My original issue with `//` was that if you had `10 // 3` we did not verify that it was correct. If we use `a % b == 0` first, then we know that `//` will always be correct. No strong opinions

autopts/wid/bap.py
```diff
@@ -73,6 +73,30 @@ class AudioConfigId(Enum):
     AC_10 = auto()
     AC_11_i = auto()
     AC_11_ii = auto()
+    AC_12 = auto()
+    AC_13 = auto()
+    AC_14 = auto()
+
+
+class BroadcastQosSetName(str, Enum):
```
Thalley: Did you consider 

```suggestion
class BroadcastQosSetName(StrEnum):
```

autopts/wid/bap.py
```diff
@@ -73,6 +73,30 @@ class AudioConfigId(Enum):
...
+    """Broadcast BAS/QoS set names defined by BAP."""
+
+    BAS_8_1_1 = "8_1_1"
```
Thalley: If BAS => Broadcast Audio Stream, then that abbreviation doesn't really work for set names IMO. Consider `SET_8_1_1` or something that's unrelated to streams :D

autopts/wid/bap.py
```suggestion
class BroadcastSetName(str, Enum):
```

I wonder if we should have the same class for unicast and broadcast, or separate ones. They share the same set names, but the QoS for each set name is slightly different between the two. I think we can just keep it specific for Broadcast like this for now, and then determine if we need to rename later if we want to use this for unicast as well

autopts/wid/bap.py
```diff
@@ -328,6 +352,152 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+BROADCAST_AUDIO_LOCATION_MONO: Final[int] = defs.PACS_AUDIO_LOCATION_FRONT_LEFT
+
+
+@dataclass(frozen=True)
+class BroadcastAudioConfigTest:
+    """Declarative broadcast STR test configuration.
+
+    Attributes:
+        test_case_str (str): PTS test case identifier, for example "BAP/BSRC/STR/BV-31-C".
+        qos_set_name (BroadcastQosSetName): BAS/QoS set name key.
```
```suggestion
        qos_set_name (BroadcastSetName): set name.
```

autopts/wid/bap.py
```diff
@@ -73,6 +73,30 @@ class AudioConfigId(Enum):
     AC_10 = auto()
     AC_11_i = auto()
     AC_11_ii = auto()
+    AC_12 = auto()
+    AC_13 = auto()
+    AC_14 = auto()
+
+
+class BroadcastQosSetName(str, Enum):
+    """Broadcast BAS/QoS set names defined by BAP."""
```
```suggestion
    """Broadcast set names defined by BAP."""
```

autopts/wid/bap.py
Thalley: @ciascai Looks like you missed this one in your last update :)

Thalley (changes requested):
A few minor comments :)

autopts/wid/bap.py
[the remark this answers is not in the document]
ciascai: done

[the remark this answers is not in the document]
Thalley: @ciascai What exactly is "done" here? `framing` is still an `int`. Is this to avoid modifying `BAS_CONFIG_SETTINGS` in this PR?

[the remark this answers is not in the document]
ciascai: yes, I reverted back to int. I think that changes should be done when we will do the refactoring and it exceeds the scope of this PR

autopts/wid/bap.py
ciascai: done

autopts/wid/bap.py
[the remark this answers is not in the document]
Thalley: That's a fair argument :)

autopts/wid/bap.py
```diff
@@ -328,6 +352,152 @@ def get_server(self, test_case: str) -> AudioConfigServer:
...
+        test_case_str (str): PTS test case identifier, for example "BAP/BSRC/STR/BV-31-C".
+        qos_set_name (BroadcastSetName): set name.
+        audio_config_id (AudioConfigId): BAP audio configuration identifier used by the testcase.
+        channel_allocation (int): PACS audio location bitfield used in codec configuration.
+        streams_per_subgroup (int): Number of BIS streams configured per subgroup.
+    """
+
+    test_case_str: str
+    qos_set_name: BroadcastSetName
+    audio_config_id: AudioConfigId
```
```suggestion
        set_name (BroadcastSetName): set name.
        audio_config_id (AudioConfigId): BAP audio configuration identifier used by the testcase.
        channel_allocation (int): PACS audio location bitfield used in codec configuration.
        streams_per_subgroup (int): Number of BIS streams configured per subgroup.
    """

    test_case_str: str
    set_name: BroadcastSetName
    audio_config_id: AudioConfigId
```

autopts/wid/bap.py
```diff
@@ -702,53 +893,47 @@ def hdl_wid_114(params: WIDParams):
...
-        qos_set_name = configurations[params.test_case_name]
+    channel_allocation = BROADCAST_AUDIO_LOCATION_MONO
+    streams_per_subgroup = 1
+
+    if is_broadcast_str_test:
+        broadcast_str_cfg = BROADCAST_STR_CONFIGS_DICT.get(params.test_case_name)
+        if broadcast_str_cfg is None:
+            raise ValueError(f"Unsupported STR testcase config: {params.test_case_name}")
+
+        qos_set_name = broadcast_str_cfg.qos_set_name
```
```suggestion
        set_name = broadcast_str_cfg.set_name
```

Generally there are several references to either `qos_set_name` or `codec_set_name`. It's just called `Set name` in BAP :)

autopts/wid/bap.py
```diff
@@ -664,11 +834,32 @@ def hdl_wid_104(params: WIDParams):
...
+    Attributes:
+        codec_set_name (str): Key into CODEC_CONFIG_SETTINGS.
+        sdu_interval_us (int): SDU interval in microseconds.
+        framing (int): Framing mode used by BAP (0 unframed, 1 framed)
+        max_sdu_octets (int): Maximum SDU size in octets for a single stream.
+        rtn (int): Retransmission number.
+        max_transport_latency_ms (int): Maximum transport latency in milliseconds.
+    """
+
+    codec_set_name: str
```
```suggestion
        set_name (str): Key into CODEC_CONFIG_SETTINGS.
        sdu_interval_us (int): SDU interval in microseconds.
        framing (int): Framing mode used by BAP (0 unframed, 1 framed)
        max_sdu_octets (int): Maximum SDU size in octets for a single stream.
        rtn (int): Retransmission number.
        max_transport_latency_ms (int): Maximum transport latency in milliseconds.
    """

    set_name: str
```

I assume you want to postpone updating this to use `BroadcastSetName` until after this PR?

ciascai: indeed

autopts/wid/bap.py
ciascai: and I believe you suggest me to do it in the whole file, right?

Thalley: For any occurrences that relate to your other changes, yes :)

Thalley (approved):

Thalley:
#AutoPTS run zephyr BAP/BSRC/STR

codecoup-tester:
Scheduled PR https://github.com/auto-pts/auto-pts/pull/1799#issuecomment-4989570235, board: native+nrf54, estimated start time: 10:06:31, test case count: 48, estimated duration: 1:29:01<details><summary>Test cases to be run</summary>BAP/BSRC/STR/BV-01-C<br>BAP/BSRC/STR/BV-02-C<br>BAP/BSRC/STR/BV-03-C<br>BAP/BSRC/STR/BV-04-C<br>BAP/BSRC/STR/BV-05-C<br>BAP/BSRC/STR/BV-06-C<br>BAP/BSRC/STR/BV-07-C<br>BAP/BSRC/STR/BV-08-C<br>BAP/BSRC/STR/BV-09-C<br>BAP/BSRC/STR/BV-10-C<br>BAP/BSRC/STR/BV-11-C<br>BAP/BSRC/STR/BV-12-C<br>BAP/BSRC/STR/BV-13-C<br>BAP/BSRC/STR/BV-14-C<br>BAP/BSRC/STR/BV-15-C<br>BAP/BSRC/STR/BV-16-C<br>BAP/BSRC/STR/BV-35-C<br>BAP/BSRC/STR/BV-36-C<br>BAP/BSRC/STR/BV-37-C<br>BAP/BSRC/STR/BV-38-C<br>BAP/BSRC/STR/BV-39-C<br>BAP/BSRC/STR/BV-40-C<br>BAP/BSRC/STR/BV-41-C<br>BAP/BSRC/STR/BV-42-C<br>BAP/BSRC/STR/BV-43-C<br>BAP/BSRC/STR/BV-44-C<br>BAP/BSRC/STR/BV-45-C<br>BAP/BSRC/STR/BV-46-C<br>BAP/BSRC/STR/BV-47-C<br>BAP/BSRC/STR/BV-48-C<br>BAP/BSRC/STR/BV-49-C<br>BAP/BSRC/STR/BV-50-C<br>BAP/BSRC/STR/BV-18-C<br>BAP/BSRC/STR/BV-19-C<br>BAP/BSRC/STR/BV-20-C<br>BAP/BSRC/STR/BV-21-C<br>BAP/BSRC/STR/BV-22-C<br>BAP/BSRC/STR/BV-23-C<br>BAP/BSRC/STR/BV-24-C<br>BAP/BSRC/STR/BV-25-C<br>BAP/BSRC/STR/BV-26-C<br>BAP/BSRC/STR/BV-27-C<br>BAP/BSRC/STR/BV-28-C<br>BAP/BSRC/STR/BV-29-C<br>BAP/BSRC/STR/BV-30-C<br>BAP/BSRC/STR/BV-31-C<br>BAP/BSRC/STR/BV-32-C<br>BAP/BSRC/STR/BV-33-C</details>

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (48)</summary>BAP     BAP/BSRC/STR/BV-01-C            PASS<br>BAP     BAP/BSRC/STR/BV-02-C            PASS<br>BAP     BAP/BSRC/STR/BV-03-C            PASS<br>BAP     BAP/BSRC/STR/BV-04-C            PASS<br>BAP     BAP/BSRC/STR/BV-05-C            PASS<br>BAP     BAP/BSRC/STR/BV-06-C            PASS<br>BAP     BAP/BSRC/STR/BV-07-C            PASS<br>BAP     BAP/BSRC/STR/BV-08-C            PASS<br>BAP     BAP/BSRC/STR/BV-09-C            PASS<br>BAP     BAP/BSRC/STR/BV-10-C            PASS<br>BAP     BAP/BSRC/STR/BV-11-C            PASS<br>BAP     BAP/BSRC/STR/BV-12-C            PASS<br>BAP     BAP/BSRC/STR/BV-13-C            PASS<br>BAP     BAP/BSRC/STR/BV-14-C            PASS<br>BAP     BAP/BSRC/STR/BV-15-C            PASS<br>BAP     BAP/BSRC/STR/BV-16-C            PASS<br>BAP     BAP/BSRC/STR/BV-18-C            PASS<br>BAP     BAP/BSRC/STR/BV-19-C            PASS<br>BAP     BAP/BSRC/STR/BV-20-C            PASS<br>BAP     BAP/BSRC/STR/BV-21-C            PASS<br>BAP     BAP/BSRC/STR/BV-22-C            PASS<br>BAP     BAP/BSRC/STR/BV-23-C            PASS<br>BAP     BAP/BSRC/STR/BV-24-C            PASS<br>BAP     BAP/BSRC/STR/BV-25-C            PASS<br>BAP     BAP/BSRC/STR/BV-26-C            PASS<br>BAP     BAP/BSRC/STR/BV-27-C            PASS<br>BAP     BAP/BSRC/STR/BV-28-C            PASS<br>BAP     BAP/BSRC/STR/BV-29-C            PASS<br>BAP     BAP/BSRC/STR/BV-30-C            PASS<br>BAP     BAP/BSRC/STR/BV-31-C            PASS<br>BAP     BAP/BSRC/STR/BV-32-C            PASS<br>BAP     BAP/BSRC/STR/BV-33-C            PASS<br>BAP     BAP/BSRC/STR/BV-35-C            PASS<br>BAP     BAP/BSRC/STR/BV-36-C            PASS<br>BAP     BAP/BSRC/STR/BV-37-C            PASS<br>BAP     BAP/BSRC/STR/BV-38-C            PASS<br>BAP     BAP/BSRC/STR/BV-39-C            PASS<br>BAP     BAP/BSRC/STR/BV-40-C            PASS<br>BAP     BAP/BSRC/STR/BV-41-C            PASS<br>BAP     BAP/BSRC/STR/BV-42-C            PASS<br>BAP     BAP/BSRC/STR/BV-43-C            PASS<br>BAP     BAP/BSRC/STR/BV-44-C            PASS<br>BAP     BAP/BSRC/STR/BV-45-C            PASS<br>BAP     BAP/BSRC/STR/BV-46-C            PASS<br>BAP     BAP/BSRC/STR/BV-47-C            PASS<br>BAP     BAP/BSRC/STR/BV-48-C            PASS<br>BAP     BAP/BSRC/STR/BV-49-C            PASS<br>BAP     BAP/BSRC/STR/BV-50-C            PASS</details>

PR #1802, opened 2026-07-13
Add initial AGENTS.md

Thalley:
Add initial version of AGENTS.md. The main purpose of this file is to improve the quality of reviews by copilot.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,11 @@
+# Copilot Instructions for `auto-pts`
+
+When generating or modifying Python code in this repository:
+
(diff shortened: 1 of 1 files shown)


sjanc:
hmm why not generic AGENTS.md instead?

Thalley:
> hmm why not generic AGENTS.md instead?

Because Copilot told me that this file is still the right choice for this :D 

image

But yeah, let me try to rename it, and then see if Copilot will actually use that file :P

sjanc (approved):
I'm not an expert on this, but if it helps I'm fine with it

PR #1803, opened 2026-07-16
wid: gatt: adjust GATT/SR/GAN/BV-02-C

piotrnarajowski:
Per Bluetooth Core Specification 6.3 (Vol 3, Part F, Section 3.4.7.4), the ATT_MULTIPLE_HANDLE_VALUE_NTF PDU requires the Handle Length Value Tuple List to be 8 to (ATT_MTU - 1) octets. The server shall not truncate a Handle Length Value Tuple.

The amount of characteristics with notify properties is different for every IUT. If we pass too many handles to `gatts_notify_mult`, the resulting PDU will likely exceed the available ATT_MTU. Since we cannot truncate the final tuple, the implementation is forced to either omit tuples or reject the request entirely. The latter causes severe test failures.

Since we need only two handles to properly issue ATT_MULTIPLE_HANDLE_VALUE_NTF, we can limit the handle list to a
maximum of 2 elements.

### DIFF
--- a/autopts/wid/gatt.py
+++ b/autopts/wid/gatt.py
@@ -1084,13 +1084,17 @@ def hdl_wid_93(params: WIDParams):
         addr_type, addr = (btp.pts_addr_type_get(), btp.pts_addr_get())
 
     db = gatt_server_fetch_db().db
-    handles = [
-        db[i].handle + 1
-        for i in range(1, len(db) + 1)
-        if isinstance(db[i], GattCharacteristic) and db[i].prop & Prop.notify
-    ]
+    handles = []
 
-    btp.gatts_notify_mult(addr_type, addr, len(handles), handles)
+    for i in range(1, len(db) + 1):
+        if isinstance(db[i], GattCharacteristic) and db[i].prop & Prop.notify:
+            handles.append(db[i].handle + 1)
+            # Limit to max 2 handles to ensure the ATT_MULTIPLE_HANDLE_VALUE_NTF PDU
+            # fits within ATT_MTU without truncating HLV tuples.
+            if len(handles) == 2:
+                break
+
+    btp.gatts_notify_mult(addr_type, addr, 2, handles)
     return True
 
 
@@ -2024,8 +2028,9 @@ def hdl_wid_304(params: WIDParams):
 
 
 def hdl_wid_308(params: WIDParams):
-    # description: Please do not send an ATT_Handle_value_Multiple_notification to Lower tester until timeout(30s).
-
+    """
+    Please do not send an ATT_Handle_value_Multiple_notification to Lower tester until timeout(30s).
+    """
     # PTS description is bit odd because in test spec UT is expected to request sending
     # ATT_Handle_value_Multiple_notification and IUT should ignore that.
 
@@ -2035,17 +2040,27 @@ def hdl_wid_308(params: WIDParams):
         addr_type, addr = (btp.pts_addr_type_get(), btp.pts_addr_get())
 
     db = gatt_server_fetch_db().db
-    handles = [
-        db[i].handle + 1
-        for i in range(1, len(db) + 1)
-        if isinstance(db[i], GattCharacteristic) and db[i].prop & Prop.notify
-    ]
+    handles = []
 
+    for i in range(1, len(db) + 1):
+        if isinstance(db[i], GattCharacteristic) and db[i].prop & Prop.notify:
+            handles.append(db[i].handle + 1)
+            # Limit to max 2 handles to ensure the ATT_MULTIPLE_HANDLE_VALUE_NTF PDU
+            # fits within ATT_MTU without truncating HLV tuples.
+            if len(handles) == 2:
+                break
     # IUT may fail here (or silently ignore and return success)
     try:
-        btp.gatts_notify_mult(addr_type, addr, len(handles), handles)
-    except BTPError:
-        pass
+        btp.gatts_notify_mult(addr_type, addr, 2, handles)
+    except BTPError as e:
+        if len(handles) < 2:
+            logging.error(
+                "gatts_notify_mult failed due to insufficient handles. The database contains %d"
+                " characteristic(s) with notify properties, but at least 2 are required. BTPError: %s", len(handles), e
+            )
+            return False
+
+        logging.error("BTPError occured but is safe to ignore with valid setup")
 
     return True
 


piotrnarajowski:
#AutoPTS run zephyr GATT/SR/GAN/BV-02-C

codecoup-tester:
Scheduled PR https://github.com/auto-pts/auto-pts/pull/1803#issuecomment-4989781671, board: native+nrf54, estimated start time: 10:31:36, test case count: 1, estimated duration: 0:10:59<details><summary>Test cases to be run</summary>GATT/SR/GAN/BV-02-C</details>

Thalley (commented):
Can you elaborate the commit message on _why_? This change does not really make any sense to me xD

piotrnarajowski:
> Can you elaborate the commit message on _why_? This change does not really make any sense to me xD

@Thalley sure. While current implementation might be working for zephyr, for other IUTs that have a large amount of characteristics with notify properties in their database, this might be problematic. We only need two handles to properly issue ATT_MUTLIPLE_HANDLE_VALUE_NTF

Thalley:
> for other IUTs that have a large amount of characteristics with notify properties in their database, this might be problematic

Why? I'm not sure why whether `handles` contains 2 handles or 200 handles make a different here :o

piotrnarajowski:
> > for other IUTs that have a large amount of characteristics with notify properties in their database, this might be problematic
> 
> Why? I'm not sure why whether `handles` contains 2 handles or 200 handles make a different here :o

Per Core Specification 6.3, Vol 3, Part F, Section 3.4.7.4, the ATT_MULTIPLE_HANDLE_VALUE_NTF PDU format requires the Handle Length Value Tuple List to be 8 to (ATT_MTU - 1) octets, and “the server shall not truncate a Handle Length Value Tuple.” That means the PDU must fit within the ATT_MTU limit as a whole; you cannot chop the last tuple or send an oversized ATT PDU.

If `handles` contains 200 handles, it obviously will exceed available size. In that case, implementation could either:

- omit some tuples until they fit (how can you decide which to omit?)
- reject request and send nothing

The latter will result in test failure.

codecoup-tester:
AutoPTS Bot results:
No failed test found.
<details><summary>Successful tests (1)</summary>GATT    GATT/SR/GAN/BV-02-C             PASS</details>

Thalley:
> > > for other IUTs that have a large amount of characteristics with notify properties in their database, this might be problematic
> > 
> > 
> > Why? I'm not sure why whether `handles` contains 2 handles or 200 handles make a different here :o
> 
> Per Core Specification 6.3, Vol 3, Part F, Section 3.4.7.4, the ATT_MULTIPLE_HANDLE_VALUE_NTF PDU format requires the Handle Length Value Tuple List to be 8 to (ATT_MTU - 1) octets, and “the server shall not truncate a Handle Length Value Tuple.” That means the PDU must fit within the ATT_MTU limit as a whole; you cannot chop the last tuple or send an oversized ATT PDU.
> 
> If `handles` contains 200 handles, it obviously will exceed available size. In that case, implementation could either:
> 
>     * omit some tuples until they fit (how can you decide which to omit?)
> 
>     * reject request and send nothing
> 
> 
> The latter will result in test failure.

Ah, I understand it now. So it's an issue because we do `btp.gatts_notify_mult(addr_type, addr, len(handles), handles)` instead of doing e.g. `btp.gatts_notify_mult(addr_type, addr, 2, handles)`, so we are attempting to send a notification from _all_ handles at once. 

Now I understand the change at least, but please do add the reasoning in the commit message, and possibly also a comment in the code as to why we limit the number of handles to 2

piotrnarajowski:
> > > > for other IUTs that have a large amount of characteristics with notify properties in their database, this might be problematic
> > > 
> > > 
> > > Why? I'm not sure why whether `handles` contains 2 handles or 200 handles make a different here :o
> > 
> > 
> > Per Core Specification 6.3, Vol 3, Part F, Section 3.4.7.4, the ATT_MULTIPLE_HANDLE_VALUE_NTF PDU format requires the Handle Length Value Tuple List to be 8 to (ATT_MTU - 1) octets, and “the server shall not truncate a Handle Length Value Tuple.” That means the PDU must fit within the ATT_MTU limit as a whole; you cannot chop the last tuple or send an oversized ATT PDU.
> > If `handles` contains 200 handles, it obviously will exceed available size. In that case, implementation could either:
> > ```
> > * omit some tuples until they fit (how can you decide which to omit?)
> > 
> > * reject request and send nothing
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The latter will result in test failure.
> 
> Ah, I understand it now. So it's an issue because we do `btp.gatts_notify_mult(addr_type, addr, len(handles), handles)` instead of doing e.g. `btp.gatts_notify_mult(addr_type, addr, 2, handles)`, so we are attempting to send a notification from _all_ handles at once.
> 
> Now I understand the change at least, but please do add the reasoning in the commit message, and possibly also a comment in the code as to why we limit the number of handles to 2

Updated code and commit message. I agree that it should have been more elaborate in a first place ;)

autopts/wid/gatt.py
```diff
@@ -1084,13 +1084,17 @@ def hdl_wid_93(params: WIDParams):
...
-        if isinstance(db[i], GattCharacteristic) and db[i].prop & Prop.notify
-    ]
+    handles = []
 
-    btp.gatts_notify_mult(addr_type, addr, len(handles), handles)
+    for i in range(1, len(db) + 1):
+        if isinstance(db[i], GattCharacteristic) and db[i].prop & Prop.notify:
+            handles.append(db[i].handle + 1)
+            # Limit to max 2 handles to ensure the ATT_MULTIPLE_HANDLE_VALUE_NTF PDU
+            # fits within ATT_MTU without truncating HLV tuples.
```
Thalley, line 1093: To be a bit nitpicky; won't the size of the data be relative to the handles chosen, and we may still attempt to send data larger than the ATT PDU still? Without controlling the data and data length from AutoPTS, this only improves the situation, but does not "ensure" anything :D

Thalley (approved):

autopts/wid/gatt.py
```diff
@@ -1084,13 +1084,17 @@ def hdl_wid_93(params: WIDParams):
...
+            if len(handles) == 2:
+                break
+
+    btp.gatts_notify_mult(addr_type, addr, 2, handles)
```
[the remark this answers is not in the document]
piotrnarajowski: ATT_MULTIPLE_HANDLE_VALUE_NTF needs at least two handles anyway so it's ok if it fails here.

autopts/wid/gatt.py
piotrnarajowski: Actually this is a good point. I'm not sure how to prevent this though.

piotrnarajowski: You could do a local read of those two characteristics, put the data together and compare it to MTU on PTS side, but this seems like an overkill

Thalley (changes requested):
I think the Copilot comments are right. Since we have 

```
    try:
        btp.gatts_notify_mult(addr_type, addr, 2, handles)
    except BTPError:
        pass
```
then if `len(handles) < 2`, then we just fail silently. We should fail loudly.

piotrnarajowski:
@Thalley implemented copilot suggestions.

TBH I think we should not take any action in this wid handler and just return true

Thalley:
> @Thalley implemented copilot suggestions.
> 
> TBH I think we should not take any action in this wid handler and just return true

If AutoPTS can detect that the IUT is unable to do the requested operation, then we should not return `True` to PTS :o

Thalley (approved):

sjanc (approved):

PR #1805, opened 2026-07-17
gatt: handle read multiple variable responses in WID 147 and 148

lylezhu2012:
Add explicit response handling for Read Multiple Variable Length Characteristic Values commands in WID handlers 147 and 148.

After sending two `gattc_read_multiple_var` commands, the handlers now wait for and process both responses by calling `gattc_read_multiple_var_rsp` with store_val=True to capture the returned values while not storing the raw response.

This ensures proper synchronization and value storage for GATT client read multiple variable length operations.

### DIFF
--- a/autopts/wid/gatt.py
+++ b/autopts/wid/gatt.py
@@ -1839,6 +1839,9 @@ def hdl_wid_147(params: WIDParams):
     hdl2 = MMI.args[1]
     btp.gattc_read_multiple_var(btp.pts_addr_type_get(), btp.pts_addr_get(), hdl1, hdl2)
     btp.gattc_read_multiple_var(btp.pts_addr_type_get(), btp.pts_addr_get(), hdl1, hdl2)
+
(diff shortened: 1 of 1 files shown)


lylezhu2012:
@sjanc , Please help review the changes. It is used to support PTS cases GATT/CL/GAR/BV-10-C and GATT/CL/GAR/BV-11-C.

For PTS case GATT/CL/GAR/BV-10-C, the CONFIG_BT_EATT_MAX needs to be changed to 1, because there is no chance to select the ATT channel (enhanced or unenhanced) for GATT read multiple variable request. Then enhanced ATT channel will be used for first GATT read multiple variable request. And unenhanced ATT channel will be used for second GATT read multiple variable request. 

Or, one member needs to be added to `struct btp_gatt_read_multiple_var_cmd`. Or a new BTP cmd is required, such as `BTP_GATT_READ_MULTIPLE_VAR_V2`.

PR #1808, opened 2026-07-22
gap: add WID 249 handler for dedicated bonding procedure

lylezhu2012:
Implement handler for WID 249 which initiates bonding procedure in bondable mode with Dedicated Bonding specified in the Authentication Requirements.

The handler enables bondable mode and starts the bonding procedure using a BR/EDR address with pairing level 2, fulfilling the test case requirement for Dedicated Bonding in bondable mode.

### DIFF
--- a/autopts/wid/gap.py
+++ b/autopts/wid/gap.py
@@ -2915,3 +2915,13 @@ def hdl_wid_357(_: WIDParams):
         return False
 
     return True
+
+
+def hdl_wid_249(_: WIDParams):
+    '''
+    Please initiate the bonding procedure in bondable mode with Dedicated Bonding specified in the
+    Authentication_Requirements.
+    '''
+    btp.gap_set_bondable_on()
+    btp.gap_pair_v2(bd_addr_type=defs.BTP_BR_ADDRESS_TYPE, level=defs.BTP_GAP_CMD_PAIR_V2_LEVEL_2)
+    return True


autopts/wid/gap.py
```diff
@@ -2915,3 +2915,12 @@ def hdl_wid_357(_: WIDParams):
...
 
     return True
+
+
+def hdl_wid_249(_: WIDParams):
+    '''
+    Please initiate the bonding procedure in bondable mode with Dedicated Bonding specified in the
+    Authentication_Requirements.
+    '''
+    btp.gap_pair_v2(bd_addr_type=defs.BTP_BR_ADDRESS_TYPE, level=defs.BTP_GAP_CMD_PAIR_V2_LEVEL_2)
```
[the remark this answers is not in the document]
lylezhu2012: Dedicated Bonding refers to a procedure wherein one device connects to another only for the purpose of pairing without accessing a particular service.

For the bonding procedure performed by application, it is a dedicated bonding. So, no additional parameters are needed.

autopts/wid/gap.py
```diff
@@ -2915,3 +2915,12 @@ def hdl_wid_357(_: WIDParams):
...
+    return True
```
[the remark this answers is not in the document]
lylezhu2012: Normally, `bondable` flag is set at the beginning of the PTS case initialisation. And it is enabled by defualt.

autopts/wid/gap.py
Thalley, line 2926: Do we need to explicitly enable bondable mode here before calling `gap_pair_v2`?

Thalley, line 2926: There is a reference to `Authentication_Requirements`; can that be anything besides Level 2?

lylezhu2012: No specific requirement. Level 2 is used by default. Normally, Level 3 or 4 is explicitly required by test cases.

lylezhu2012: Currently, the bondable is default. Non-bondable needs to be set explicitly.

Thalley: My thought was that we might as well follow the instructions of the WID, rather than relying on what may or may not be default, as that may be platform dependent. I think it would make sense to explicitly enable bondable mode here

Thalley: We should not rely on defaults IMO. I don't see a reference to `Authentication_Requirements` anywhere in AutoPTS though, so not sure what it actually refers to, but if `Authentication_Requirements` could be level 3 or 4, then we should set the level accordingly

lylezhu2012: `Authentication_Requirements` is a ssp parameter. It consists of MITM and bonding info. 

WID 249 is required by case `GAP/IDLE/BON/BV-04-C`. The case only focuses on bonding mode.

So the default value `BTP_GAP_CMD_PAIR_V2_LEVEL_2` is used.

Thalley: And the Level can never be changed based on the ICS values?

lylezhu2012: I think it depends on the test specification. But I think for case GAP/IDLE/BON/BV-04-C, it is fine if the security level is not less than 2.

But may other cases or new cases will leverage WID 249, the security level depends on the specific test cases.

lylezhu2012: Got it.

Although, there are many cases where the dialog message is not accurate, but I think the bondable flag can be explicitly set here. Let me update it.

Thalley (approved):

Issue #1605, opened 2025-11-11
Unexpected timeouts on events should fail the test immediately


Thalley:
The `wait_for_event` and `wait_event_with_condition` functions start a timer when waiting for an event. However if that event is never received, the tests just continue operating as nothing has happened. In some cases this still lead to passing tests, as some of the events that AutoPTS is waiting for, are not actually needed (except to implement verification on the AutoPTS side). In worst cases, the timeout triggers an "OK" or similar to a WID with user action, thus causing false positives, as we claim that we have received an event we may not have. 

When the timer expires in the 2 above functions, the currently running test should fail immediately with information about what even timed out, and possible also dump the current list of received events. 

It may be useful to also be able to provide an option to allow the event to timeout, which is useful for tests where we expect no event, e.g. "Verify that no event has been received" WIDs. 

I fully expect that this change will expose a number of false positives in the test suites.

PR #1809, opened 2026-07-22
btp: gap: add min_level parameter to gap_wait_for_sec_lvl_change

piotrnarajowski:
Allow waiting for a minimum security level in
gap_wait_for_sec_lvl_change_ev rather than strictly requiring an exact level match.

Receiving an event with lower security level than required resulted in event timeout and extended test case execution time by 30 sec.

An established connection may have a lower security level than the waiter specified by default but is still sufficient to successfully copmlete the test case. This is applicable to most GATT/SR test cases.

Maintains  backward compatibility for all existing callers expecting exact level matching.

### DIFF
--- a/autopts/ptsprojects/stack/layers/gap.py
+++ b/autopts/ptsprojects/stack/layers/gap.py
@@ -489,7 +489,7 @@ class Gap:
     def set_connection_sec_level(self, addr, level):
         self.connections[addr].sec_level = level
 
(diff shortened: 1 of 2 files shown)


Thalley (approved):

piotrnarajowski:
@sjanc PTAL

sjanc (approved):

piotrnarajowski:
partially addresses #1605

PR #1811, opened 2026-07-24
doc: Add bearer selection and identifier tracking to GATT client BTP

lylezhu2012:
Update BTP GATT client documentation to support EATT bearer selection and command identifier tracking across all GATT client operations.

Add Bearers parameter (1 octet) to command parameters for all GATT client operations (opcodes 0x03-0x0c and 0x0e-0x14), allowing selection between ATT bearer (bit 0) and EATT bearer (bit 1).

Add Identifier (4 octets) to response parameters for all commands and corresponding events. The identifier links BTP commands to their completion events, enabling proper command-response correlation.

All corresponding response events (0x81-0x91) now include the Identifier parameter to match the initiating command.

### DIFF
--- a/doc/btp_gatt_client.txt
+++ b/doc/btp_gatt_client.txt
@@ -38,12 +38,21 @@ Commands and responses:
                 Controller Index:       <controller id>
                 Command parameters:     Address_Type (1 octet)
                                         Address (6 octets)
-                Response parameters:    <none>
+                                        Bearers (1 octet)
+                Response parameters:    Identifier (4 octets)
 
                 This procedure is used by a client to discover all primary
                 services on a server.
                 Services found during discovery are returned in the
                 Discover All Primary Services Response event.
+
+                Available Bearers:
+                bit 0 - ATT bearer
+                bit 1 - EATT bearer
+
+                The returned Identifier represents the specific BTP command. When the current BTP
+                command completes, the corresponding event will include this Identifier to indicate
+                which BTP command the event relates to.
 
                 In case of an error, the error response will be returned.
 
@@ -52,14 +61,23 @@ Commands and responses:
                 Controller Index:       <controller id>
                 Command parameters:     Address_Type (1 octet)
                                         Address (6 octets)
+                                        Bearers (1 octet)
                                         UUID_Length (1 octet)
                                         UUID (2 or 16 octets)
-                Response parameters:    <none>
+                Response parameters:    Identifier (4 octets)
 
                 This procedure is used by a client to discover primary services
                 with specific UUID on a server.
                 Services found during discovery are returned in the Discover
                 Primary Service by UUID Response event.
+
+                Available Bearers:
+                bit 0 - ATT bearer
+                bit 1 - EATT bearer
+
+                The returned Identifier represents the specific BTP command. When the current BTP
+                command completes, the corresponding event will include this Identifier to indicate
+                which BTP command the event relates to.
 
                 In case of an error, the error response will be returned.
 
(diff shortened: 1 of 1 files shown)


doc/btp_gatt_client.txt
```diff
@@ -38,120 +38,186 @@ Commands and responses:
...
+                                        Bearers (1 octet)
                                         Service_Start_Handle (2 octets)
                                         Service_End_Handle (2 octets)
-                Response parameters:    <none>
+                Response parameters:    Identifier (4 octets)
 
                 This procedure is used by a client to discover service
                 relationships to other services.
                 Services found during discovery are returned in the rFind
                 Included Services Response event.
```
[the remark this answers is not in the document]
lylezhu2012: Fixed.

doc/btp_gatt_client.txt
```diff
@@ -38,120 +38,186 @@ Commands and responses:
                 Controller Index:       <controller id>
                 Command parameters:     Address_Type (1 octet)
                                         Address (6 octets)
-                Response parameters:    <none>
+                                        Bearers (1 octet)
+                Response parameters:    Identifier (4 octets)
```
Thalley, line 42: This is a breaking protocol change. I think we may need a `_v2` version here instead

doc/btp_gatt_client.txt
```diff
@@ -38,120 +38,186 @@ Commands and responses:
...
 
                 This procedure is used by a client to discover all primary
                 services on a server.
                 Services found during discovery are returned in the
                 Discover All Primary Services Response event.
 
+                Available Bearers:
+                bit 0 - ATT bearer
+                bit 1 - EATT bearer
+
```
Thalley, line 52: Do we ever only need to pick between ATT and EATT, and not a specific (EATT) bearer?

Thalley: If we instead provide a specific bearer, then we also don't need the `Identifier` field and can instead use the bearer ID instead

Thalley (dismissed):
This seems like a quite significant breaking change

doc/btp_gatt_client.txt
sjanc: yes, I was wondering if there should be V2 but since this is essentially needed for all commands I would consider simply fixing this here (currently only NimBLE is using this and we should be able to coordinate update)

doc/btp_gatt_client.txt
sjanc: this is actually quite interesting observation...   currently there is no need to select specific EATT bearer however if "bearer_id" could be used as identifier I'm also fine with that

Thalley: One thing to consider, is that at least the Zephyr API does not have a way to choose a specific bearer (but you can choose a specific bearer type)

doc/btp_gatt_client.txt
Thalley: I see; may not be a blocker then.

doc/btp_gatt_client.txt
lylezhu2012: The case `GATT/CL/GAR/BV-10-C` requires the GATT requests both ATT and EATT bearers. In current BTP, there is no way to select the bearer. The only way to pass the case is that changing EATT count from 5 to 1.

lylezhu2012: > If we instead provide a specific bearer, then we also don't need the `Identifier` field and can instead use the bearer ID instead

How to identity the different channels of EATT? The `Identifier` is `Unenhanced ATT` or `EATT + channel`.

Thalley: > > If we instead provide a specific bearer, then we also don't need the `Identifier` field and can instead use the bearer ID instead
> 
> How to identity the different channels of EATT? The `Identifier` is `Unenhanced ATT` or `EATT + channel`.

I guess the IUT would need to provide some sort of identifier when an EATT is connected?

sjanc: To keep those changes minimal we could have dedicated command for bearer control, a bit similar on how things are handled in GAP service

@lylezhu2012 are any of those tests also executed over BR/EDR?

lylezhu2012: > I guess the IUT would need to provide some sort of identifier when an EATT is connected?

BTP would send the request through the specified bear via the protocol stack interface. If the request was sent successfully, it would save and return the current identifier and also increment the identifier.

lylezhu2012: > To keep those changes minimal we could have dedicated command for bearer control, a bit similar on how things are handled in GAP service

Do you mean?
- Set bear ATT
- Send gatt request
- Set bear EATT
- Send gatt request

> @lylezhu2012 are any of those tests also executed over BR/EDR?

GATT is unsupported transmission via BR/EDR now.

PR #1824, opened 2026-08-05
bap: bass: implement BV-08-C PAST handling for WID 388/389

ciascai:
Introduce complete BV-08-C PAST handling for add/modify
broadcast-source sequence and align synchronization
behavior across peers.

BAP/BA/BASS/BV-08-C

Testing
nRF5340 Audio DK
nRF54L15 DK


autopts/wid/bap.py
```diff
@@ -2477,6 +2477,112 @@ def hdl_wid_349(params: WIDParams):
...
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, PaSyncState.SYNC_INFO_REQ, timeout=10, remove=True)
+    if ev is None:
+        return False
+
+    btp.bap_send_past(src_id)
+
+    expected_subgroups = struct.pack('<IB', bis_sync, metadata_len)
+    ev = stack.bap.wait_broadcast_receive_state_ev(
```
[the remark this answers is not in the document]
ciascai: I don't think this is necessary

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,112 @@ def hdl_wid_349(params: WIDParams):
...
+        btp.bap_send_past(ev['src_id'])
+
+        ev = stack.bap.wait_broadcast_receive_state_ev(
+            broadcast_id, addr_type, addr, broadcaster_addr_type,
+            broadcaster_addr, PaSyncState.SYNCED, timeout=10, remove=True)
+
+        if ev is None:
+            return False
+
+    return True
```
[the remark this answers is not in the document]
ciascai: done

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+
+    if stack.bap.wait_bis_found_ev(ev['broadcast_id'], 10, False) is None:
+        return False
+
+    advertiser_sid = ev['advertiser_sid']
+    broadcast_id = ev['broadcast_id']
+    padv_interval = ev['padv_interval']
+    padv_sync = PaSyncState.SYNC_INFO_REQ
+    num_subgroups = 1
+    bis_sync = 1
```
Thalley, line 2502: I think this is fine for now, but we could consider getting the PA Sync value and BIS sync value from the description

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+    metadata_len = 0
```
Thalley: We could consider extracting the metadata from the `ev` and provide that in the `bap_add_broadcast_src` call

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+    subgroups = struct.pack('<IB', bis_sync, metadata_len)
+    btp.bap_add_broadcast_src(advertiser_sid, broadcast_id, padv_sync,
+                              padv_interval, num_subgroups, subgroups,
+                              broadcaster_addr_type, broadcaster_addr,
+                              addr_type, addr)
+
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, WildCard(), timeout=10, remove=True)
+
```
Thalley, line 2514: I don't think we should use the WildCard() here for the PA sync state, but instead do


```suggestion
    ev = stack.bap.wait_broadcast_receive_state_ev(
        broadcast_id, addr_type, addr, broadcaster_addr_type,
        broadcaster_addr, PaSyncState.SYNC_INFO_REQ, timeout=10, remove=True)

```

Thalley: Hmm, I see that you have
```python
    # Allow a fast-path where PA sync completes before we observe SYNC_INFO_REQ.
    if ev['pa_sync_state'] == PaSyncState.SYNCED:
        return True

    if ev['pa_sync_state'] != PaSyncState.SYNC_INFO_REQ:
        return False
```

below. 

We should _always_ see `SYNC_INFO_REQ` first, as that is what triggers PAST

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+    metadata_len = 0
+    subgroups = struct.pack('<IB', bis_sync, metadata_len)
+    btp.bap_add_broadcast_src(advertiser_sid, broadcast_id, padv_sync,
+                              padv_interval, num_subgroups, subgroups,
+                              broadcaster_addr_type, broadcaster_addr,
+                              addr_type, addr)
+
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, WildCard(), timeout=10, remove=True)
```
```suggestion
        broadcaster_addr, WildCard(), remove=True)
```

We can omit setting timeout explicitly IMO

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+        return True
+
+    if ev['pa_sync_state'] != PaSyncState.SYNC_INFO_REQ:
+        return False
+
+    btp.bap_send_past(ev['src_id'])
+
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, PaSyncState.SYNCED, timeout=10, remove=True)
```
```suggestion
        broadcaster_addr, PaSyncState.SYNCED, remove=True)
```

We can omit setting timeout explicitly IMO

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+    if stack.bap.wait_bis_found_ev(ev['broadcast_id'], 10, False) is None:
+        return False
+
+    broadcast_id = ev['broadcast_id']
+    padv_interval = ev['padv_interval']
+
+    # Reuse source id from the latest receive state after first round teardown.
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, WildCard(), timeout=10, remove=True)
```
Thalley: Why are we waiting for another event here? There are no actions to trigger such an event, and in `hdl_wid_388` which I assume is before this, we do `remove=True`. 

If the purpose of this is to the get `src_id`, then I would suggest to store that in the `stack` instead. (I generally do not like the concept of (mis)using events as a database/cache/storage)

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, PaSyncState.SYNCED, timeout=10, remove=True)
+
+    if ev is None:
+        return False
+
+    return True
+
+
+def hdl_wid_389(params: WIDParams):
```
Thalley, line 2539: Several of my comments for `hdl_wid_388` also applies to the similar lines here

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+                                 num_subgroups, subgroups,
+                                 addr_type, addr)
+
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, PaSyncState.SYNC_INFO_REQ, timeout=10, remove=True)
+    if ev is None:
+        return False
+
+    btp.bap_send_past(src_id)
```
Thalley, line 2578: Why do we have `bap_send_past` here, but not in `hdl_wid_388`?

Thalley (changes requested):

autopts/wid/bap.py
ciascai: in `hdl_wid_388` , we only send PAST when the initial receive-state event shows `SYNC_INFO_REQ`. If the source is already `SYNCED`, we take the fast path and return. In `hdl_wid_389`, we are modifying an existing source, so we still need to issue `bap_send_past(src_id)` after the modify to drive the transition from `SYNC_INFO_REQ` to `SYNCED` for the updated source.

autopts/wid/bap.py
ciascai: done

autopts/wid/bap.py
ciascai: done

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+    stack.bap.set_broadcast_src_id(ev['src_id'])
+
+    # Allow a fast-path where PA sync completes before we observe SYNC_INFO_REQ.
+    if ev['pa_sync_state'] == PaSyncState.SYNCED:
+        return True
+
+    if ev['pa_sync_state'] != PaSyncState.SYNC_INFO_REQ:
+        return False
+
+    btp.bap_send_past(ev['src_id'])
```
Thalley: Something doesn't sound right here. The description states that we shall add a source with `PAST(01)`, there should never be a PA Sync until we do `btp.bap_send_past(ev['src_id'])`, so the `if ev['pa_sync_state'] == PaSyncState.SYNCED:` seems like an error

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,116 @@ def hdl_wid_349(params: WIDParams):
...
+    broadcast_id = ev['broadcast_id']
+    padv_interval = ev['padv_interval']
+    src_id = stack.bap.broadcast_src_id
+    if src_id is None:
+        return False
+
+    padv_sync = int(re.search(r'PAST\((\d+)\)', params.description).group(1), 10)
+    num_subgroups = 1
+    bis_sync = int(re.search(r'BIS INDEX:\s*(0x[0-9A-Fa-f]+)', params.description).group(1), 16)
+    metadata = base_found_ev['ltvs']
```
Thalley, line 2558: In this handler you do regex to get information from the description, but you don't do that in `hdl_wid_388` even though they are effectively the same and the only difference is that we are doing `add_src` in one and `mod_src` in the other. Suggest to be consistent

Thalley (changes requested):

autopts/wid/bap.py
ciascai: done it in 388 also

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -27,6 +27,7 @@ def __init__(self):
         self.ase_configs = []
         self.broadcast_id = None
         self.broadcast_id_2 = None
+        self.broadcast_src_id = None
```
Thalley: We really should document each field :D Because now we store 3 broadcast IDs which have 3 different uses / semantics

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,112 @@ def hdl_wid_349(params: WIDParams):
...
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, PaSyncState.SYNC_INFO_REQ, timeout=10, remove=True)
+
+    if ev is None:
+        return False
+
+    stack.bap.set_broadcast_src_id(ev['src_id'])
+
+    if ev['pa_sync_state'] != PaSyncState.SYNC_INFO_REQ:
+        return False
```
```suggestion
```

You already have this as a condition in `wait_broadcast_receive_state_ev` above

autopts/wid/bap.py
```diff
@@ -2477,6 +2477,112 @@ def hdl_wid_349(params: WIDParams):
...
+                              addr_type, addr)
+
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type,
+        broadcaster_addr, PaSyncState.SYNC_INFO_REQ, timeout=10, remove=True)
+
+    if ev is None:
+        return False
+
+    stack.bap.set_broadcast_src_id(ev['src_id'])
```
Thalley: `src_id` and `broadcast_src_id` are semantically different. 

`src_id` refer to the ID in the Broadcast Receive State and `broadcast_src_id` indicates that this is a Broadcast Source ID, which is a completely different thing 😅 

Perhaps `set_broadcast_src_id` should be `set_brs_src_id` where brs => Broadcast Receive State (with a similar rename for the field itself)

Thalley (changes requested):

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -25,8 +25,12 @@ def __init__(self):
     def __init__(self):
         self.peers = {}
         self.ase_configs = []
+        # BIG Broadcast_ID used by local Broadcast Source procedures.
         self.broadcast_id = None
+        # Secondary BIG Broadcast_ID used by dual-broadcast test flows.
         self.broadcast_id_2 = None
+        # Broadcast Receive State src_id assigned by Add Source/Receive State.
+        self.broadcast_src_id = None
```
Thalley:
```suggestion
        self.brs_src_id = None
```

Thalley: This is not how we have documented class fields/attributes in several other classes in AutoPTS. Suggest to follow the same approach as the others (You can search for `Attributes:`)

autopts/ptsprojects/stack/layers/bap.py
```diff
@@ -63,6 +67,9 @@ def set_broadcast_id(self, broadcast_id):
     def set_broadcast_id_2(self, broadcast_id):
         self.broadcast_id_2 = broadcast_id
 
+    def set_brs_src_id(self, src_id):
+        self.broadcast_src_id = src_id
```
Thalley:
```suggestion
        self.brs_src_id = src_id
```

Thalley (changes requested):

Thalley (approved):
LGTM. @ciascai @sjanc The test will still fail due to https://support.bluetooth.com/hc/en-us/requests/190836 - Should we merge this without it actually passing the test, or should we wait until there's a fix?

ciascai:
> LGTM. @ciascai @sjanc The test will still fail due to https://support.bluetooth.com/hc/en-us/requests/190836 - Should we merge this without it actually passing the test, or should we wait until there's a fix?

your call @Thalley

Thalley:
> > LGTM. @ciascai @sjanc The test will still fail due to https://support.bluetooth.com/hc/en-us/requests/190836 - Should we merge this without it actually passing the test, or should we wait until there's a fix?
> 
> your call @Thalley

I'd like to wait for input from @sjanc :) We cannot verify the full correctness of these WID handlers until PTS solves the issue, so that the test passes, right?

sjanc (approved):
I'm fine with merging this

PR #1828, opened 2026-08-10
rfcomm: support RFCOMM/DEVA-DEVB/RFC/BV-14-C test case

lylezhu2012:
* rfcomm: add support for sending Remote Line Status command

Add BTP command to send RFCOMM Remote Line Status (RLS) command over an established RFCOMM connection. The RLS command is used to indicate line status errors to the remote device.

Implement rfcomm_send_rls() function in BTP to send RLS with configurable line status including overrun, parity, and framing errors. Define RLS error constants for no error, overrun error, parity error, and framing error conditions.

Update BTP RFCOMM documentation with opcode 0x08 command details including address, channel, and line status parameters.

* rfcomm: support RFCOMM/DEVA-DEVB/RFC/BV-14-C test case

Extend WID 23 handler to support sending RFCOMM Remote Line Status (RLS) command with framing error status for test case `RFCOMM/DEVA-DEVB/RFC/BV-14-C`.

Update WID 23 docstring to reflect dual functionality for both data transmission and RLS command initiation.


sjanc (approved):

doc/btp_rfcomm.txt
```diff
@@ -115,6 +115,32 @@ Commands and responses:
...
+                                        Channel (1 octet)
+                                        Line_Status (1 octet)
+                Response parameters:    <none>
+
+                This command is used to send Remote Line Status (RLS)
+                command over RFCOMM connection.
+                Address is the peer BD_ADDR and Channel is the RFCOMM
+                server channel number.
+
+                Available Line_Status:
```
sjanc, line 132: I'm fine if you wan to keep this a bit field, however this could be simply defined as values 
0x00 - no error
0x01 - overrun error
0x02 - parity error
0x03 - framing error

since you use that only as values, not bit mask

or multiple errors are possible at same time? (you don't seem to use it like that in wid)

lylezhu2012: The value originates from RFCOMM Specification. I suggest using the original value defined in the specification.

PR #1830, opened 2026-08-13
wid: cap: Use non-default audio context in audio updating tests

sjanc:
PTS 8.14 default TSPX_Streaming_ASE Audio_Contexts is little endian 0002 (0x0200). Use non-default context type in audio update tests so that it is not rejected by IUT.

### DIFF
--- a/autopts/wid/cap.py
+++ b/autopts/wid/cap.py
@@ -204,12 +204,12 @@ def hdl_wid_202(_: WIDParams):
 
 
 wid_310_settings = {
-    "CAP/INI/UST/BV-32-C": pack_metadata(stream_context=0x0200),
-    "CAP/INI/UST/BV-33-C": pack_metadata(stream_context=0x0200),
-    "CAP/INI/UST/BV-34-C": pack_metadata(stream_context=0x0200, ccid_list=[0x00]),
-    "CAP/INI/UST/BV-35-C": pack_metadata(stream_context=0x0200, ccid_list=[0x00]),
-    "CAP/INI/UST/BV-36-C": pack_metadata(stream_context=0x0200, ccid_list=[0x00, 0x01]),
-    "CAP/INI/UST/BV-37-C": pack_metadata(stream_context=0x0200, ccid_list=[0x00, 0x01]),
+    "CAP/INI/UST/BV-32-C": pack_metadata(stream_context=0x0100),
+    "CAP/INI/UST/BV-33-C": pack_metadata(stream_context=0x0100),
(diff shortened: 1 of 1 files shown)


autopts/wid/cap.py
```diff
@@ -204,12 +204,12 @@ def hdl_wid_202(_: WIDParams):
 
 
 wid_310_settings = {
-    "CAP/INI/UST/BV-32-C": pack_metadata(stream_context=0x0200),
-    "CAP/INI/UST/BV-33-C": pack_metadata(stream_context=0x0200),
-    "CAP/INI/UST/BV-34-C": pack_metadata(stream_context=0x0200, ccid_list=[0x00]),
-    "CAP/INI/UST/BV-35-C": pack_metadata(stream_context=0x0200, ccid_list=[0x00]),
-    "CAP/INI/UST/BV-36-C": pack_metadata(stream_context=0x0200, ccid_list=[0x00, 0x01]),
-    "CAP/INI/UST/BV-37-C": pack_metadata(stream_context=0x0200, ccid_list=[0x00, 0x01]),
+    "CAP/INI/UST/BV-32-C": pack_metadata(stream_context=0x0100),
```
Thalley, line 207: Wouldn't it make more sense to have something like `TSPX_Streaming_ASE_Audio_Contexts << 1` to ensure that it's always different from `TSPX_Streaming_ASE_Audio_Contexts`?

sjanc: yeah, I guess that should work too,  I'll update PR

sjanc: actually, there is no API to get current value of IXIT so we would have to keep it in stack (set to known value) and use it here

I'd rather get rid of setting IXITs than adding more, especially if not absolutely required to do so

Thalley: > I'd rather get rid of setting IXITs than adding more, especially if not absolutely required to do so

I understand that this is wanted, but it does also come with the significant caveat that we cannot get the PIXIT values at runtime in the Python code, and must thus result to using more hardcoded values, and if PTS at any time change their defaults, we'll probably see test failures and will need to update our hardcoded values. This seems like a maintenance nightmare IMO. 

It should be possible to remove some of the "setting IXITs" and then, based on the workspace, read the default values from the XML file(s), so that we have a 1:1 between what PTS think the values are, and what AutoPTS think the values are. This is outside the scope of this PR though IMO, and should be considered a separate task

Thalley (approved):

autopts/wid/cap.py
sjanc: yeah,  reading from workspace xml is possible workaround (ideally PTS should have RPC for this but well...)

PR #1836, opened 2026-08-18
GMCS: Use Goto Track for MCP inactive track setup

ciascai:
BV-41-C and BV-45-C require the current track to be neither first nor
last before PTS sends the First Track or Last Track opcode. Use the
Goto Track control point opcode with parameter 2 in WID 44 to place
the player on the middle track directly.

GMCS/SR/MCP/BV-41-C

Testing
nRF5340 Audio DK
nRF54L15 DK

### DIFF
--- a/autopts/wid/gmcs.py
+++ b/autopts/wid/gmcs.py
@@ -177,11 +177,9 @@ def hdl_wid_44(params: WIDParams):
        current track to a track other than the first or last within a group
        of tracks."""
 
-    if params.test_case_name in ['GMCS/SR/MCP/BV-41-C', 'GMCS/SR/MCP/BV-45-C']:
-        # This applies only for testcases where media player is initially
-        # in inactive state
-        MEDIA_PROXY_OP_NEXT_TRACK = 0x31
-        btp.gmcs_control_point_cmd(MEDIA_PROXY_OP_NEXT_TRACK, 0)
+    MEDIA_PROXY_OP_GOTO_TRACK = 0x34
+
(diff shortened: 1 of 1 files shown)


autopts/wid/gmcs.py
```diff
@@ -181,7 +181,13 @@ def hdl_wid_44(params: WIDParams):
         # This applies only for testcases where media player is initially
         # in inactive state
         MEDIA_PROXY_OP_NEXT_TRACK = 0x31
+        MEDIA_PROXY_OP_FIRST_TRACK = 0x32
+
+        # Relative track opcodes are ignored while inactive, so activate with
+        # First Track before moving, then restore the Inactive initial state.
+        btp.gmcs_control_point_cmd(MEDIA_PROXY_OP_FIRST_TRACK, 0)
         btp.gmcs_control_point_cmd(MEDIA_PROXY_OP_NEXT_TRACK, 0)
+        btp.gmcs_inactive_state_set()
```
Thalley: The WID handler states that we need to go to any other track than the first or last. 

The comment here seems specific to Zephyr (as other IUT may support NEXT_TRACK while inactive), so this is not a good general solution. 

My suggestion would be to use `Goto Track` track instead of `First Track` followed by `Second Track` followed by changing the state manually. 

Using `Goto Track` we can just go to the 2nd track in the group, and only need to perform a single action here, instead of 3. 

Not sure why this code is guarded by a test case name check either; the WID is quite generic.

Thalley (changes requested):

Thalley (approved):

PR #1839, opened 2026-08-20
BAP: fix BASS Add/Modify/Remove/Set Broadcast Code ordering

ciascai:
BV-05-C, 6 and 7 timed out on Lower Tester 1 because the IUT
wrote to the Broadcast Audio Scan Control Point before the
Scan Delegator was ready, so PTS acknowledged the ATT write but
never processed it as the expected test step.
WID 346: wait for the Broadcast Receive State reporting the
requested BIS_Sync before writing Modify Source.
WID 347: add the source with the PA SYNC 0x00/BIS INDEX 0x00000000
values from the WID description, otherwise the source is still
synced and PTS answers "Can't perform remove operation".
WID 348: send PAST first and write Set Broadcast Code only after
the Scan Delegator reports Broadcast_Code_Required.

BAP/BA/BASS/BV-05-C
BAP/BA/BASS/BV-06-C
BAP/BA/BASS/BV-07-C 

Testing
nRF5340 Audio DK
nRF54L15 DK

### DIFF
--- a/autopts/pybtp/btp/bap.py
+++ b/autopts/pybtp/btp/bap.py
@@ -22,9 +22,9 @@ from autopts.pybtp.btp.btp import CONTROLLER_INDEX, btp_hdr_check, pts_addr_get,
 from autopts.pybtp.btp.btp import get_iut_method as get_iut
 from autopts.pybtp.btp.gap import __gap_current_settings_update
 from autopts.pybtp.types import (
-    BASSPASyncState,
     BIGEncryption,
     BTPError,
+    PaSyncState,
     addr_str_to_le_bytes,
     hex_str_to_le_bytes,
     le_bytes_to_hex_str,
@@ -402,7 +402,7 @@ def bap_scan_delegator_add_src(broadcaster_addr_type, broadcaster_addr,
                                big_encryption, num_subgroups, subgroups):
     logging.debug("")
 
-    if not BASSPASyncState.NOT_SYNCED <= pa_sync_state <= BASSPASyncState.NO_PAST:
+    if not PaSyncState.NOT_SYNCED <= pa_sync_state <= PaSyncState.NO_PAST:
         raise BTPError(f'Invalid pa_sync_state {pa_sync_state}')
 
     if not BIGEncryption.NOT_ENCRYPTED <= big_encryption <= BIGEncryption.BAD_CODE:
--- a/autopts/wid/bap.py
+++ b/autopts/wid/bap.py
@@ -41,7 +41,9 @@ from autopts.pybtp.types import (
     AdType,
     ASCSState,
     AudioDir,
+    BIGEncryption,
     BTPError,
+    PaSyncRequest,
     PaSyncState,
     WIDParams,
     create_lc3_ltvs_bytes,
@@ -2211,7 +2213,7 @@ def hdl_wid_345(params: WIDParams):
     advertiser_sid = ev['advertiser_sid']
     broadcast_id = ev['broadcast_id']
     padv_interval = ev['padv_interval']
-    padv_sync = 0x02
+    padv_sync = PaSyncRequest.SYNC_NO_PAST
     num_subgroups = 1
     bis_sync = 1
     metadata_len = 0
@@ -2264,7 +2266,7 @@ def hdl_wid_346(params: WIDParams):
     advertiser_sid = ev['advertiser_sid']
     broadcast_id = ev['broadcast_id']
     padv_interval = ev['padv_interval']
-    padv_sync = 0x02
+    padv_sync = PaSyncRequest.SYNC_NO_PAST
     num_subgroups = 1
     bis_sync = 1
     metadata_len = 0
@@ -2284,14 +2286,18 @@ def hdl_wid_346(params: WIDParams):
     if ev['pa_sync_state'] == PaSyncState.SYNC_INFO_REQ:
         btp.bap_send_past(ev['src_id'])
 
+    # PTS drops the Modify Source write if the BIG sync is not established yet.
+    # PTS may report PA and BIG sync in the same notification, so only wait for
+    # another one if the one we already have does not satisfy that condition.
+    if ev['pa_sync_state'] != PaSyncState.SYNCED or ev.get('subgroups') != subgroups:
         ev = stack.bap.wait_broadcast_receive_state_ev(
-            broadcast_id, addr_type, addr, broadcaster_addr_type,
-            broadcaster_addr, WildCard(), timeout=10, remove=True)
+            broadcast_id, addr_type, addr, broadcaster_addr_type, broadcaster_addr,
+            PaSyncState.SYNCED, subgroups=subgroups, remove=True)
 
         if ev is None:
             return False
 
-    padv_sync = 0x00
+    padv_sync = PaSyncRequest.NO_SYNC
     bis_sync = 0
     metadata_len = 0
     subgroups = struct.pack('<IB', bis_sync, metadata_len)
@@ -2301,7 +2307,7 @@ def hdl_wid_346(params: WIDParams):
 
     ev = stack.bap.wait_broadcast_receive_state_ev(
         broadcast_id, addr_type, addr, broadcaster_addr_type,
-        broadcaster_addr, WildCard(), timeout=10, remove=True)
+        broadcaster_addr, PaSyncState.NOT_SYNCED, subgroups=subgroups, remove=True)
 
     if ev is None:
         return False
@@ -2332,9 +2338,9 @@ def hdl_wid_347(params: WIDParams):
     broadcast_id = ev['broadcast_id']
     padv_interval = ev['padv_interval']
     num_subgroups = 1
-    # Ignore wrong values of PA SYNC and BIS INDEX in WID description
-    padv_sync = 0x02
-    bis_sync = 1
+    # Add receive state without requesting to PA or BIS Sync as per the WID description
+    padv_sync = PaSyncRequest.NO_SYNC
+    bis_sync = 0
     metadata_len = 0
     subgroups = struct.pack('<IB', bis_sync, metadata_len)
     btp.bap_add_broadcast_src(advertiser_sid, broadcast_id, padv_sync,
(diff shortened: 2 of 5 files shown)


autopts/wid/bap.py
```diff
@@ -2291,6 +2292,14 @@ def hdl_wid_346(params: WIDParams):
         if ev is None:
             return False
 
+    # PTS drops the Modify Source write if the BIG sync is not established yet
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type, broadcaster_addr,
+        WildCard(), subgroups=subgroups, timeout=10, remove=True)
```
```suggestion
        PaSyncState.SYNCED, subgroups=subgroups, timeout=10, remove=True)
```

```suggestion
        WildCard(), subgroups=subgroups, remove=True)
```

Thalley: Timeout is already default to `10`

autopts/wid/bap.py
```diff
@@ -2291,6 +2292,14 @@ def hdl_wid_346(params: WIDParams):
         if ev is None:
             return False
```
```suggestion
```

We don't know if we will receive a BRS with just PA sync state set, or with both PA and BIS sync states set, and since this has `remove=true`, then we may not get another one for just the BIS sync.

autopts/wid/bap.py
```suggestion
        broadcaster_addr, PaSyncState.NOT_SYNCED, subgroups=subgroups, remove=True)
```

autopts/wid/bap.py
```diff
@@ -2332,9 +2341,9 @@ def hdl_wid_347(params: WIDParams):
     broadcast_id = ev['broadcast_id']
     padv_interval = ev['padv_interval']
     num_subgroups = 1
-    # Ignore wrong values of PA SYNC and BIS INDEX in WID description
-    padv_sync = 0x02
-    bis_sync = 1
+    # BASS rejects Remove Source while the source is synced, so add it unsynced
```
```suggestion
    # Add receive state without requesting to PA or BIS Sync as per the WID description
```

autopts/wid/bap.py
```suggestion
        broadcaster_addr, PaSyncState.NOT_SYNCED, subgroups=subgroups, remove=True)
```

Thalley, line 1: Somewhat unreleated, we have 
```python
class PaSyncState:
    NOT_SYNCED = 0x00
    SYNC_INFO_REQ = 0x01
    SYNCED = 0x02
    FAILED_TO_SYNC = 0x03
    NO_PAST = 0x04

class BASSPASyncState:
    """Periodic Advertising state reported by the Scan Delegator.

    Values match enum bt_bap_pa_state in Zephyr.
    """
    NOT_SYNCED = 0x00
    INFO_REQ = 0x01
    SYNCED = 0x02
    FAILED = 0x03
    NO_PAST = 0x04
```

We should remove one of these, as they are the same.

autopts/wid/bap.py
```diff
@@ -2291,6 +2292,14 @@ def hdl_wid_346(params: WIDParams):
         if ev is None:
             return False
```
Thalley, line 2299: These lines should be removed

We don't know if we will receive a BRS with just PA sync state set, or with both PA and BIS sync states set, and since this has `remove=true`, then we may not get another one for just the BIS sync.

autopts/wid/bap.py
Thalley: We should never do PAST here, since we shall set `PA SYNC: 0x00`

Thalley: Somewhat unrelated, but we should consider defining the 3 possible `padv_sync` values in an enum and use that enum

```suggestion
        broadcaster_addr, PaSyncState.NOT_SYNCED, remove=False)
```

autopts/wid/bap.py
```diff
@@ -2410,17 +2419,19 @@ def hdl_wid_348(params: WIDParams):
...
     if ev['pa_sync_state'] == PaSyncState.SYNC_INFO_REQ:
         btp.bap_send_past(ev['src_id'])
 
-        ev = stack.bap.wait_broadcast_receive_state_ev(
-            broadcast_id, addr_type, addr, broadcaster_addr_type,
-            broadcaster_addr, WildCard(), timeout=10, remove=True)
+    # PTS drops the Set Broadcast Code write until it reports Broadcast_Code_Required
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type, broadcaster_addr,
+        WildCard(), big_encryption=BIGEncryption.BROADCAST_CODE_REQUIRED,
```
```suggestion
        PaSyncState.SYNCED, big_encryption=BIGEncryption.BROADCAST_CODE_REQUIRED,
```

autopts/wid/bap.py
```diff
@@ -2410,17 +2419,19 @@ def hdl_wid_348(params: WIDParams):
...
+        timeout=10, remove=True)
```
```suggestion
        remove=True)
```

Thalley (changes requested):
Some the requests are only partially related to your other fixes, but since you are fixing these tests up, might as well do it 100% :D

autopts/wid/bap.py
ciascai: it make sense what you say, but these are request values, not `PaSyncState` values: 0x00 means no PA sync, 0x01 means PA sync with PAST, and 0x02 means PA sync without PAST. For that I will need to define soemthing like PaSyncRequest beside the existing state constants and replace the literal assignments across BAP/CAP handlers, and IMO this will exceed the scope of this PR

autopts/wid/bap.py
```diff
@@ -2332,9 +2341,9 @@ def hdl_wid_347(params: WIDParams):
     broadcast_id = ev['broadcast_id']
     padv_interval = ev['padv_interval']
     num_subgroups = 1
-    # Ignore wrong values of PA SYNC and BIS INDEX in WID description
-    padv_sync = 0x02
-    bis_sync = 1
+    # BASS rejects Remove Source while the source is synced, so add it unsynced
+    padv_sync = 0x00
+    bis_sync = 0
```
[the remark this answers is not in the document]
ciascai: this will create confusion IMO

autopts/wid/bap.py
Thalley: Fair, but yes, it would be a new enum :)

autopts/wid/bap.py
```diff
@@ -2284,12 +2285,13 @@ def hdl_wid_346(params: WIDParams):
     if ev['pa_sync_state'] == PaSyncState.SYNC_INFO_REQ:
         btp.bap_send_past(ev['src_id'])
 
-        ev = stack.bap.wait_broadcast_receive_state_ev(
-            broadcast_id, addr_type, addr, broadcaster_addr_type,
-            broadcaster_addr, WildCard(), timeout=10, remove=True)
+    # PTS drops the Modify Source write if the BIG sync is not established yet
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type, broadcaster_addr,
+        PaSyncState.SYNCED, subgroups=subgroups, remove=True)
```
Thalley: In the case that the first `wait_broadcast_receive_state_ev` has `PaSyncState.SYNCED` and `subgroups=subgroups`, then we won't get another one here. 

So if PTS sends a single notification for both PA and BIG sync, then we would fail here.

autopts/wid/bap.py
```diff
@@ -2410,17 +2402,19 @@ def hdl_wid_348(params: WIDParams):
...
         btp.bap_send_past(ev['src_id'])
 
-        ev = stack.bap.wait_broadcast_receive_state_ev(
-            broadcast_id, addr_type, addr, broadcaster_addr_type,
-            broadcaster_addr, WildCard(), timeout=10, remove=True)
+    # PTS drops the Set Broadcast Code write until it reports Broadcast_Code_Required
+    ev = stack.bap.wait_broadcast_receive_state_ev(
+        broadcast_id, addr_type, addr, broadcaster_addr_type, broadcaster_addr,
+        PaSyncState.SYNCED, big_encryption=BIGEncryption.BROADCAST_CODE_REQUIRED,
+        remove=True)
```
Thalley: Ditto; if the first even we receive has both `PaSyncState.SYNCED` and `big_encryption=BIGEncryption.BROADCAST_CODE_REQUIRED` then we won't get another event and we would fail here

Thalley (changes requested):
There are still some potentially problematic paths in the code. We should probably only do `wait_broadcast_receive_state_ev` if the previous event does not already have all the information that we wait for

autopts/wid/bap.py
ciascai: yeah, but you suggested this changes...

Thalley: Not sure why that matters - If you disagree with the changes or change requests, please explain why :)

autopts/wid/bap.py
```diff
@@ -2410,18 +2405,24 @@ def hdl_wid_348(params: WIDParams):
...
-    btp.bap_set_broadcast_code(ev['src_id'], stack.bap.broadcast_code)
-
     if ev['pa_sync_state'] == PaSyncState.SYNC_INFO_REQ:
         btp.bap_send_past(ev['src_id'])
 
+    # PTS drops the Set Broadcast Code write until it reports Broadcast_Code_Required.
+    # PTS may report PA sync and the code requirement in the same notification, so
+    # only wait for another one if the one we already have does not satisfy that.
+    if ev['pa_sync_state'] != PaSyncState.SYNCED or \
+            ev.get('big_encryption') != BIGEncryption.BROADCAST_CODE_REQUIRED:
```
```suggestion
    if ev['pa_sync_state'] != PaSyncState.SYNCED or ev.get('big_encryption') != BIGEncryption.BROADCAST_CODE_REQUIRED:
```

Thalley: No need to split this line :)

Thalley (approved):
Minor formatting request, otherwise LGTM

sjanc (approved):

Thalley (approved):

autopts/wid/bass.py
```diff
@@ -232,7 +232,7 @@ def hdl_wid_116(_: WIDParams):
     btp.bap_scan_delegator_add_src(
         broadcaster_addr_type, broadcaster_addr,
         advertiser_sid, broadcast_id,
-        BASSPASyncState.NOT_SYNCED, big_encryption, 1, subgroups)
+        PaSyncState.NOT_SYNCED, big_encryption, 1, subgroups)
```
Thalley:
```suggestion
        PaSyncRequest.NO_SYNC, big_encryption, 1, subgroups)
```

Thalley (changes requested):
What Copilot is complaining about :) 

This is where something like type hinting would have also been handy

autopts/pybtp/btp/bap.py
```diff
@@ -402,7 +402,7 @@ def bap_scan_delegator_add_src(broadcaster_addr_type, broadcaster_addr,
                                big_encryption, num_subgroups, subgroups):
     logging.debug("")
 
-    if not BASSPASyncState.NOT_SYNCED <= pa_sync_state <= BASSPASyncState.NO_PAST:
+    if not PaSyncRequest.NO_SYNC <= pa_sync_state <= PaSyncRequest.SYNC_NO_PAST:
```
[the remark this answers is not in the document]
ciascai: yes, that make sense

autopts/wid/bass.py
ciascai: solved in the other commit

Thalley (approved):

PR #1845, opened 2026-08-27
BAP: synchronize all discovered BISes

ciascai:
Build the BIS_Sync bitfield from all BIS_FOUND events
for the current broadcast ID instead of relying on
test-case number ranges. This ensures BSNK and BSRC
multiple-BIS tests synchronize to every BIS advertised
in the BASE while preserving support for single-BIS
tests.

BAP/BSNK/STR/BV-59-C

Testing
nRF5340 Audio DK
nRF54L15 DK

### DIFF
--- a/autopts/wid/bap.py
+++ b/autopts/wid/bap.py
@@ -759,23 +759,28 @@ def hdl_wid_100(params: WIDParams):
                                 5, 100, False, 0, ev['addr_type'], ev['addr'])
 
     broadcast_id = ev['broadcast_id']
-    ev = stack.bap.wait_bis_found_ev(broadcast_id, 20, False)
-    if ev is None:
-        log(f'BIS not found, broadcast ID {broadcast_id}')
-        return False
 
     if params.test_case_name.startswith('BAP/BA'):
         return True
 
-    # BIS_Sync bitfield uses bit 0 for BIS Index 1
-    requested_bis_sync = 1
-    bis_ids = [1]
-    if params.test_case_name.startswith('BAP/BSNK/STR') or \
-            params.test_case_name.startswith('BAP/BSRC/STR'):
-        tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-        if tc_num >= 18 and tc_num <= 34:
-            requested_bis_sync |= 2
-            bis_ids.append(2)
+    bis_events = {}
+    while True:
+        # Only apply a long timeout for the first event as any remaining events would come right after
+        wait_timeout = 20 if not bis_events else 1
+        bis_event = stack.bap.wait_bis_found_ev(broadcast_id, wait_timeout, True)
+        if bis_event is None:
+            break
+        bis_events[bis_event['bis_id']] = bis_event
+
+    if not bis_events:
+        log(f'BIS not found for broadcast ID {broadcast_id}')
+        return False
+
+    bis_ids = sorted(bis_events)
+    requested_bis_sync = 0
+    for bis_index in bis_ids:
+        # Add the BIS index to the bitfield (BIS index 0x01 => BIT(0), index 0x02 => BIT(1), etc.)
+        requested_bis_sync |= defs.BIT(bis_index - 1)
 
     btp.bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync)
 


autopts/wid/bap.py
```diff
@@ -772,10 +770,14 @@ def hdl_wid_100(params: WIDParams):
...
         tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-        if tc_num >= 18 and tc_num <= 34:
+        if (18 <= tc_num <= 34) or (51 <= tc_num <= 66):
+            requested_bis_sync |= 2
+            bis_ids.append(2)
+    elif params.test_case_name.startswith('BAP/BSRC/STR'):
+        tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
+        if 18 <= tc_num <= 34:
             requested_bis_sync |= 2
             bis_ids.append(2)
```
Thalley: Is there any case where we don't want to sync to all BIS available? 

We get a `wait_bis_found_ev` per BIS in the BASE. Instead of doing these test-case specific workarounds using the tc_num, I would suggest to implement a generic solution that uses the `wait_bis_found_ev` event(s) to determine how many BIS there are, and then sync to all of them, unless otherwise specified. 

This is of course based on the assumption that PTS will be using an appropriate broadcast source for the test, and that the IUT is not reporting more BIS than it supports.

autopts/wid/bap.py
Thalley: Since this returns after the first `wait_bis_found_ev` event, there is a (very) small chance that when we do `for bis_event in stack.bap.event_queues[defs.BTP_BAP_EV_BIS_FOUND]`, the `event_queues` may still only have a single event, even if more is coming. 

We should consider doing `wait_bis_found_ev` in a loop and wait a small time for more. 

Maybe something like

```python
    ev = stack.bap.wait_bis_found_ev(broadcast_id, 20, False)
    if ev is None:
        log(f'BIS not found, broadcast ID {broadcast_id}')
        return False

    while (True):
        ev = stack.bap.wait_bis_found_ev(broadcast_id, 1, False)
        if ev is None:
            break
```

autopts/wid/bap.py
```diff
@@ -769,15 +767,14 @@ def hdl_wid_100(params: WIDParams):
...
-        tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-        if tc_num >= 18 and tc_num <= 34:
-            requested_bis_sync |= 2
-            bis_ids.append(2)
+    bis_events = {
+        bis_event['bis_id']: bis_event
+        for bis_event in stack.bap.event_queues[defs.BTP_BAP_EV_BIS_FOUND]
+        if bis_event['broadcast_id'] == broadcast_id
+    }
+    bis_events[ev['bis_id']] = ev
```
Thalley: Why is this line needed?

autopts/wid/bap.py
```diff
@@ -769,15 +767,14 @@ def hdl_wid_100(params: WIDParams):
...
-    requested_bis_sync = 1
-    bis_ids = [1]
-    if params.test_case_name.startswith('BAP/BSNK/STR') or \
-            params.test_case_name.startswith('BAP/BSRC/STR'):
-        tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-        if tc_num >= 18 and tc_num <= 34:
-            requested_bis_sync |= 2
-            bis_ids.append(2)
+    bis_events = {
+        bis_event['bis_id']: bis_event
```
Thalley: Poor naming from BTP, but `bis_id` is actually `bis_index` (not a change request, just a note since normally id != index)

autopts/wid/bap.py
```diff
@@ -769,15 +767,14 @@ def hdl_wid_100(params: WIDParams):
...
+        for bis_event in stack.bap.event_queues[defs.BTP_BAP_EV_BIS_FOUND]
+        if bis_event['broadcast_id'] == broadcast_id
+    }
+    bis_events[ev['bis_id']] = ev
+    bis_ids = sorted(bis_events)
+    requested_bis_sync = sum(1 << (bis_id - 1) for bis_id in bis_ids)
```
Thalley: This also works, but I'm wondering if

```
for bis_index in bis_ids:
    requested_bis_sync |= BIT(bis_index - 1)
```

Makes it a bit clearer what this line does (i.e. converting the list of BIS indexes to a bitfield

autopts/wid/bap.py
ciascai: Is needed because `wait_bis_found_ev(..., remove=False)` returns the matching event but does not guarantee that the event is present in the queue snapshot used by the comprehension at that point. Adding it explicitly ensures the first discovered BIS is included when building` bis_ids` and `requested_bis_sync`. It is also harmless if the event is already queued, because the dictionary is keyed by `bis_id` and simply overwrites the same entry.

autopts/wid/bap.py
ciascai: I implemented the suggested loop, with one adjustment. I use `remove=True` when waiting for each `BIS_FOUND` event. In this codebase, `remove=False` leaves the event in the queue, so the next 1-second wait would immediately return the same event and could loop forever. The code waits up to 20 seconds for the first BIS, then waits in 1 second intervals for additional BIS events. It deduplicates them by `bis_id`, builds the sync bitfield from all discovered BISes, and then synchronizes to all of them.

autopts/wid/bap.py
ciascai: I’ll keep the existing `bis_id` naming here to match the BTP API and event payload, but I agree that its value represents the BIS index rather than an allocated identifier.

autopts/wid/bap.py
```diff
@@ -761,23 +759,26 @@ def hdl_wid_100(params: WIDParams):
...
-    if params.test_case_name.startswith('BAP/BSNK/STR') or \
-            params.test_case_name.startswith('BAP/BSRC/STR'):
-        tc_num = int(re.findall(r'\d+', params.test_case_name)[0])
-        if tc_num >= 18 and tc_num <= 34:
-            requested_bis_sync |= 2
-            bis_ids.append(2)
+    bis_events = {}
+    while True:
+        wait_timeout = 20 if not bis_events else 1
+        bis_event = stack.bap.wait_bis_found_ev(broadcast_id, wait_timeout, True)
```
Thalley, line 770:
```suggestion
    while True:
        # Only apply a long timeout for the first event as any remaining events would come right after
        wait_timeout = 20 if not bis_events else 1
        bis_event = stack.bap.wait_bis_found_ev(broadcast_id, wait_timeout, True)
```

autopts/wid/bap.py
```diff
@@ -761,23 +759,26 @@ def hdl_wid_100(params: WIDParams):
...
+        if bis_event is None:
+            break
+        bis_events[bis_event['bis_id']] = bis_event
+
+    if not bis_events:
+        log(f'BIS not found, broadcast ID {broadcast_id}')
```
```suggestion
        log(f'BIS not found for broadcast ID {broadcast_id}')
```

autopts/wid/bap.py
```diff
@@ -761,23 +759,26 @@ def hdl_wid_100(params: WIDParams):
...
+        return False
+
+    bis_ids = sorted(bis_events)
+    requested_bis_sync = 0
+    for bis_index in bis_ids:
+        requested_bis_sync |= defs.BIT(bis_index - 1)
```
Thalley, line 783:
```suggestion
    requested_bis_sync = 0
    while True:
        wait_timeout = 20 if not bis_events else 1
        bis_event = stack.bap.wait_bis_found_ev(broadcast_id, wait_timeout, True)
        if bis_event is None:
            break
        
        # Add the BIS index to the bitfield (BIS index 0x01 => BIT(0), index 0x02 => BIT(1), etc.)
        requested_bis_sync |= defs.BIT(bis_event['bis_id'] - 1)

    if requested_bis_sync == 0:
        log(f'BIS not found, broadcast ID {broadcast_id}')
        return False
```

We can simplify this code as this

Thalley (changes requested):
Minor suggestions to reduce code and add some additional comments, otherwise LGTM

autopts/wid/bap.py
ciascai: Tbh, I prefer the dictionary approach for correctness. IMO it ensures we only sync to BISes where we actually received `BIS_FOUND` events, and provides an explicit trail of what triggered the sync. Reconstructing BIS IDs from a bitmask is less verifiable and harder to debug if issues arise.

ciascai: Added the comments tho.

Thalley: Not sure I understand. They are functionally identical; the `requested_bis_sync` just has the bits set directly from the events, rather than going through a dictionary than is then sorted and then iterated on. 

Can you elaborate on what you think is the issue with the more direct approach?

ciascai: fair point - they are functionally equivalent. I was erring on the side of the defensive programming, but if you consider the direct approach cleaner, happy to simplify it

Thalley: It took me a while to understand the current design at least. If you have any arguments to keep the dictionary version, then please let me know :) You mention defensive programming, but I'm not sure I follow where the dictionary approach adds anything there: Can you elaborate?

ciascai: also your suggestion has 2 issues
Line 768 -  wait_timeout = 20 if not bis_events else 1 — bis_events doesn't exist. This will crash at runtime
Line 787 - for bis_id in bis_ids: — bis_ids is never created. Another error at runtime

ciascai: Also for the direct approach, I still need `bis_ids` for the second loop, and to get those IDs, I have to reconstruct them from the bitmask with a list comprehension - which is kind of less verifiable and I also will lose the trail of what events triggered the sync

Thalley: > also your suggestion has 2 issues 
> Line 768 - wait_timeout = 20 if not bis_events else 1 — bis_events doesn't exist. This will crash at runtime 

`wait_timeout = 20 if not bis_events else 1` would obviously be `wait_timeout = 20 if requested_bis_sync == 0 else 1`

> Line 787 - for bis_id in bis_ids: — bis_ids is never created. Another error at runtime

`bis_ids` can easily be replaced by looping over the bits in `requested_bis_sync`, or we can easily reestablish it by simply doing `bis_event['bis_id'] = {}` followed by `bis_ids[bis_event['bis_id']] = bis_event['bis_id']`

I think you are misinterpreting my above suggestion as a "copy this into the code" rather than a suggest _design_. I have not tested it, but it's good that you noticed these short comings

> which is kind of less verifiable 

Can you elaborate on why? 

> lose the trail of what events triggered the sync

Not sure I understand this one: Do you mean that you cannot e.g. loop over and print `bis_events` during debugging? What values in these events are useful that the bitfield, which is what we are using to sync to, do not contain? 

It's not really a big deal anyhow; performance is not important here, so the additional loop and memory copies etc. doesn't really matter. The reason why I brought it up is that it's non-intuitive to me to store a complete dictionary of events that is then reduced to a list of integers via `sorted`, when all you want is a list of BIS indexes / bitfield of BIS indexes

Thalley (approved):

PR #1849, opened 2026-09-03
GMCS: implement Media Player Name oversized-value

ciascai:
Add BTP support for updating oversized Media Player Name
value on GMCS, wire up hdl_wid_20141/hdl_wid_20142 to
trigger these instead of returning False.

GMCS/SR/SPN/BV-01-C

Testing
nRF5340 Audio DK
nRF54L15 DK


autopts/wid/gmcs.py
```diff
@@ -241,18 +241,19 @@ def hdl_wid_20141(params: WIDParams):
     """Please update Media Player Name/Track Title characteristic with value
        whose length is greater than the (ATT_MTU-3). Click OK when it is ready."""
 
-    # TODO: GMCS/SR/SPN/BV-01-C, GMCS/SR/SPN/BV-02-C
-    #  There is no way to do that with current API.
+    if params.test_case_name == 'GMCS/SR/SPN/BV-01-C':
+        return True
```
Thalley, line 246: I don't understand this change. The WID description is telling us to update the name to be larger than ATT MTU, but the implementation is not doing anything. Shouldn't we set the name here?

autopts/wid/gmcs.py
```diff
@@ -241,18 +241,19 @@ def hdl_wid_20141(params: WIDParams):
...
 
     return False
 
 
 def hdl_wid_20142(params: WIDParams):
-    """Please update Track Title characteristic(Handle = 0x005E) and send a
+    """Please update Media Player Name/Track Title characteristic and send a
```
Thalley, line 252: Is PTS calling both this and `hdl_wid_20141`? :s

autopts/wid/gmcs.py
```diff
@@ -241,18 +241,19 @@ def hdl_wid_20141(params: WIDParams):
...
        notification containing the updated value of the characteristic with
        a different value whose length is greater than the (ATT_MTU-3)."""
 
-    # TODO: GMCS/SR/SPN/BV-01-C, GMCS/SR/SPN/BV-02-C
-    #  There is no way to do that with current API.
+    if params.test_case_name == 'GMCS/SR/SPN/BV-01-C':
+        btp.gmcs_player_name_set("Player0 oversized update")
```
Thalley: Please remove all mentions of "oversized": it's not oversized; it's just larger than the (default) ATT MTU

Thalley (changes requested):

autopts/wid/gmcs.py
ciascai: yes

autopts/wid/gmcs.py
ciascai: and what do you want to mention in the new name?

autopts/wid/gmcs.py
ciascai: WID 20141 only acknowledges readiness because it occurs before PTS reads the initial oversized value. The actual update happens in WID 20142, after CCCD enable and the initial read, so PTS can verify the notification and 0x80 “Value Changed During Read Long” response.

Thalley: Ah, now I get it. 

So in `20141` we need to set a name > ATT MTU, so that PTS then can perform a read (of <= ATT_MTU), and then in 20142 we are requested to update the value, so that when PTS attempts to read the second part of the name, we reject that as the MCS specification. 

In that case, we need to set a long (> ATT MTU) name in both handlers with different names. 

It can just be something as simple as 

`"This is a very long Media Player Name #1"`
`"This is a very long Media Player Name #2"`

It's also possible to set the WID number in it for debugging purposes (e.g. ``"This is a very long Media Player Name (20241)"``)

autopts/wid/gmcs.py
Thalley: See https://github.com/auto-pts/auto-pts/pull/1849/changes#r3949972544

autopts/wid/gmcs.py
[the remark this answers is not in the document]
ciascai: done

ciascai: done

autopts/pybtp/btp/gmcs.py
```diff
@@ -148,5 +151,15 @@ def gmcs_parent_group_set():
     return gmcs_command_rsp_succ()
 
 
+def gmcs_player_name_set(name):
+    logging.debug("")
```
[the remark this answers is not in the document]
ciascai: the pattern in this file is not with documented helpers

autopts/pybtp/btp/gmcs.py
```diff
@@ -148,5 +151,15 @@ def gmcs_parent_group_set():
...
+
+    data = bytes([len(name)]) + name.encode("utf-8")
```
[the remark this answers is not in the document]
ciascai: I don't think that this test will ever have a non-ASCII player name

autopts/wid/gmcs.py
```diff
@@ -241,18 +241,20 @@ def hdl_wid_20141(params: WIDParams):
...
 def hdl_wid_20142(params: WIDParams):
-    """Please update Track Title characteristic(Handle = 0x005E) and send a
+    """Please update Media Player Name/Track Title characteristic and send a
        notification containing the updated value of the characteristic with
        a different value whose length is greater than the (ATT_MTU-3)."""
 
-    # TODO: GMCS/SR/SPN/BV-01-C, GMCS/SR/SPN/BV-02-C
-    #  There is no way to do that with current API.
+    if params.test_case_name == 'GMCS/SR/SPN/BV-01-C':
+        btp.gmcs_player_name_set("Player Name Updated Value 2")
```
Thalley: The `1` and `2` does not really (semantically) make sense when the remaining parts are changed. Suggest to just remove the numbers 


```suggestion
        btp.gmcs_player_name_set("Player Name Updated Value")
```

Thalley (changes requested):

autopts/pybtp/defs.py
```diff
@@ -899,6 +899,7 @@ def BIT(bit):
 BTP_GMCS_CMD_NEXT_TRACK_OBJ_ID_GET = 0x04
 BTP_GMCS_CMD_INACTIVE_STATE_SET = 0x05
 BTP_GMCS_CMD_PARENT_GROUP_SET = 0x06
+BTP_GMCS_CMD_PLAYER_NAME_SET = 0x07
```
Thalley, line 902: This new command must be added and documented in `btp_gmcs.txt`

Thalley (changes requested):

autopts/pybtp/defs.py
ciascai: done

doc/btp_gmcs.txt
```diff
@@ -55,34 +55,43 @@ Commands and responses:
...
+        Opcode 0x05 - Set Inactive State
 
                 Controller Index:       <controller id>
                 Command parameters:     <none>
                 Response parameters:    Media Player State (1 octet)
 
                 Set media player into inactive state.
 
-        Opcode 0x07 - Set Parent Group
+        Opcode 0x06 - Set Parent Group
```
Thalley, line 82: Thanks for fixing these too :D

doc/btp_gmcs.txt
```diff
@@ -55,34 +55,43 @@ Commands and responses:
...
                 Controller Index:       <controller id>
                 Command parameters:     <none>
                 Response parameters:    <none>
 
                 Set Media Player's Current Group to be it's own Parent Group.
+
+        Opcode 0x07 - Set Player Name
+
+                Controller Index:       <controller id>
+                Command parameters:    Name length (1 octet)
```
Thalley:
```suggestion
                Controller Index:      <controller id>
                Command parameters:    Name length (1 octet)
```

Thalley (approved):
Minor comment

Thalley (approved):

PR #1852, opened 2026-09-08
BAP: wait for BIGInfo before synchronizing BISes

ciascai:
Wait for the Periodic BIGInfo event after discovering BISes
and before requesting BIS synchronization. This avoids racing
BIS discovery against BIGInfo processing on the broadcast sink
and fixes BSNK STR synchronization failures.

### DIFF
--- a/autopts/wid/bap.py
+++ b/autopts/wid/bap.py
@@ -765,11 +765,14 @@ def hdl_wid_100(params: WIDParams):
     if params.test_case_name.startswith('BAP/BA'):
         return True
 
+    biginfo = stack.gap.read_periodic_biginfo(timeout=20)
+    if biginfo is None:
+        log(f'BIGInfo not found for broadcast ID {broadcast_id}')
+        return False
+
     bis_events = {}
-    while True:
-        # Only apply a long timeout for the first event as any remaining events would come right after
-        wait_timeout = 20 if not bis_events else 1
-        bis_event = stack.bap.wait_bis_found_ev(broadcast_id, wait_timeout, True)
+    for _ in range(biginfo.num_bis):
+        bis_event = stack.bap.wait_bis_found_ev(broadcast_id, 20, True)
         if bis_event is None:
             break
         bis_events[bis_event['bis_id']] = bis_event
--- a/autopts/wid/pbp.py
+++ b/autopts/wid/pbp.py
@@ -79,6 +79,11 @@ def hdl_wid_100(_: WIDParams):
 
     bis_id = 1
     requested_bis_sync = 1
+
+    if stack.gap.read_periodic_biginfo() is None:
+        log(f'BIGInfo not found for broadcast ID {broadcast_id}')
+        return False
+
     btp.bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync, addr_type, addr)
 
     ev = stack.bap.wait_bis_synced_ev(broadcast_id, bis_id, 20, False)
--- a/autopts/wid/bass.py
+++ b/autopts/wid/bass.py
@@ -188,6 +188,11 @@ def hdl_wid_113(_: WIDParams):
 
     # BIS_Sync bitfield uses bit 0 for BIS Index 1
     requested_bis_sync = 1 << (ev['bis_id'] - 1)
+
+    if stack.gap.read_periodic_biginfo(timeout=20) is None:
+        log(f'BIGInfo not found for broadcast ID {broadcast_id}')
+        return False
+
     btp.bap_broadcast_sink_bis_sync(ev['broadcast_id'], requested_bis_sync)
 
     bis_id = ev['bis_id']
--- a/autopts/wid/tmap.py
+++ b/autopts/wid/tmap.py
@@ -104,6 +104,10 @@ def hdl_wid_100(params: WIDParams):
     requested_bis_sync = 1
     bis_ids = [1]
 
+    if stack.gap.read_periodic_biginfo(timeout=20) is None:
+        log(f'BIGInfo not found for broadcast ID {broadcast_id}')
+        return False
+
     btp.bap_broadcast_sink_bis_sync(broadcast_id, requested_bis_sync)
 
     for bis_id in bis_ids:


autopts/wid/bap.py
```diff
@@ -784,6 +784,10 @@ def hdl_wid_100(params: WIDParams):
         # Add the BIS index to the bitfield (BIS index 0x01 => BIT(0), index 0x02 => BIT(1), etc.)
         requested_bis_sync |= defs.BIT(bis_index - 1)
 
+    if stack.gap.read_periodic_biginfo(timeout=20) is None:
+        log(f'BIGInfo not found for broadcast ID {broadcast_id}')
+        return False
```
Thalley: Don't we need this for bass.py:hdl_wid_113, pbp.py:hdl_wid_100 and tmap.py:hdl_wid_100 as well?

ciascai: not for bass, b/c wid 113 does not issue that command. it starts sink synchronization and only waits for BIS_FOUND and BIS_SYNCED, so ading `read_periodic_biginfo()` would merely consume a queued event without controlling the synchronization request

but pbp wid 100 explictitly calls `btp.bap_broadcast_sink_bis_sync(...)` so it should wait for BIGInfo immediately before that call, just like in BAP wid 100 and also the same for tmap

do you want me to address them in the same PR with different commits? I would rather create a new PR for them, to focus this PR only for the regression

Thalley: > not for bass, b/c wid 113 does not issue that command. it starts sink synchronization and only waits for BIS_FOUND and BIS_SYNCED, so ading read_periodic_biginfo() would merely consume a queued event without controlling the synchronization request

I'm not sure I understand this comment. `hdl_wid_113` calls `bap_broadcast_sink_bis_sync`, so why shouldn't it wait for the BIGInfo?

Thalley: > do you want me to address them in the same PR with different commits? I would rather create a new PR for them, to focus this PR only for the regression

Yes, let's fix it everywhere

ciascai: > > not for bass, b/c wid 113 does not issue that command. it starts sink synchronization and only waits for BIS_FOUND and BIS_SYNCED, so ading read_periodic_biginfo() would merely consume a queued event without controlling the synchronization request
> 
> I'm not sure I understand this comment. `hdl_wid_113` calls `bap_broadcast_sink_bis_sync`, so why shouldn't it wait for the BIGInfo?

yeah, you are right, my bad looked between the lines in bass wid 100 and 113
>Yes, let's fix it everywhere

In this PR with a separate commit or a new PR?

Thalley: > In this PR with a separate commit or a new PR?

I think this PR makes sense to do it, as it's the same fix applies multiple places, but either way works

autopts/wid/pbp.py
```diff
@@ -79,6 +79,11 @@ def hdl_wid_100(_: WIDParams):
 
     bis_id = 1
     requested_bis_sync = 1
+
+    if stack.gap.read_periodic_biginfo(timeout=20) is None:
```
Thalley: Any reason why we can't just use the default timeout of 10 seconds?

ciascai: no particular reason, just used 20s to match the BIS discovery above

ciascai: I think using the default 10-s timeout should be fine also for `read_periodic_biginfo()`

Thalley: Either way works I guess :D But generally I would recommend just using the default unless other timeouts are necessary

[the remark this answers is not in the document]
ciascai: @Thalley this make sense I believe. The `read_periodic_biginfo()` currently consumes whichever BIGInfo is first in the global queue, which could belong to another broadcast?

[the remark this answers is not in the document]
Thalley: If we only have one sync, then that's not an issue, but otherwise I agree. We could expand `read_periodic_biginfo` to take some parameters so that we can ensure that we are waiting for the "right" BIGInfo. Maybe that should be done as a `wait_biginfo` instead similar to `wait_bis_found_ev` and the others. I'm not sure why `read_periodic_biginfo` has been implemented as it has. There doesn't seem to be any tests for multiple BIGs (please correct me if I'm wrong) so this is not immediately needed and we can probably create a GH issue for adding this functionality instead of needing to fix it in this PR (unless there are tests where we sync to multiple BIGs)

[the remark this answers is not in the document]
ciascai: I don't see any current test cases where the IUT syncs to multiple BIGs concurrently in a single test - `broadcast_id_2` only shows up on the broadcast source side for the second announcement, not as a concurrent sink sync. Yeah, we can leave it as is for now

autopts/wid/bap.py
```diff
@@ -765,11 +765,14 @@ def hdl_wid_100(params: WIDParams):
...
+
     bis_events = {}
-    while True:
-        # Only apply a long timeout for the first event as any remaining events would come right after
-        wait_timeout = 20 if not bis_events else 1
-        bis_event = stack.bap.wait_bis_found_ev(broadcast_id, wait_timeout, True)
+    for _ in range(biginfo.num_bis):
+        bis_event = stack.bap.wait_bis_found_ev(broadcast_id, 20, True)
```
Thalley, line 775: Ah, that's a nice change to use the count from BIGInfo to use that to verify the number of BIS in the BASE :)

Thalley (approved):

autopts/wid/bap.py
```diff
@@ -784,6 +787,10 @@ def hdl_wid_100(params: WIDParams):
         # Add the BIS index to the bitfield (BIS index 0x01 => BIT(0), index 0x02 => BIT(1), etc.)
         requested_bis_sync |= defs.BIT(bis_index - 1)
```
Thalley, line 788: Outside the scope of this PR, but we could consider updating the BASS, PBP and TMAP parts to also actually sync to all BIS in the broadcast, instead of the currently hardcoded BIS Index 0x01 only. 

Probably could put this whole ("Wait for BIGInfo and the appropriate number of wait_bis_found_ev and then sync to the entire broadcast") part into a helper function to apply everywhere

